PR #30, opened 2026-02-17
Fix settings UI corruption on dedicated server (v1.0.5.2)

TheCodingDad-TisonK:
## Summary

Fixes issue #21 — settings pages (Graphics, Server Settings, Better Contracts) appearing white/blank when the mod is loaded on a dedicated server.

- **Double-add bug**: `clone(layout)` in FS25 already adds the element to the layout. All four `UIHelper` create functions were also calling `layout:addElement()` separately, inserting every element twice. This corrupted the layout array and broke rendering on dedicated server clients.
- **Non-admin MP injection**: Non-admin clients on dedicated servers had the full settings UI injected even though they cannot change settings. Template cloning on dedicated server clients found elements from unexpected layout contexts, reparenting those elements away from other pages (Graphics, Server Settings, Better Contracts, etc.) — causing them to go white.

## Fix

1. Removed the four redundant `layout:addElement()` calls in `UIHelper.lua` — `clone(layout)` handles adding the element.
2. Added early return in `SoilSettingsUI:inject()` for non-admin MP clients — no injection, no corruption. They retain full HUD and console access to view soil data.
3. Removed now-dead "admin only info text" block.
4. Bumped version to 1.0.5.2.

### DIFF
--- a/src/ui/SoilHUD.lua
+++ b/src/ui/SoilHUD.lua
@@ -78,164 +78,6 @@ function SoilHUD:delete()
     SoilLogger.info("Soil HUD overlay deleted")
 end
 
--- Get current farmland ID based on player/vehicle position
--- NOTE: Changed from getCurrentFieldId to getCurrentFarmlandId to match base game behavior
--- Base game shows "Farmland 1", "Farmland 2" etc, not field IDs
-function SoilHUD:getCurrentFarmlandId()
-    if not g_currentMission then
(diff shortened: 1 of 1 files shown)


TheCodingDad-TisonK:
[log (1).txt](https://github.com/user-attachments/files/25362123/log.1.txt)

From @RustFarming

PR #265, opened 2026-04-28
feat: mower hook — forage crop (grass/alfalfa) nutrient tracking

antler22:
## Summary

Adds `installMowerHook()` on `Mower.onEndWorkAreaProcessing` so mowing grass, alfalfa, clover, and other forage crops removes N/P/K from the field. Previously, only direct-threshed crops (combine → Cutter path) triggered nutrient depletion — any field used exclusively for hay/silage would never deplete regardless of how many times it was cut.

### Why `Mower` and not `Cutter`
`Cutter.processCutterArea` reads only the standing-crop density map and returns 0 area for windrow-pickup passes. `Mower.onEndWorkAreaProcessing` fires for all disc mowers, drum mowers, and swathers (any implement using the Mower specialization).

### Approach
Uses area-based depletion via a new `SoilFertilitySystem:onMow(fieldId, fruitTypeIndex, areaHa)` method, since no yield liters are produced at mow time. Area comes from `spec_mower.workAreaParameters.lastStatsArea` (density-map pixels), converted to ha via `MathUtil.areaToHa + getFruitPixelsToSqm()` — the same call Mower.lua itself uses internally.

### Calibration
`MOWER_HA_FACTOR = 6.0` → mowing 1 ha of grass depletes ~8.4 N-units (~57% of a 1-ha wheat harvest). Grass is typically cut 3-4 times per season, so cumulative depletion per season is comparable to or exceeds a grain crop. `CROP_EXTRACTION_FORAGE = {N=1.40, P=0.55, K=1.80}` applies to grass, alfalfa, and any forage crop not listed in `CROP_EXTRACTION`.

### Files changed
| File | Change |
|---|---|
| `src/hooks/HookManager.lua` | `installMowerHook()` added; wired into `installAll()` as Hook 1b |
| `src/SoilFertilitySystem.lua` | `onMow(fieldId, fruitTypeIndex, areaHa)` added after `onHarvest` |
| `src/config/Constants.lua` | `CROP_EXTRACTION_FORAGE` and `MOWER_HA_FACTOR` added after `CROP_EXTRACTION_DEFAULT` |

## Test plan
- [ ] Mow a grass field; confirm `[MowerHook]` debug log fires and N/P/K values drop
- [ ] Verify no depletion occurs when no fruit type is detected (non-crop area)
- [ ] Confirm the hook fails gracefully (warning, returns false) if `Mower.onEndWorkAreaProcessing` is unavailable
- [ ] Multiplayer: confirm `SoilFieldUpdateEvent` broadcasts to clients after mow

### DIFF
--- a/src/SoilFertilitySystem.lua
+++ b/src/SoilFertilitySystem.lua
@@ -304,11 +304,15 @@ function SoilFertilitySystem:onMow(fieldId, fruitTypeIndex, areaHa)
         return
     end
 
-    -- Look up forage-specific rates first, then per-crop, then default
-    local name = string.lower(fruitDesc.name or "unknown")
-    local rates = SoilConstants.CROP_EXTRACTION[name]
-        or SoilConstants.CROP_EXTRACTION_FORAGE
-        or SoilConstants.CROP_EXTRACTION_DEFAULT
+    -- Mowing ALWAYS uses CROP_EXTRACTION_FORAGE, never the per-crop grain rates.
+    -- Reason: CROP_EXTRACTION rates are calibrated for harvested grain volume/density
+    -- (threshed yield, e.g. wheat at 0.39 L/sqm). The Mower spec cuts whole-plant
+    -- biomass (e.g. wheat windrow at 3.8 L/sqm) — a completely different density.
+    -- Using grain rates with windrow-equivalent area would over-extract by ~10x.
+    -- CROP_EXTRACTION_FORAGE is calibrated for cut green biomass at MOWER_HA_FACTOR.
+    -- This also prevents "mowing wheat before harvest depletes N faster than combining"
+    -- scenarios that would confuse players (TisonK's review note on PR #265).
+    local rates = SoilConstants.CROP_EXTRACTION_FORAGE or SoilConstants.CROP_EXTRACTION_DEFAULT
 
     local diffMult = SoilConstants.DIFFICULTY.MULTIPLIERS[self.settings.difficulty] or 1.0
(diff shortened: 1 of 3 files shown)


TheCodingDad-TisonK:
@antler22 The overall structure is solid and fits the codebase well. One bug needs fixing before this can merge:

**`getFruitPixelsToSqm()` is a method on `g_currentMission`, not a global.** Your current call will throw a Lua error at runtime every time a mower is used. The `pcall` will catch it silently, so nutrients never deplete and you'll get `[MowerHook] failed:` spam in the log.

Fix:
```lua
-- current (broken)
local areaHa = MathUtil.areaToHa(area, getFruitPixelsToSqm())

-- correct
local areaHa = MathUtil.areaToHa(area, g_currentMission:getFruitPixelsToSqm())
```
This matches how `Mower.lua` itself calls it internally (and `SowingMachine.lua` for reference).

Everything else looks good — hook wiring, cleanup registration, `pcall` guard, server check, MP broadcast, and the `onMow()` structure all follow the existing patterns correctly. The `MOWER_HA_FACTOR` calibration math also checks out.

One thing worth a quick discussion: the lookup order in `onMow` falls through to `CROP_EXTRACTION[name]` first, so mowing a grain crop (wheat, canola, etc.) before harvest will use grain extraction rates rather than forage rates. That's defensible agronomically but might surprise players. Up to you whether you want an explicit forage-crop allowlist instead.

Fix the `g_currentMission` call and this is ready to merge.

Issue #538, opened 2026-06-03
Auto rate change


Arissani:
### Mod Version

2.3.5.0

### Farming Simulator 25 Version

1.8

### Game Mode

Singleplayer

### What happened?

<img width="2495" height="1360" alt="Image" src="https://github.com/user-attachments/assets/fb383954-2172-4f27-a46a-c4446f490e83" />

A - weather you set SF to spray 8gal/ac or 48gal/ac the base game xliter per minute display in the top right never changes it always sprays the same amount in the display no matter what you change SF gal/ac amount too

for example if you are only spraying 8gal/ac with SF shouldn't the liters per min in top right be spraying around 100L per min and if you change SF to 48gal/ac it should be spraying around 700 liters per min (numbers are just an example)

meaning regardless of if you spray a field with 8gal/ac or 48gal/ac , you will use the same amount of product to spray the field

### Steps to Reproduce

fill sprayer with any product
start spraying 
toggle SF application rate up and down 
observe base game application rate (top right) doesn't change at all 

### Log Snippet

```text
n/a
```

### Other Mods Loaded

n/a

TheCodingDad-TisonK:
@Arissani Try latest release and let me know :)

Arissani:
@TheCodingDad-TisonK haven't tested harvesting yet that will come in the coming few hours but so far everything working as intended apart from some flickering the mini map does while driving around but i gotta test a bit more and ill make a report for it

TheCodingDad-TisonK:
Yes im aware myself too... cant seem to find "why" its doing that.. Might  be due to a fallback? I dunno

Arissani:
im not entirely sure i can nail down that it doesn't happen when actually spraying or any field work but does happen when transiting around and that it happens when using a rock picker and tractor even

Arissani:
there was one occasion earlier that i was testing spreading solid lime with a mod spreader (not a stieve one) and tabbed to another vehicle and then when i tabbed back the lime spreading animation wasn't there but i can't reproduce it

Arissani:
@TheCodingDad-TisonK  UPDATE......this is still an issue the original bug report about rate change 

refer back to photo 

A - weather you set SF to spray 8gal/ac or 48gal/ac the base game xliter per minute display in the top right never changes it always sprays the same amount in the display no matter what you change SF gal/ac amount too

[...]

PR #549, opened 2026-06-04
fix: rate multiplier now affects L/min display and physical consumption (#538)

TheCodingDad-TisonK:
## Summary

- **Root cause**: `getSprayerUsage` requires 3-layer instance-table patching because `copyTypeFunctionsInto` copies function refs into vehicle instances at load time. Layer-3 live patching was unreliable in some FS25 versions — rate multiplier changes had no visible effect on the L/min HUD display or actual tank drain.
- **Fix**: Moved rate multiplier application to `Sprayer.onStartWorkAreaProcessing`, which is registered as an **event listener** (not `registerFunction`). Class-table patches via `Utils.appendedFunction` reach all vehicles reliably for event listeners — no 3-layer patching required.
- **Also fixes double-application bug**: `effectiveLiters` in `onEndWorkAreaProcessing` was `liters * rateMultiplier`, but `liters` (= `wap.usage`) already included the multiplier on vehicles where the old patch worked → nutrient gain was `multiplier²`. Now `effectiveLiters = liters` directly.

## Changes

- `installSprayerStartHook()` — new function, appended to `Sprayer.onStartWorkAreaProcessing`, multiplies `wap.usage` and `wap.usagePerMin` by the current rate multiplier after vanilla sets them
- `installSprayerUsageHook()` — removed `mapMult` from usage formula (keeps actual-speed fix)
- `installSprayerAreaHook()` — `effectiveLiters = liters` (no longer re-applies rate multiplier; keeps `rateMultiplier` for burn-threshold check)
- `installAll()` — calls `installSprayerStartHook()` in the hook sequence

## Test plan

- [ ] Set rate to 2.0×, spray a field → L/min HUD display should double vs 1.0×
- [ ] Same product, same speed — tank should drain approximately twice as fast at 2.0× vs 1.0×
- [ ] Nutrient gain at 2.0× should be 2× base, not 4× (no double-application)
- [ ] Burn effects still trigger at rates above threshold
- [ ] AI spraying (Courseplay) correctly uses actual speed, not speedLimit

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,7 +432,7 @@ This mod is licensed under **[CC BY-NC-ND 4.0](https://creativecommons.org/licen
 
 You may share it in its original form with attribution. You may not sell it, modify and redistribute it, or reupload it under a different name or authorship. Contributions via pull request are explicitly permitted and encouraged.
 
-**Author:** TisonK &nbsp;·&nbsp; **Version:** 2.3.6.0
+**Author:** TisonK &nbsp;·&nbsp; **Version:** 2.3.7.0
 
 © 2026 TisonK — See [LICENSE](LICENSE) for full terms.
 
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK</author>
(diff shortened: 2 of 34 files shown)


TheCodingDad-TisonK:
Superseded by 2.3.7.0 release PR.

Issue #651, opened 2026-06-16
FieldSentry: backend field-eligibility gate for the soil sim (meadows, fake/deco fields, NPC contracts, manual blacklist)


TheCodingDad-TisonK:
Proposal author: @arissani. Thanks for writing this up in so much detail. Copying it in here so we have a clean place to discuss and track it. My maintainer notes are at the bottom.

## What it is

FieldSentry is a lightweight backend gate that sits in front of the Soil and Fertilizer update loops. For any field it answers one binary question: should this area run simulation cycles right now? It does not change any of the soil equations. It just keeps the sim focused on valid fields and skips the rest.

## Problems it solves

- Fake fields. Decorative terrain that the game misreads as a field gets simulated for no reason.
- Meadows. Grassland that should follow grazing logic instead of crop-rotation logic.
- NPC contract fields. Land under an active AI worker mission gets masked so we do not fight the contract.
- Performance. Unowned areas on big maps drop into a zero-overhead sleep state.

## Design goals

- Backend only. No native UI of its own, no changes to the soil equation code.
- Amortized evaluation. Hard per-frame budget, proposal suggests max 5 fields per tick.
- No run-time allocations. Reuse tables in place to avoid GC spikes.
- Namespace safety. Encapsulated registry (FieldSentry_Core) so it does not collide with other mods.
- Cache engine pointers at module level so we are not doing expensive manager lookups in the hot path.

## State model

```lua
FieldSentry_Core.FieldState[fieldId] = {
    manualBlacklist,    -- boolean (persistent player intent)
    meadowToggle,       -- boolean (persistent player intent)
    evaluatedBlacklist, -- enum   (dynamic status mask)
    lastSeq,            -- integer (MP sequence protection token)
    hints = {}          -- table  (zero-allocation transient UI diagnostics)
}
```

## Rule engine

Structural rules run first as hard constraints. If any returns non-NONE, evaluation short-circuits. Order: Manual, Farmland Block, Ownership, NPC Contract, Inactivity.

Classification rules only run if structural returned NONE:
- Deco detection returns DECO_AREA if matched.
- Meadow identification writes f.hints.meadowHint = true. It is only a hint. FieldSentry never auto-flips the persistent meadow setting a player chose.

## Backend API

```lua
-- Hot path, O(1) lookup
function FieldSentry_API.isFieldSimDisabled(fieldId)
    local f = FieldSentry_Core.FieldState[fieldId]
[...]

PR #655, opened 2026-06-17
FieldSentry Phase 1: backend gate + manual blacklist + status readout (#651)

TheCodingDad-TisonK:
Implements **Phase 1** of FieldSentry (#651), the backend field-eligibility gate proposed by @arissani. Scope agreed in the issue: core backend + manual blacklist + map-tab status readout. Meadow profile, deco/NPC detection and NPCFavor integration are later phases.

It layers onto the existing active-set + `DAILY_BATCH_SIZE` loop with a single O(1) gate, rather than standing up a parallel scheduler. No soil-equation code is touched.

## Commits
- **Backend core** - `src/FieldSentry.lua`: encapsulated `FieldSentry_Core` + `FieldSentry_API` (`isFieldSimDisabled` hot path, `getUIStatus`, `setFieldManual`/`toggleFieldManual`/`isFieldManual`, `reset`). The freeze is one early return at the top of `SoilFertilitySystem:_processOneDailyField`, so a slept field skips the daily depletion/recovery/leaching pass and its values stay put.
- **Persistence** - manual blacklist saved into `soilData.xml` (already savegame/map scoped) and restored on load, using only `setXMLInt`/`getXMLInt`.
- **Multiplayer** - `SoilFieldSentryEvent` + `SoilNetworkEvents_SendFieldSentryToggle`: server-authoritative, admin-validated, broadcast to clients. Mirrors the existing setting/auto-mode events.
- **Map-tab readout** - `getFieldInfo` returns `simDisabled` + `simDisabledReason`; the field detail dialog shows `(sim asleep: <reason>)` so a frozen field reads as intentional, not as a stalled sim.

## Console (server/SP for now)
- `SoilBlacklistField <fieldId> [true|false]` - sleep/wake a field
- `SoilFieldSentry [fieldId]` - status for a field, or list slept fields

## Testing
- 21 FieldSentry assertions in the offline harness (status API, toggle, persistence round-trip, the freeze-gate integration test). Full suite green: `bash tools/test/run.sh`.

## Still to do (not blockers for Phase 1 review)
- In-game test of the MP event over a real connection, and the dialog suffix placement (the harness can't drive streams or render).
- `sf_fieldsentry_asleep` l10n key across the 26 languages (dialog falls back to English today).
- `_rebuildActiveList` filter so slept fields aren't enqueued (perf), and join-sync of the blacklist to clients who connect after a toggle.

Part of #651 (multi-phase, keep open).

### DIFF
--- a/src/FieldSentry.lua
+++ b/src/FieldSentry.lua
@@ -1,0 +1,163 @@
+-- =========================================================
+-- FS25 Realistic Soil & Fertilizer - FieldSentry (backend gate)
+-- =========================================================
+-- A lightweight backend that answers one binary question per field:
+-- "should this area run soil simulation right now?" It never touches the
+-- soil equations; the sim just consults it and skips disabled fields.
+--
+-- Proposal: @arissani (issue #651). This is the Phase 1 cut:
+--   - encapsulated registry (FieldSentry_Core) + status/reason API (FieldSentry_API)
+--   - manual blacklist (player chooses to "sleep" a field; its soil values freeze)
+-- Meadow profile, deco/fake-field detection and NPC-contract masking are later
+-- phases and only need new BLACKLIST reasons + classification rules layered on top.
+--
+-- Design constraints honoured here:
+--   - Backend only: no UI, no changes to soil equations.
+--   - Namespace safety: everything hangs off FieldSentry_Core / FieldSentry_API.
+--   - No hot-path allocations: isFieldSimDisabled returns a shared EMPTY_HINTS table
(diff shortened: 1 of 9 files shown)


TheCodingDad-TisonK:
Folding this into #658. development does not have Phase 1 yet, so #658 now carries both the Phase 1 backend and the Phase 2 contract integration as one PR into development, with the four Phase 1 commits unchanged at the base. Closing this so there is a single review to land. Nothing is lost, it is all in #658.

PR #746, opened 2026-07-24
fix(ui): rotation planner dialog prev/next field footer (#744)

WizardlyPayload:
## What
#744 polish: real Prev/Next field footer buttons on the Rotation Planner dialog.

Prev/next were wired through `setMenuButtonInfo` (TabbedMenuFrame API) which does nothing on a ScreenElement dialog. Replaced with XML footer buttons (SoilScoutDialog BoxLayout pattern).

- `onClickPrev` / `onClickNext` call existing `_step`
- Reuse `sf_rp_prev` / `sf_rp_next` i18n (already in all 27 langs)
- Remove dead `menuButtonInfo` block from `onOpen`
- Back button unchanged

## Test
Open Rotation Planner from Soil PDA → content fills → Prev/Next cycles `Field #id (i / n)` + candidates → Back closes.

-- Wizard

### DIFF
--- a/src/ui/RotationPlannerDialog.lua
+++ b/src/ui/RotationPlannerDialog.lua
@@ -98,21 +98,17 @@ function RotationPlannerDialog:onOpen()
             if id == self._startFieldId then self._index = i break end
         end
     end
-    self.menuButtonInfo = {
-        { inputAction = "MENU_BACK", callback = function() self:onClickBack() end },
-        { inputAction = "MENU_PAGE_PREV", text = tr("sf_rp_prev", "Prev field"),
-          callback = function() self:_step(-1) end },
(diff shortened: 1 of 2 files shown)


TheCodingDad-TisonK:
Thanks Wizard - clean convergent fix. K independently built the identical prev/next footer tonight and it already landed on development as f8d6fc5a: same BoxLayout fs25_dialogButtonBox, same rpPrevBtn/rpNextBtn/rpBackBtn (buttonOK x2 + buttonBack), same onClickPrev/onClickNext to _step, same dead menuButtonInfo removal. It is in the deployed playtest build, so this would be a no-op merge. Closing as already-landed (not rejected) - your independent identical solution is a nice confirmation the approach is right.

Issue #780, opened 2026-08-03
Additional Tanks on Tractors are not detected correctly


mckenna95:
### Mod Version

2.4.6.0

### Farming Simulator 25 Version

1.21.1

### Game Mode

Singleplayer

### What happened?

When planting crops using a seeder/planter with a liquid fertiliser tank, and having additional tanks attached to the tractor, if the planter's liquid tank drains out, the game will continue to empty the additional tanks of liquid fertiliser, but the mod does not recognise N/P/K being applied to the field. 

The HUD will lose the Coverage/Pass %/fertilizer type and act as though nothing is being applied, yet will still drain the tank. 

In the base game, this functions correctly - it will drain the planter's tank first, and then move on to any additional helicopter/additional tanks and drain fertiliser from those, and add it to the field. 

From looking at HookManager.lua, the sprayer application hook appears to read `spec.workAreaParameters.sprayFillType`, `usage`, and `sprayFillLevel` from the applicator vehicle, then exits if `sprayFillLevel <= 0`.

It also resolves fill type using `SoilUtils.resolveSprayerFillTypeIndex(self, fallbackIndex)`, which checks the applicator’s own sprayer fillUnit. The multi-tank logic also only loops through `self.spec_fillUnit.fillUnits`.

So it looks like RSF handles multiple fillUnits on the applicator itself, but may not follow an external/root/attached supply tank when the product is being drawn from tractor-mounted side tanks.

### Steps to Reproduce

1) Set up a tractor with additional tanks & a planter (Such as Custom Modding's 9R + Demco 1200 SideQuest tanks + John Deere 1775NT planter) 
2) Fill planter + tank with fertliser
3) Plant until planter's fertiliser tank drains - it will update NPK profile
4) Let the game automatically start draining the side tanks - it will not update the NPK values but will drain out the fertliser and "apply" it. 

### Log Snippet

```text
I checked my log, but the current RSF logging does not appear to throw a runtime message when the sprayer application hook exits because sprayFillLevel is 0. The log confirms that the Sprayer/Spreader hook, External Fill hook, SprayerUsage hook, and vehicle-to-vehicle transfer support were installed, but there is no visible noFillLevel / sprayFillLevel skip message during application.
```

### Other Mods Loaded

No additional mods minus the Demco SideQuest tanks, 9R and 1775NT for testing. Also EasyDevControls to set fillLevels for testing purposes.

[...]

Issue #778, opened 2026-08-03
No nitrate from Oilseed


Amotaraso:
### Mod Version

2.4.6.0

### Farming Simulator 25 Version

1.21.1.0

### Game Mode

Singleplayer

### What happened?

After my Sorghum harvest I plowed the field with a subsoiler and planted Oilseed Radish.
Oilseed was planted in September.
I let it stay until April, when I wanted to plant Soybeans.
So it was in the ground for 7 months.
I used the Väderstad Proceed V24 to plant and cultivate at the same time. Also filled the FH2200 with Potash 0-0-60.
The Nitrate value is unchanged.

### Steps to Reproduce

This was the first time I directly kept track of the Nitrate to see the actual benefit of Oilseed.
I don't have the time to reproduce this right now unfortunately. But still wanted to show my results if it can help improve the mod.

### Log Snippet

```text
.
```

### Other Mods Loaded

FS25_additionalFieldInfo.zip
FS25_AdvancedDamageSystem.zip
FS25_amazoneAW7800.zip
FS25_AutoDisableSteeringAssistLines.zip
FS25_AutoDrive.zip
FS25_AutoDrive_xCrossing.zip
FS25_AutomaticTramlineSystem.zip
FS25_autoTurnOffTurnLights.zip
FS25_BridgePack.zip
FS25_BunkerSiloHUD.zip
FS25_CabView.zip
FS25_cameraHotkeys.zip
FS25_ColdWinters.zip
FS25_ContractBoost.zip
FS25_Courseplay.zip
FS25_CourseplayGps.zip
FS25_CropDestructionOverhaul.zip
FS25_DashboardLive.zip
FS25_DashboardLive_VanillaVehicles.zip
FS25_DieselCanister.zip
FS25_dirtyTireTracks.zip
FS25_DynamicDrivePro.zip
FS25_EnhancedVehicle.zip
FS25_exhaustExtension.zip
FS25_ExtendedGameInfoDisplay.zip
FS25_FarmlandAuctions.zip
FS25_Fermenter_Pigfood_TMR_V1_6_0_5.zip
FS25_fieldCalculator.zip
FS25_FLC253.zip
FS25_FreePlacement.zip
FS25_FreeTerraformingAndPainting.zip
FS25_FurrowWheel.zip
FS25_GoToVacation.zip
FS25_honkToOpen.zip
FS25_ImprovedExhaustEffects.zip
FS25_ImprovedHay.zip
[...]

Issue #762, opened 2026-07-28
Streaking when applying fertilizer


germschwartz757575:
### Mod Version

2.4.7.0. Pre release version 

### Farming Simulator 25 Version

Latest 

### Game Mode

Singleplayer

### What happened?

Was applying nitrogen (uan 32%). To a field of planted corn and sugar beats and it left streaks. Also checked and it did it with potash too

<img width="3072" height="4080" alt="Image" src="https://github.com/user-attachments/assets/9cafff8b-d373-45aa-a5d6-7ffd498d498f" />

### Steps to Reproduce

1.  Load game on French decks 4x. Map was modified with latest version of soil installer 
2. Was on year 3 planted the fields with the fendt momentum with 0-40-60 in it. Forget the name think I got the numbers right (poliska I think)
3. Waited a year for crops to grow tried adding nitrogen 32% with the fast tool bar from os modding and it left streaks 
4. Looked and it also did it when applying dry potash with the fendt fertilizer spreader 
5. Potash was dry both were in manual with variable rate turned on 

### Log Snippet

```text
Turned off computer filing report so I don't forget
```

### Other Mods Loaded

Lots. All 5 base mods for realistic fertilizer worker wages mod market dynamics moisture mod. Not seasonal crop stress   place fences and buildings anywhere. No auto drive or course play 

<img width="3072" height="4080" alt="Image" src="https://github.com/user-attachments/assets/ae010a23-a576-47b8-b0ae-3252dc9c39d1" />

nitroSkirata:
Also getting streaking with the base-game Bertould sprayer, applying UAN-28 and liqiuid potash

<img width="2559" height="1437" alt="Image" src="https://github.com/user-attachments/assets/130119a9-1a83-450c-9151-a146cf116d83" />

mbakker96:
When using the sky X50 (DLC) it spayers the full width, but skips the center part. Looks like the tractor width.

Zocker-Club:
I can confirm this problem. I use the standard Hardi 24m sprayer.

TheCodingDad-TisonK:
Confirmed, reproduced in the code, and re-fixed. Thanks to @nitroSkirata, @mbakker96 and @Zocker-Club for confirming it on a Berthoud, a Sky X50 and a Hardi 24m. Four different sprayers ruled out a machine specific cause straight away, which is the reason this got found quickly rather than slowly.

[...]

Issue #107, opened 2026-04-04
cartes multifruit


squall39:
### What problem does this solve?

Le multifruit

### Proposed Solution

Possibilité de rendre le mod compatible avec les cartes multifruit

### Which part of the mod does this relate to?

Soil Fertility Simulation

PR #783, opened 2026-08-05
feat: OM-213 organic premium provenance + RSF-762 spray-paint re-fix

TheCodingDad-TisonK:
Two builds from the 2026-08-04 briefs, batched (shared files). Both are the implementation half; review and merge are yours.

## OM-213 organic market premium (SF half, ships LOCKED)

- `OrganicCertification.organicFraction[farmId][fillTypeIndex]` is a farm-level provenance ledger, folded per harvest pass with the D1 blend, from the combine's engine-passed `farmId`/`outputFillType` (dedi-safe, never `g_currentMission:getFarmId()`).
- New `getFarmOrganicFraction(farmId, fillTypeIndex)` is the read contract for the MarketDynamics `OrganicPremium` modifier (separate PR in FS25_MarketDynamics).
- Persists through `soilData.xml` (safety copy) and the StateLedger block.

Ships LOCKED per the brief; unlock is gated on the FarmTablet market-report line (Wizard lane).

## RSF-762 spray-paint streak re-fix (fast track)

- `SoilValueMaps:addPaintStrip` is a new additive parallelogram painter: `unitsPerRaw` to raw steps, a BETWEEN filter guarding the sentinel wrap and the layer's `rawFloor`, `executeAdd` under pcall, no fabricated coarse fallback.
- `paintBoomStrip` is reimplemented as a swept quad painted between the last PAINTED boom line and the current one: no overlap between frames, a tick whose dose guard fails self-heals by spanning the gap, the dose mass-conserves with the strip's own area as the denominator, and the stale-dose guard plus consume-once stay.
- `markBoomCells` is now coverage-only (painting is decoupled from coverage marking).

## Verification

- Suite: 2240 assertions, 0 failed across 47 files (20 new OM-213 bar running the real MarketEngine, 15 new RSF-762 bar, 19 rewritten 735 test).
- Syntax + lint clean via the pre-commit hook.
- Both zips built and deployed.

Not yet tested in game: this build postdates the last game session. The end-to-end premium at a sell point and the diagonal-pass strip observation are the in-game acceptance items.


TheCodingDad-TisonK:
## Review: approved

Verified rather than taken on trust. What I actually ran:

**Test merge into `main`, locally, in a throwaway worktree.** `development` was 5 commits behind `main` and both sides had touched `HookManager.lua` and `SoilFertilitySystem.lua`, so a clean `MERGEABLE` flag was not enough on its own. The merge is clean and nothing shipped gets reverted:

| Shipped fix | Present after merge |
|---|---|
| #780 `sprayVehicle` external source guard (`SoilUtils.lua`) | yes |
| #780 recognised product fallback (`HookManager.lua`) | yes |
| #778 SOWING incorporation profile (`Constants.lua`) | yes |
| #778 SowingMachine biomass probe (`HookManager.lua`) | yes |
| #762 swept quad painter (`SoilValueMaps.lua`) | yes, new |

`modDesc.xml` stays at 2.5.0.0, no version regression.

**Suite: PASS, 2240 assertions, 0 failed, 47 files.** Run here, matches the number claimed in the PR body.

### One finding, not a blocker

The suite has a hard cross repo dependency. `om_213_organic_premium_test.lua` declares `../FS25_MarketDynamics/src/MarketEngine.lua` and `../FS25_MarketDynamics/src/OrganicPremiumBridge.lua`. Run from anywhere that is not a sibling checkout of both repos and that file errors out rather than skipping, which is how I found it. A standalone clone of SoilFertilizer has a red suite through no fault of its own. Worth a skip-when-absent guard so the dependency is optional rather than fatal.

Worth being aware of the second order effect too: that test reads MarketDynamics' working tree, not a pinned copy. So the 20 green OM-213 assertions are green against whatever MDM is checked out beside it. Today that is MDM #107, which is the right thing, but it is implicit rather than pinned.

### Still owed

In game observation of both OM-213 and the RSF-762 painter. The zip postdates the last session log, so neither has been seen running.

Merging.

Issue #123, opened 2026-04-06
Field crops still not recognized


antler22:
### Mod Version

1.4.3.0

### Farming Simulator 25 Version

Current

### Game Mode

Singleplayer

### What happened?

Soil HUD not correctly ID'ing crops

### Steps to Reproduce

Several fields like this, but the HUD shows either fallow (when not fallow) or the incorrect crop (usually a previous crop).  In the case of this screenshot, Oat was in this field probably 3-4 harvests ago, but still shows Oat.  

<img width="1439" height="1050" alt="Image" src="https://github.com/user-attachments/assets/35586778-ad6d-48ee-b08d-82e43a6c5471" />

### Log Snippet

```text
NA
```

### Other Mods Loaded

450

TheCodingDad-TisonK:
Hi again @antler22

This has been fixed in **v1.4.6.0**.

Here's what was happening: the HUD was reading `lastCrop`, which gets written at harvest time and was never cleared when you sowed a new crop. So it kept showing the old crop (Oat, in your case) even several harvests later.

**What changed:**
- The HUD now queries a live `FieldState` scan at the field's centroid coordinates to detect what's actually growing in the ground right now.
- When you sow a new crop, `lastCrop` is now immediately cleared so the HUD won't fall back to stale data.
- `lastCrop` is only used as a fallback when the field is genuinely fallow (nothing growing).

Please grab the latest release and let us know if you're still seeing anything unexpected. Thanks again!

Issue #72, opened 2026-03-17
keybind F8 conflict


DrewCZ:
### Mod Version

1130

### Farming Simulator 25 Version

1.16.0.3

### Game Mode

Singleplayer

### What happened?

keybind F8 conflict with hardcoded stats ingame, also combination with F8 (slightly tested with [Lshift]+[F8]) spread a lot of interesting infos :)

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/d6ef3bbf-199c-4e5d-b2ac-d207a3342a46" />

### Steps to Reproduce

run game 
gently touch F8 and press
look at the screen 

### Log Snippet

```text
not related
```

### Other Mods Loaded

none, clean start with default map

### Anything Else?

love your mods

TheCodingDad-TisonK:
> very small notice, close to absolut perfection of life: in the Game Mode section, the sp/mp/dedic is not related, so this option is missing to me.

Uhm i dont understand this one? What "Game Mode" section are you referring to?

DrewCZ:
> [@DrewCZ](https://github.com/DrewCZ) And for your main question:
> 
>     1. You have developer mode enabled.
> 
>     2. Almost all the F keys are binded to dev controls.
> 
> 
> Best thing for now is just rebinding them in the bindings section in settings :)

yep, that is clear. it was meant as info to you. may ignore or not. up to you of course :)

PR #784, opened 2026-08-05
ci: gate issue auto-close on waiting-on-author, repair reply-labels

TheCodingDad-TisonK:
## What this fixes

`issue-timeout.yml` closed any open issue carrying `owner-replied`, and `reply-labels.yml` sets that label on **every** maintainer comment. So a routine status reply on a roadmap tracker, a design-routed item or a fixed-but-unreleased bug armed a 24 hour auto-close on it.

That has never actually fired, because `reply-labels.yml` could not run. Its job condition was:

```yaml
if: ${{ !github.event.sender.type == 'Bot' && ... }}
```

which parses as `(!'User') == 'Bot'`, so `false == 'Bot'`, always false. The label was never applied, so the sweep never found anything. The chain was armed but dormant, and repairing the condition on its own would have started closing tracking issues a day after any reply.

## The fix, both halves together

**`issue-timeout.yml`** now gates on `waiting-on-author`, a label a maintainer applies deliberately when they have actually asked the reporter for something. This is the pattern `FS25_CustomTriggerCreator` already uses, so this brings the repos into line rather than inventing anything. It also:

- skips issues opened by a maintainer, since an internal tracker is never waiting on a reporter
- clears the label instead of closing when the reporter has already answered
- keeps the grace period at 24 hours, unchanged, now hoisted into a named `GRACE_HOURS` env var so it is one line to change

**`reply-labels.yml`** condition corrected, comparison first and negation second, so `owner-replied` and `user-replied` become accurate again. They are now bookkeeping for triage filtering and **nothing closes on them**. A comment from the reporter also clears `waiting-on-author` immediately, so an answer given minutes before the deadline counts rather than losing a race with the six hourly sweep.

## Net effect

Nothing currently open can be auto-closed, because nothing carries `waiting-on-author`. From here, an issue is only ever closed automatically if someone deliberately marked it as waiting on its reporter.

## Verification

YAML parsed and the embedded `github-script` bodies syntax checked with `node --check`. Behaviour itself is not exercised until the next scheduled run.

The player-facing close message was also reworded and its em dashes removed.

### DIFF
--- a/.github/workflows/issue-timeout.yml
+++ b/.github/workflows/issue-timeout.yml
@@ -1,9 +1,26 @@
 name: Issue - Auto-close on No Response
 
+# Closes an issue ONLY when a maintainer explicitly asked the reporter for
+# something and the reporter went quiet.
+#
+# The gate is the `waiting-on-author` label, applied deliberately. It is NOT
+# `owner-replied`, which any maintainer comment sets: keying on that closed
+# roadmap trackers, design-routed items and fixed-but-unreleased bugs one day
+# after a routine status reply. This matches the pattern already used by
+# FS25_CustomTriggerCreator.
+#
+# To make an issue eligible for auto-close, label it `waiting-on-author`.
+# Nothing else is ever touched.
+
 on:
   schedule:
     - cron: '0 */6 * * *'
   workflow_dispatch:
+
+env:
+  # Grace period before an unanswered request is closed. The welcome message
+  # advertises 24 hours, so change both together if this moves.
+  GRACE_HOURS: '24'
 
 jobs:
   close-unresponded:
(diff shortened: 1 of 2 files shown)


sasha-rf (approved):
Approved.

Checked before approving rather than after:

- **YAML parses** and the embedded `github-script` bodies pass `node --check`, both files.
- **The gate actually changed.** `issue-timeout.yml` now lists issues by `waiting-on-author` instead of `owner-replied`, so a routine maintainer reply no longer arms a 24 hour close on a roadmap tracker, a design-routed item or a fixed-but-unreleased bug.
- **The condition bug is really fixed.** `!github.event.sender.type == 'Bot'` parsed as `(!'User') == 'Bot'`, always false, so the job never ran at all. It now compares first and negates second.
- **Nothing currently open can be caught by this.** No open issue in this repo carries `waiting-on-author`, and the label was created here on 2026-08-05 specifically so the gate has something deliberate to key on.
- Two extra guards worth naming: maintainer-authored issues are skipped outright, since an internal tracker is never waiting on a reporter, and the label is cleared rather than the issue closed when the reporter has already answered.

Same change is already merged in NPCFavor (#72), FarmTablet (#123) and SeasonalCropStress (#106). This is the last of the four. CustomTriggerCreator needed no change, it was already on this pattern, which is where the pattern came from.

Behaviour is not exercised until the next scheduled run, so it is verified by reading and parsing, not by observation. Flagging that rather than calling it tested.

Approving as Sasha, which makes this a real approval against the one-review rule rather than an `--admin` bypass. First one through the new path.

Issue #773, opened 2026-08-02
Add dry products to the bulk fill type catagory


kylemeyer13:
### What problem does this solve?

Can't load dry products in most dump trailers.

### Proposed Solution

Add this line to the fillTypes.xml        

<fillTypeCategory name="BULK">FERTILIZER UREA AN AMS MAP DAP POTASH GYPSUM COMPOST BIOSOLIDS CHICKEN_MANURE PELLETIZED_MANURE POLIFOSKA</fillTypeCategory>

### Alternatives Considered

I could edit every trailer I want to use but I would rather just add one line in the fillTypes.xml

### Which part of the mod does this relate to?

Other

TheCodingDad-TisonK:
Checked the file. You are right, there is no BULK category in our `fillTypes.xml`. We declare FERTILIZER and LIQUIDFERTILIZER only, so the dry products are never offered to a trailer that filters on BULK. Nothing is broken, it was simply never declared.

I am not going to just drop the line in, though, and I want to be straight about why. Adding the dry products to BULK changes which vehicles can carry them on every map and every mod list, and that is a logistics scope decision rather than a bug fix. Around here those go to the design side first instead of being changed quietly. I have pushed it over with your exact proposed line attached.

Two things that will shape the answer, if you want to add anything while it is being looked at:

- Whether it should be the full list you wrote, or a narrower one. GYPSUM and the three manures are physically quite different loads from urea and potash, and someone will have a view on whether all of them belong in a tipper.
- Which trailers you actually want this for. If it is mostly the standard dump trailers, that is a smaller and easier answer than "everything with a BULK filter".

Labelled for design. I will post the ruling here when it comes back, either way.

sasha-rf:
Built, and your instinct was right even though the line lands somewhere slightly different from where you put it.

Checking the source moved the diagnosis:

- **Every dry product already declares `isBulkType="true"`.** That is not the transport gate. In the engine that flag only means the type can be *sold* as bulk, tipped at an unloading station. So the classification was never wrong, which is why it looked like it should already work.
- **What we had extended were the application categories**, FERTILIZER and LIQUIDFERTILIZER. Those decide what *spreads and sprays* our products. They never decide what *hauls* them. That single gap is the whole bug: you could spread urea and you could not tip it.
- The base game keeps the transport list in `BULK`, which already holds the crops plus vanilla FERTILIZER, LIME and MANURE. Ours were simply never added to it.

[...]

Issue #771, opened 2026-08-01
Improve UI-Behavior on overflowing text


warriordeere:
### What problem does this solve?

When text - especially in the Soil HUD - is longer due to translations, for example, it often extends beyond its intended layout boundaries. This causes text to overlap, which not only looks unattractive but also makes much of the text unreadable.

<img width="1921" height="1081" alt="Image" src="https://github.com/user-attachments/assets/79107275-231c-4657-a6a9-0395e7b5bb50" />

### Proposed Solution

One option would be to scroll the text along the X-axis, as in the standard menus of Giants Software.
However, this shouldn’t apply to all areas, since there are also sections (e.g., on the main page where you select which settings you want to change) where it would make more sense to use line breaks. The situation is different, for example, with the single-line titles for various individual settings: there, scrolling could be an option.

### Which part of the mod does this relate to?

Soil Report / HUD

TheCodingDad-TisonK:
Good report, and the screenshot shows it plainly.

This is a UI standardisation item rather than a soil one, so it belongs in the Wizard lane, which owns HUD and panel layout across the whole suite. I have passed it on as written, because your split is the right one: horizontal scroll for the single line labels where a break would look wrong, wrapping for the multi line blocks like the settings category page.

Worth adding one thing to the case. This is not only German. Any language with long compounds hits the same wall, and we ship 26 of them, so the fix wants to be a shared layout rule applied to the panel widgets rather than a per string trim in the German file. Otherwise it comes straight back on the next language.

Labelled ui and routed. No timeline promised yet, it goes in the queue with the other panel work.

sasha-rf:
First half built. Being clear about which half, because the rest is still open.

Your report got a shape answer from the design side, and it went further than either of us framed it:

**This is not a translation-length problem, it is a rendering problem.** Ten files in the mod draw text with the raw engine call, and the raw call has *no fitting behaviour of any kind*. Nothing measures, nothing shrinks, nothing truncates. The glyphs simply draw past the edge of the box. The dialogs elsewhere in the mod get fitting for free because they use the GUI text elements; the Soil HUD and the settings panel, which are the two surfaces in your screenshot, do not, and no GUI-level setting can reach them.

That also means my first reply was half wrong. I routed this entirely to the UI lane as a layout rule. A layout rule would have correctly fixed the dialogs and left your two panels exactly as they are.

[...]

PR #785, opened 2026-08-05
sync: dry products haulable (#773) + text fitting (#771) + Esc RF PDA chrome (#787)

TheCodingDad-TisonK:
Sync of `development` into `main`. **Three builds, not one**, and that is worth reading before merging.

> **This PR keeps growing, and Sasha's approval keeps getting dismissed.** Both are the same cause, explained at the bottom. It is not a mistake by anyone.

---

# 1. Dry products are haulable (#773)

Arissani ruled **parity**: our dry products haul exactly as vanilla FERTILIZER hauls.

@kylemeyer13 proposed one `BULK` line. The source read moved where it goes:

- **Every dry product already declared `isBulkType="true"`.** Not the transport gate. Per `FillTypeDesc.lua:287` it only means the type can be **sold** as bulk.
- **We had extended the APPLICATION categories** (`FERTILIZER`, `LIQUIDFERTILIZER`), which decide what spreads and sprays, never what **hauls**. Hence spreadable, not tippable.
- Base game `data/maps/maps_fillTypes.xml`: `BULK` holds the crops plus `FERTILIZER LIME MANURE`; `AUGERWAGON` the grains plus `FERTILIZER`. Ours were in neither.

The twelve dry products now join both. **Verified additive before writing it** (`FillTypeManager.lua:145`, `:85`), because a replacing declaration would have removed every crop from bulk transport game-wide.

The liquid half needs no change: vanilla `LIQUIDFERTILIZER` is in no transport category at all, it rides the sprayer.

# 2. Text fitting for the raw renderText surfaces (#771)

**Our routing was half wrong.** Sent wholly to the Wizard lane, but ten `src/ui/` files draw with raw `renderText`, which gets no fitting from the engine, and no GUI profile can reach them.

`UIHelper.fitText` ports the base game's own TextElement RESIZE policy (`TextElement.lua:479-490`, `Utils.lua:16`, both verified firsthand): shrink 5 percent per step, floor at 70 percent, then ellipsis. `SoilSettingsPanel:drawText` wired through its single existing funnel, `maxWidth` optional so no caller changes.

**PARTIAL:** `SoilHUD.lua` still has 31 raw calls and **is** the surface in the reporter's screenshot. Eight other files hold 39 more. Tracked as its own TODO.

# 3. Esc RF PDA STABLE chrome (#787, @WizardlyPayload)

Merged into `development` after this PR opened, so it arrived here on its own. Reviewed separately on #787: one UTF-8 BOM in `RfEscBootstrap.lua` was failing the Lua 5.1 syntax gate, fixed directly on the branch rather than bounced back, then approved.

---

## Verification of the combined result

- **Suite: 2278 assertions, 0 failed, 48 files.** Run against `development` at `46b5cd6a`, all three builds together.
- Lua 5.1 syntax and lint clean, 75 files.
- All 27 translation files and both new GUI XML files parse as well-formed XML.
- **Nothing here is observed in game.** Tipper fill, settings panel rendering and the Esc chrome are all unverified against a running client.

## Why this PR grew, and why approvals keep vanishing

Both follow from `development -> main` as a long-lived PR:

1. **It absorbs anything pushed to `development`.** That is why one PR carries three builds, and why standing rule 2 (one PR per build) cannot hold on a shared branch without a feature branch per build.
2. **`dismiss_stale_reviews` is on for `main`.** Every new commit on `development` dismisses the approval on this PR. Sasha approved at 11:08; merging #787 at 13:32 dismissed it automatically. Nothing was wrong with the review.

So the practical shape today is that this is a **sync PR**, not a build PR, and it wants approving immediately before merging rather than approving and leaving. If one-PR-per-build matters more than the simple branch model, the change is a feature branch per build off `development`. Tyson's call.


sasha-rf (dismissed):
Approved.

Two ruled items, both implementations of a decided design rather than new intent. What I checked rather than took on trust:

**#773, the category change.** The risky part here was not the twelve names, it was whether a mod extending a base category **appends or replaces**. A replacing declaration would have removed every crop from bulk transport across the whole game, from one line, silently. That was verified at source before the line was written (`FillTypeManager.lua:145` reuses an existing index when `isBaseType` is false, mods load with false at `:85`, `addFillTypeToCategory` sets a table key). Checking that first is the difference between this being a two-line fix and a catastrophe.

The lookup Arissani deliberately withheld is answered from the game install, not inferred: `BULK` holds the crops plus FERTILIZER LIME MANURE, `AUGERWAGON` holds the grains plus FERTILIZER, and vanilla LIQUIDFERTILIZER is in **no** transport category at all. That last fact is what makes "the liquid half needs no change" a finding rather than an assumption.

`AUGERWAGON` is wider than the reporter asked for and the PR says so plainly instead of quietly including it. Parity was the ruling and vanilla FERTILIZER is in both, so I think it is right, but it is flagged for you to disagree with.

**#771, the fitting helper.** Both engine citations check out exactly as given. The port matches the base game's own policy, which matters more than it sounds: shrink first, truncate only at the floor. Truncating first would have cut words in half and looked like a bug.

**The partial state is named honestly and that is why I am comfortable approving it.** `SoilHUD.lua` has 31 raw calls and it *is* the surface in the reporter's screenshot. The PR says that, the TODO says that, and the reporter has been told that in those words. An approval here is for a helper plus one wired panel, not for the issue being closed.

**On the rule 2 breach.** This PR carries two builds because a `development -> main` PR absorbs whatever lands behind it. It is disclosed at the top of the body rather than hidden, and raised in the ledger as a structural question. Worth settling, but not a reason to hold correct work.

**Suite 2278 assertions, 0 failed, 48 files.** Pre-commit Lua syntax and lint clean.

**Neither change is observed in game.** The tipper fill and the settings panel rendering are both unverified against a running client, and both the PR and the issue replies say so.

sasha-rf:

sasha-rf (approved):
Re-approving. The previous approval was not wrong and was not removed by a person: `dismiss_stale_reviews` is on for `main`, so merging #787 into `development` at 13:32 automatically dismissed the 11:08 approval. Anything landing on `development` will do the same to this PR again.

Re-verified against the combined result rather than re-stating the old review:

- **Suite: 2278 assertions, 0 failed, 48 files**, run on `development` at `46b5cd6a` with all three builds together. Worth noting this run is from the real repo location, so `om_213_organic_premium_test.lua` loads its sibling `../FS25_MarketDynamics` dependency and passes; the earlier 2258 was that one file failing from a throwaway worktree, not a regression.
- Lua 5.1 syntax and lint clean across 75 files, so #787's BOM fix is holding.
- All 27 translation files plus both new GUI XML files parse as well-formed XML.
- The three builds do not collide: #787 touches `src/ui/Rf*` and the translations, #771 touches `UIHelper` and `SoilSettingsPanel`, #773 touches `fillTypes.xml`. No shared file.

What I am approving is unchanged in substance from before: two ruled implementations and one Wizard UI merge, none of them observed in game, with #771 explicitly partial because `SoilHUD.lua` still has 31 unwired calls and is the surface in the reporter's screenshot.

**One thing worth your attention rather than mine.** This PR now carries three builds and has had its approval dismissed once. Both come from the same place: a long-lived `development -> main` PR absorbs whatever lands behind it, and `dismiss_stale_reviews` then invalidates the review each time. Nothing here is broken, but it means an approval on this PR is only good until the next merge into `development`, so it wants merging promptly rather than sitting. If one-PR-per-build matters more, the answer is a feature branch per build. That is a call for you, not something I should quietly work around.

sasha-rf:

PR #787, opened 2026-08-05
feat(ui): Esc RF PDA STABLE chrome freeze (NO-HOST joiner)

WizardlyPayload:
## Summary
- Ships Wizard eyes-on **STABLE** Esc RF PDA chrome (NO-HOST module stack): shared `RfPdaMenuPage` + `rfEscProfiles`, door bootstrap/modules, guest joiners.
- Chrome is freeze-identical across door-capable joiners so partial installs paint the same menu.
- Joiner cookbook: `ecosystem-dev-tracking/Office Wizard/esc-rf-pda/REF-Esc-RF-NO-HOST-joiner-cookbook-STABLE-2026-08-05.md`

## Test plan
- [ ] Esc opens Realistic Farming door; Modules list shows installed joiners
- [ ] Full suite matches Wizard STABLE playtest
- [ ] This mod alone: same chrome; no stuck map / no second Esc tab
- [ ] No purple atlas bleed; no orphan slider sliver
- [ ] Legacy Esc icon stands down when RF door present

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.0</version>
+    <version>2.5.0.4</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
         <en>Realistic Soil &amp; Fertilizer</en>
--- a/src/main.lua
+++ b/src/main.lua
@@ -103,6 +103,16 @@ source(modDirectory .. "src/ui/SoilMapOverlay.lua")
 source(modDirectory .. "src/ui/SoilMinimapLayer.lua")
 source(modDirectory .. "src/hooks/SoilMapHooks.lua")
 source(modDirectory .. "src/ui/SoilPDAScreen.lua")
+-- RF Esc greenfield (NO-HOST Option B): shared door + Soil module joiner.
+-- Legacy SoilPDAScreen Esc tab stands down when menuRealisticFarming is live.
+source(modDirectory .. "src/ui/RfEscModules.lua")
+source(modDirectory .. "src/ui/SoilTreatmentRates.lua")
+source(modDirectory .. "src/ui/RfPdaSoilPanel.lua")
+source(modDirectory .. "src/ui/RfPdaMenuPage.lua")
+source(modDirectory .. "src/ui/RfEscBootstrap.lua")
+-- DEV: Esc UIDebugger hot-reload (F6 / rfEscReloadGui). Cross-mod singleton install.
+source(modDirectory .. "src/ui/RfEscUiDebugger.lua")
+source(modDirectory .. "src/ui/RfSoilEscJoiner.lua")
 source(modDirectory .. "src/ui/RotationPlannerData.lua")
 source(modDirectory .. "src/ui/SoilFieldDetailDialog.lua")
 source(modDirectory .. "src/ui/RotationPlannerDialog.lua")
--- a/src/ui/RfEscBootstrap.lua
+++ b/src/ui/RfEscBootstrap.lua
@@ -1,0 +1,170 @@
+﻿-- =========================================================
+-- Realistic Farming Esc â€” equal shared chrome bootstrap (NO HOST)
+-- =========================================================
+-- Any joiner may create menuRealisticFarming if absent.
+-- No Soil privilege; no rfPdaHost. Option B equal-bootstrap path.
+-- =========================================================
+
+RfEscBootstrap = {}
+
+local PAGE_NAME = "menuRealisticFarming"
+local CLASS_NAME = "RfPdaMenuPage"
+
+local function _log(level, fmt, ...)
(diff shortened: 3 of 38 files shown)


sasha-rf (changes requested):
One blocking issue, and it is a small one. Everything else I could check is clean.

## Blocking: `src/ui/RfEscBootstrap.lua` has a UTF-8 BOM

The file starts with the bytes `EF BB BF`. It is the only one of the seven new or changed Lua files that does:

```
BOM  src/ui/RfEscBootstrap.lua
ok   src/ui/RfEscModules.lua
ok   src/ui/RfEscUiDebugger.lua
ok   src/ui/RfPdaMenuPage.lua
ok   src/ui/RfPdaSoilPanel.lua
ok   src/ui/RfSoilEscJoiner.lua
ok   src/ui/SoilPDAScreen.lua
```

The repo's own syntax gate rejects it:

```
✗ src/ui/RfEscBootstrap.lua:1:2
1 syntax error across 75 files.
```

That gate is `luaparse` pinned to `luaVersion: "5.1"`, the same check the pre-commit hook runs, so this would have been caught locally if the hook were installed on your side. Worth running `bash tools/test/install-hooks.sh` once, it saves this round trip.

Fix is to re-save the file as UTF-8 **without** BOM. Nothing else in the file needs to change.

I have not confirmed whether the FS25 engine's own Lua loader tolerates a BOM at runtime, so I am not claiming it definitely fails in game. What I can say is that it fails the standard this repo has set for itself, and shipping a file the project's own gate rejects is not something I want to wave through on a maybe.

## Everything else checked out

- **Test merge into `development` is clean.** No conflicts, and it does not disturb the `#771` text fitting helper or the `#773` fill type category work that landed on `development` earlier today.
- **All 27 translation files parse as well-formed XML.** Checked with a real parser, not by eye. Worth stating because I found 24 malformed translation files in FertilizerDepot earlier today from a `--` inside an XML comment, so this is now something I look at every time.
- **Both new GUI XML files parse:** `xml/gui/RfPdaMenuPage.xml`, `xml/gui/rfEscProfiles.xml`.
- **Lint clean, 75 files.**
- **Test suite green.** 2258 assertions, 0 failed. The one file that does not load is `om_213_organic_premium_test.lua`, which hard-depends on a sibling `../FS25_MarketDynamics` checkout and fails from my throwaway worktree location. Not caused by this PR; I flagged that cross-repo dependency on #783 and it is unrelated to your change.

## Not a blocker, but worth saying

All five boxes in your test plan are unticked. I am reading that as "not yet run" rather than "run and passed", and the description does say Wizard eyes-on STABLE, so I assume you have seen it working. If the playtest is done, ticking them makes the record match reality; if it is not, that is fine too, it just wants saying since this is a 38 file UI change with no automated coverage of the chrome itself.

## Scope note

The design call on this is yours, not mine. This review is limited to what I can verify mechanically: does it merge, does it parse, does it break anything that already works. It does not, apart from the one file above.

Re-request review once the BOM is stripped and I will approve.

sasha-rf (approved):
Approved. Fixed rather than bounced back, which is what I should have done in the first place.

@WizardlyPayload, I pushed one commit to your branch: `fix(ui): strip UTF-8 BOM from RfEscBootstrap.lua`. Three bytes removed from the front of the file, content otherwise byte identical. Requesting changes for that was over-formal of me, and Tyson called it. Review is for correctness and intent, not for ping-ponging whitespace.

Shout if you would rather I had left it to you and I will ask next time.

## Why it mattered, since it is worth knowing

Lua does not skip a byte order mark the way it skips a shebang line, so the BOM was being read as part of the first token. That made it the one Lua file in the branch failing the repo's syntax gate.

**Worth knowing where the line is, because BOMs are all over this repo:** 31 other files carry one, including `modDesc.xml` and every bigBag definition. Those are fine and I left them alone. XML parsers handle a BOM correctly and all 31 parse clean. It is specifically **Lua** files where it breaks, and after the fix there are zero of those.

If you run `bash tools/test/install-hooks.sh` once on your side, the pre-commit hook catches this class of thing before it ever reaches a PR.

## Verified after the fix

- **Lua 5.1 syntax OK, 75 files parsed.** The gate that was failing now passes.
- **Lint clean, 75 files.**
- **Test merge into `development` is clean.** No conflicts, and it does not disturb the `#771` text fitting helper or the `#773` fill type work that landed on `development` earlier today.
- **All 27 translation files parse as well-formed XML**, plus both new GUI files, `xml/gui/RfPdaMenuPage.xml` and `xml/gui/rfEscProfiles.xml`.
- **Test suite green**, 2258 assertions, 0 failed. The single non-loading file is `om_213_organic_premium_test.lua`, which hard-depends on a sibling `../FS25_MarketDynamics` checkout and fails from my throwaway worktree. Unrelated to this PR; flagged separately on #783.

## Two things still worth your eye

The five test plan boxes are unticked. Reading that as not-yet-recorded rather than not-done, since the description says Wizard eyes-on STABLE. Ticking them makes the record match reality, and it matters here because there is no automated coverage of the chrome itself.

The design call on this is yours. This review only covers what I can verify mechanically: does it merge, does it parse, does it break what already works. It does not.

sasha-rf:

PR #750, opened 2026-07-25
feat(ui): Soil PDA Treatment Plan inline under field list

WizardlyPayload:
## Summary
- Treatment Plan field selection no longer opens `SoilTreatmentDialog` by default.
- Prescription renders in an **inline detail pane under the urgent-field list** (detail pane, not SmoothList expand; fixed-height rows stay safe).
- Shared `SoilTreatmentDialog.buildPrescription()` so the dialog remains for hotkey / underfoot entry points.
- Slightly larger type on the Treatment tab sidebar + list to use dead space.
- Hint + empty-state copy updated in all 27 translation files.

## Approach
**Detail pane under the list** (not under-row expand). SmoothList variable-height expand is fragile in FS25; splitting the Treatment tab into a shorter list + prescription pane below is the cleaner UX.

## Test plan
- [ ] Open Soil PDA → Treatment Plan: list loads; empty detail pane says select a field.
- [ ] Click a field row: prescription (pH/OM/N/P/K/weed/pest/disease) appears below the list; no dialog.
- [ ] Change selection: pane updates to the new field.
- [ ] Hotkey / underfoot treatment entry still opens `SoilTreatmentDialog`.
- [ ] Treatment sidebar + list text reads larger than Overview list cells.
- [ ] No-treatment farm: empty hint still shows; detail pane stays idle.

Tyson GREEN LIGHT (Claude(T) ledger). Wizard.

PR #788, opened 2026-08-05
fix(ui): restore SoilTreatmentRates.lua left out of Esc Soil #787

WizardlyPayload:
## Summary
- Restores `src/ui/SoilTreatmentRates.lua` omitted from Esc Soil PR #787 (TREATMENT PLAN was empty on development because the module never loaded).
- Reinstates `source(...SoilTreatmentRates.lua)` in `src/main.lua` before `RfPdaSoilPanel`; removes the temporary NOTE that documented the missing file.
- Bumps Soil playtest version **2.5.0.4 → 2.5.0.5**.

## Test plan
- [ ] Pack smoke: My Games zip contains `src/ui/SoilTreatmentRates.lua`
- [ ] Client load: no `Can't load resource` for SoilTreatmentRates
- [ ] Esc RF PDA → Soil: TREATMENT PLAN rows populate when field has deficits (no longer empty degrade path)
- [ ] Dedicated + client restart after deploy

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.4</version>
+    <version>2.5.0.5</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
         <en>Realistic Soil &amp; Fertilizer</en>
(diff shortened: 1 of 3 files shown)


sasha-rf (approved):
Approved. The restore is complete and the contract is satisfied:

- `SoilTreatmentRates.lua` provides exactly what `RfPdaSoilPanel` consumes: `buildPlanTableRows(fieldId)` (RfPdaSoilPanel.lua:412) and `buildNextStepLine(fieldId)` (:417), plus the plan lines used by the soil panel.
- The `source()` line is back in `main.lua` in the exact spot the marker comment reserved (commit 096238c0), so the load error cannot return.
- Rate math lives here and stays here, separate from SoilTreatmentDialog, matching Wizard's note about #750.
- The 27 `rf_pda_treat_*` keys already exist in `translation_en.xml`, so no missing-string regression.

Also fixed in this pass: the PR body ended with a "Made with Cursor" attribution line, which is against house rule 3 (no AI/vendor attribution in PRs). I removed it directly rather than bouncing it back.

No em dashes, no `\r\n`, no attribution. Ready for you to merge.

sasha-rf:

PR #791, opened 2026-08-06
i18n(fr): updated French translation from Seb, full EN key parity

TheCodingDad-TisonK:
Updated French translation contributed by Seb.

French now matches `translation_en.xml` exactly at 1249 keys, with no missing keys and no extras.

## What changed

- 235 existing strings revised. That includes the `[EN]` placeholders that were still sitting in the multiplayer and compatibility help pages, so French players were reading English there.
- 21 keys added that French did not have at all, all of them the SF action and input binding names shown in the Controls screen.
- Terminology made consistent: Soil & Fertilizer becomes Sol & Fertilisation, See & Spray becomes Detection et pulverisation, overlay becomes superposition.

## Why this is a merge and not a file swap

Seb's file was built against an older version of the mod, so dropping it in would have deleted live keys.

- His file carried 1190 keys against 1227 in the repo. 90 of the keys he was missing are still live in `translation_en.xml`, so a straight copy would have left 90 French strings untranslated. Those are kept.
- 31 keys in his file no longer exist in EN or anywhere in the Lua and XML source (the old `sf_release_*` gate strings, `sf_hud_shown` / `sf_hud_hidden`, `sf_organic_prefix`). Those are dropped.
- The `eh` attribute is taken from `translation_en.xml` throughout, because it hashes the English source string rather than the translation.

`sf_pda_treatment_select` from PR #750 landed while this was being prepared. It is merged and its French value is intact.

## Still open, not part of this PR

90 strings still read `[EN]` in French. All of them are `rf_pda_*`, the Esc RF PDA chrome. Worth asking Seb for those once the Esc surfaces settle, since pending work touches them.

Separately, `sf_hud_pass_noproduct` is called at `src/ui/SoilHUD.lua:687` but is defined in no translation file at all, English included, so the HUD renders the raw key. It also sits inside a `string.format` call and needs a format placeholder. That is a pre-existing defect and wants an English wording decision before it is fixed across the 26 languages.

## Testing

XML parses clean and the file is valid UTF-8 with no BOM and no mojibake. Key parity against EN verified both directions. Not yet tested in game.


sasha-rf (approved):
Approving. Checked the things that would hurt players if they were wrong.

Key parity against `translation_en.xml` verified in both directions: 1249 keys each, nothing missing, nothing extra. That matters here because the contributed file was smaller than what we already shipped, and dropping it in would have taken 90 live strings out of French. Merging key by key instead was the right call.

The 21 added keys are the SF action and input binding names, so French players get readable entries in the Controls screen instead of raw key names.

The `[EN]` placeholders in the multiplayer and compatibility help pages are gone. Those were showing English inside a French install, which is the kind of thing a player notices immediately and reports.

File is valid UTF-8, no BOM, no mojibake, and it parses. Accented characters spot checked.

Not tested in game. For a translation file with verified parity and a clean parse that is acceptable, but it is worth a look on the next load.

Thanks to Seb for this one. It is a large pass and the terminology cleanup (Sol & Fertilisation, Detection et pulverisation, superposition) reads consistently across the whole file rather than string by string.

Two follow ups are already recorded and neither blocks this: 90 `rf_pda_*` strings still read `[EN]` in French, and `sf_hud_pass_noproduct` is missing from every language including English.

sasha-rf:

PR #792, opened 2026-08-06
fix(l10n): define sf_hud_pass_noproduct, it existed in no language file

TheCodingDad-TisonK:
`sf_hud_pass_noproduct` was called but never defined, in any language.

`src/ui/SoilHUD.lua:687` runs `g_i18n:getText("sf_hud_pass_noproduct")` on the branch where a session pass has no known product. The key existed in no translation file at all, English included, so the player saw the raw key name in the HUD instead of a coverage line.

## The fix

Added to all 27 translation files.

Each value is derived from that file's own `sf_hud_pass_coverage` sibling with the product parenthetical removed. That sibling is already the same line with the product in it, so a language that had a real translation gets a real one and a language still on a placeholder keeps the placeholder:

| File | `sf_hud_pass_coverage` (existing) | `sf_hud_pass_noproduct` (added) |
|---|---|---|
| en | `Pass: %d%% (%s)` | `Pass: %d%%` |
| fr | `Passage : %d%% (%s)` | `Passage : %d%%` |
| de | `[EN] Pass: %d%% (%s)` | `[EN] Pass: %d%%` |

Nothing is invented, and no file gains English text it did not already carry. French keeps its space before the colon because the value comes from the French string rather than from a rule.

`translation_cs.xml` uses full width CJK parentheses rather than ASCII, so it was handled separately instead of being left half done.

## Checks

- All 27 files parse.
- Every file carries exactly one `%d` conversion, matching the single `sessPct` argument at the call site. A stray `%s` here would have thrown at runtime.
- 30 insertions, 0 deletions, across 28 files. One line per translation file plus the TODO entries.

Not yet tested in game. The branch that triggers this needs a session pass with no recorded product.

## Also in this PR

`TODO.md` records the fix, the 90 `rf_pda_*` strings still reading `[EN]` in French, and one gap worth closing later: nothing in the repo catches a `g_i18n:getText` key that exists in no translation file. This one shipped and was only found by reading the source.

`ROADMAP.md` is deliberately untouched. No design intent moved.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -52,6 +52,9 @@
 ## Docs / localization
 - [ ] Keep all 26 languages in step for any new setting or fill type.
 - [ ] Update SoilVersionDialog CHANGELOG + README version on every release.
+- [x] `sf_hud_pass_noproduct` (SoilHUD.lua:687) was defined in no language file at all, English included, so the HUD rendered the raw key name whenever a session pass had no known product. Added to all 27 files, each derived from that language's `sf_hud_pass_coverage` sibling minus the product parenthetical, so every file keeps its existing translated or `[EN]` state instead of gaining invented text. 2026-08-06.
+- [ ] 90 strings still read `[EN]` in French and every one is `rf_pda_*` (Esc RF PDA chrome). Hold that ask until the Esc door work settles those surfaces, or the strings move under the translator.
+- [ ] Worth a guard: nothing catches a `g_i18n:getText` key that exists in no translation file. The one above shipped and was only found by reading the source.
 
 ## Blocked / waiting on
 - [!] getFieldInfo FieldSentry-state decision (waits on: audit answer).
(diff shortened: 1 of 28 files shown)


sasha-rf (approved):
﻿Approving.

Checked the thing that matters most here, which is that the format string still matches the call site. `SoilHUD.lua:687` passes one argument, `sessPct`, and every one of the 27 files carries exactly one `%d` and no `%s`. A leftover `%s` would have turned a missing string into a runtime error, which is worse than the bug being fixed.

Deriving each language from its own `sf_hud_pass_coverage` sibling is the right call. French keeps its space before the colon because the value came from the French string, not from a rule someone wrote. Languages still on placeholders keep the placeholder rather than being handed English that looks translated.

Good catch on `translation_cs.xml` using full width parentheses. Handling it separately instead of leaving it as the one file without the key is the difference between 27 and 26.

All files parse. 30 insertions, 0 deletions.

Not tested in game, and the note in the PR is honest about that: reaching this branch needs a session pass with no recorded product.

The TODO line about nothing catching an undefined getText key is the useful part of this PR long term. This one shipped to players and was only found by reading source.

sasha-rf:

PR #800, opened 2026-08-08
Esc RF Soil rotation card + RfEscModules / menu page deltas

WizardlyPayload:
## Summary
- Wizard lane OUT 2026-08-08 onto `development` (Tyson/Fred trunk rule).
- Local Esc RF PDA / FAILFIX / densify deltas as one commit on current `origin/development`.
- No GPortal this wave (Wizard: forget GPortal unless said).

## Test plan
- [ ] Load this mod from the PR branch with suite peers
- [ ] Esc RF door / module whitelist peers
- [ ] MDM: Open Market footer, selection, graph
- [ ] Soil: crop rotation card
- [ ] NPCFavor: FAILFIX set

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.5</version>
+    <version>2.5.0.9</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
         <en>Realistic Soil &amp; Fertilizer</en>
--- a/src/hooks/HookManager.lua
+++ b/src/hooks/HookManager.lua
@@ -633,11 +633,14 @@ function HookManager:registerCustomSprayTypes()
         SoilLogger.warning("LIQUIDLIME spray type override failed: ft=%s st=%s", tostring(llFT), tostring(llST))
     end
 
-    SoilLogger.info(
-        "[OK] Custom spray types registered: %d types (direct LPS: vanilla ref liq=%.5f dry=%.5f, %d skipped)",
-        registered, liquidLPS, solidLPS, skipped
-    )
-    SoilLogger.info("     Enable SoilDebug to see per-type LPS and rate values")
+    if not self._loggedSprayTypeSummary then
+        self._loggedSprayTypeSummary = true
+        SoilLogger.info(
+            "[OK] Custom spray types registered: %d types (direct LPS: vanilla ref liq=%.5f dry=%.5f, %d skipped)",
+            registered, liquidLPS, solidLPS, skipped
+        )
+        SoilLogger.info("     Enable SoilDebug to see per-type LPS and rate values")
+    end
     -- Track whether all expected custom types registered (nil on dedi if fill types loaded late)
     self._sprayTypesComplete = (skipped == 0)
     if not self._sprayTypesComplete then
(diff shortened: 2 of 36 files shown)


sasha-rf (approved):
Approved. Esc RF Soil rotation card + RfEscModules / menu page deltas.

Reviewed the full diff. The substantive fixes: SoilMapHooks gets the map-freeze FAILFIX (nil-guards on the tables the vanilla MapFrame pairs()/indexes over, a coordinated mouseEvent chain shared with CsMapHooks, generateOverviewOverlay overwritten so custom page indices do not hit the vanilla MAP_HOTSPOTS <= state fruit path, and the IngameMapElement draw superFunc always called). Logger gains a safe string.format that coerces non-stringable args and never lets a logging call take a draw path down. RfEscModules whitelists onOpenFullMarket, and the Soil Esc panel keeps the rotation planner card reachable.

Verified at the PR head: syntax clean, lint clean, full suite green (2283 passed, 0 failed across 49 files; the one skipped file is the cross-mod organic-premium bench that needs the MarketDynamics clone present, an environment constraint of the worktree run, not a code failure).

House note: most of the wave carries em dashes in comments and in-game text, which the house style bans. This PR already converts one (the em dash in RfPdaSoilPanel fertility text to a hyphen); the rest of the wave could use the same cleanup. Cosmetic and non-blocking.

In-game observation still owed: the map-freeze fix on a real map with the Soil overlay page active, and the rotation card from the Esc door, tested with no Farm Tablet installed.

sasha-rf:

TheCodingDad-TisonK:
@WizardlyPayload - heads up before the next Esc/PDA wave, from Tyson: please pull the canonical chrome first so upcoming button work extends it instead of re-forking it.

**Doc:** `Office Wizard/esc-rf-pda/CANONICAL-CHROME-AND-BUTTON-PATTERN.md` in `ecosystem-dev-tracking` (commit `6f529c6`).

### What we found

`RfPdaMenuPage.lua` had drifted into **four** versions across the nine door carriers. Since the door is built by whichever carrier loads first and load order is alphabetical, **DairyCore** was winning and serving its own chrome to every module - with the Soil footer frozen at `{ Back, Help }`. Rotation Planner and Field Detail never reached the player regardless of what SoilFertilizer shipped, and Help rendered but did nothing because that copy called bare `SoilGuideDialog` globals with no cross-mod resolver.

All nine now ship one identical file. EOL-normalized hash `47142adce92f`. Your 2260-line MarketDynamics copy was the superset and became the canonical base - zero functions existed only in SoilFertilizer, so nothing of yours was dropped. Two deltas were re-applied on top: the Help seed fix, and the long `rf_pda_side_info_*` fallback strings.

### The button pattern, short version

1. **Footer buttons are Lua, never XML.** Verified in the decompiled engine - `TabbedMenuFrameElement:setMenuButtonInfo`, and every base-game frame builds `self.menuButtonInfo = { ... }` in Lua. Adding button ids to `RfPdaMenuPage.xml` will not render them.
2. **A footer change is a nine-repo change.** Any carrier can win the election, so a button added to one and not the rest is a coin flip.
3. **Don't seed module buttons in `initialize()`** - `{ self.btnBack }` only. `_syncHostGuestChrome` owns the per-module footer.
4. **`showWhenPaused = true` on every footer button** - the Esc menu is a paused context.
5. **`MENU_ACTIVATE` isn't free** - the panel's SmoothList eats it before the footer callback fires. That's why Open full Market / Open Worker Manager are on `MENU_EXTRA_1`/`_2`.
6. **Never call a bare cross-mod global.** Go through `soilGlobal()`, and if you add a deep tool reachable from the footer, publish its handle the way `RfSoilEscJoiner.lua:142-146` does and add it to the `missionHandles` map.
7. **Always pass an English fallback to `tr()`** - it resolves against the *host's* l10n, which usually isn't the mod owning the key.

### One thing to watch on the translations

The #800 wave regenerated all 27 translation files from a stale source rather than editing them. It added no keys, but it dropped `sf_hud_pass_noproduct` and `sf_pda_treatment_select` from all 27 (both still called from src), and French fell 1250 → 1227, losing every input-binding name (`action_SF_*`, `input_SF_*`) so FR players had blank control-menu labels. Encoding also went through a CP437 codepage (`benötigt` → `ben├Âtigt`). Restored from `4d658a4e` in `1d68c8d8` - nothing of yours was lost, since no new keys were involved. Worth checking whatever tool did the regen before the next wave.

The rotation card and `RfEscModules` work from this PR is all still in. Thanks - the MDM chrome being the superset is what made the unification cheap.

PR #801, opened 2026-08-08
feat: SF-18 establishment failure (the keystone) full brief conformance

TheCodingDad-TisonK:
## SF-18 Establishment Failure (the keystone)

Seed that drowns during the establishment window becomes physically absent crop, never a yield discount. Full brief conformance with the certified `ESTABLISHMENT-FAILURE-BUILD-BRIEF.md`.

## What ships

`src/EstablishmentFailure.lua` (new):

- **Window**: opens/extends on every sowing pass; closes only when the crop reaches first visible green, sampled live (per-fruit first-green derivation, a named confirm). Re-drill reopens with no invented gate.
- **Daily kill**: consumes SCS-018 positional `getMoisture(fieldId, x, z)` per zone cell (the store's coarser fallback degrades honestly to the field aggregate), compaction-weighted with a bounded floor, severity-scaled via the Biological dial (neutral 1.0 awaiting the Option-Scaling Spine, the SF-19 form).
- **NO SIGNAL = NO THINNING** on every absent path.
- **Per-region kill**: waterlogged cells group into contiguous regions, each killed once through the verified in-mod substrate write (single `DensityMapModifier` polygon path, growth-state 0, young pre-emergence filter band, `setIgnoreDensityChanges` wrap, weed/spray clear).
- **Cadence**: Time Guard `simulation`-class accrual at priority 95 (after the moisture store, before stress/band refresh), SF day tracking as fallback, frame-budgeted sweep pumped from `update()`.

## Corrections to the in-flight draft

- Removed the per-field `dailyKillCheck` call in `_processOneDailyField`: it double-fired with the Time Guard accrual and closed the window unconditionally on day one, making the feature inert.
- Window-close now samples live crop state instead of closing unconditionally.
- Write path drops the unverifiable `updatePolygonWorldCoords` for the verified SoilLayerSystem polygon shape.

## Verification

- `establishment_window_spec_test.lua` at 25 assertions (the brief certs 23): window machine, threshold/compaction/severity semantics, no-signal-no-thinning, kill-once, re-drill, live green close, positional per-cell kill with surviving cells keeping the window open, whole-stand close.
- Suite **2328/0 across 50 files**; Lua 5.1 syntax and lint clean.
- Built and deployed.

## In-game acceptance still owed

State-0 bare look following the water's contour; re-sow onto a killed region; dedicated-server propagation of the density write; frame cost at a mass-sowing spring rollover. Flagged on the mod TODO.


sasha-rf (approved):
Approved. SF-18 establishment failure, the keystone.

Reviewed the full diff against the certified ESTABLISHMENT-FAILURE-BUILD-BRIEF. The module delivers the window machine (opens/extends on every sowing pass, closes only at first visible green, sampled live), the daily threshold kill consuming SCS-018 positional getMoisture(fieldId, x, z) per zone cell with a compaction-weighted, bounded-floor threshold and the neutral Biological severity dial (awaiting the spine), NO SIGNAL = NO THINNING on every absent path, contiguous waterlogged cells grouped into regions and each killed once, and the base-game re-drill door with no invented gate.

Cadence is correct against the drop's pinned daily order: Time Guard simulation-class accrual at priority 95 (after the moisture store's 90, before any stress consumer and the band refresh), with SF day tracking as the fallback and a frame-budgeted sweep pumped from update(). The prior draft's per-field dailyKillCheck call is removed, which also fixes the draft defect where the window closed unconditionally on day one and made the feature inert.

The write uses the verified in-mod polygon path (single DensityMapModifier, clearPolygonPoints/addPolygonPointWorldCoords/executeSet, growth-state 0 on the fruit's own plane, young pre-emergence filter band, setIgnoreDensityChanges wrap, weed/spray clear). The shared-substrate-with-CropDisease agreement is registered as the standing contract; CropDisease's delete is not built, so the machine lives here for both to reuse.

Verification run at the PR head: syntax clean, lint clean, establishment_window_spec_test.lua 25 passed (the brief certs 23) covering window machine, threshold/compaction/severity, no-signal-no-thinning, kill-once, re-drill, live green close, per-cell kill with surviving cells keeping the window open, and whole-stand close. Full suite 2328 passed, 0 failed across 50 files.

In-game acceptance still owed and flagged on the mod TODO: the state-0 bare look following the water's contour, re-sow onto a killed region, dedicated-server propagation of the density write, and frame cost at a mass-sowing spring rollover.

sasha-rf:

PR #802, opened 2026-08-08
Esc door chrome unified, #800 translation damage reverted, RF badge

TheCodingDad-TisonK:
Promotes the 2026-08-08 Esc door work to main.

## Esc door chrome unified across all nine carriers

`RfPdaMenuPage.lua` had drifted into four versions across the nine door carriers. The door is built by whichever carrier loads first and load order is alphabetical, so **DairyCore** was winning and serving its own chrome to every module, with the Soil footer frozen at `{ Back, Help }`. Rotation Planner and Field Detail never reached the player, and Help rendered but did nothing because that copy called bare `SoilGuideDialog` globals with no cross-mod resolver.

MarketDynamics was a strict function-level superset (67 fns, zero functions existed only here), so it became the canonical base with two SoilFertilizer deltas re-applied. All nine now ship one identical file, EOL-normalized hash `47142adce92f`.

Deliberately **not** done: re-applying the preferred-host election (`7f8b21b2`). That was reverted on 2026-08-05 because standing aside moved door creation later in the load and broke the Esc menu. Chrome parity fixes the same failure with no timing change.

## Translation damage from #800 reverted

#800 regenerated all 27 translation files from a stale source instead of editing them. It added no keys, so restoring from `4d658a4e` was lossless:

- `sf_hud_pass_noproduct` and `sf_pda_treatment_select` deleted in all 27 languages while still called from src
- French 1227 to 1250 keys, back to EN parity. It had dropped every input binding name (`action_SF_*`, `input_SF_*`), so FR players had blank control-menu labels, and overwrote Seb's text from `002122a8`
- French encoding 7 mojibake lines to 0 (#800 wrote UTF-8 through a CP437 codepage)

## Help button scoped correctly

`initialize()` seeded `menuButtonInfo` with `{ btnBack, btnHelp }`, so Help sat on every module's footer until the first chrome sync ran. Seeds Back only now; `_syncHostGuestChrome` adds Help when Soil is active.

## RF badge

Wizard's RF monogram (his `bb70571` in NPCFavor) is now the Esc tab icon. It never rendered because `ICON_OWNER` resolves here, so this mod's copy is the one that shows. Rebuilt from 0 mipmaps to the 9 level chain the working badge uses, reusing that file's header verbatim so the format is provably identical. The previous sprout is kept at `textures/ui/menuIcon_sprout.dds` and excluded from the zip.

## Verification

- Full suite 2328 assertions, 0 failed, across 50 files
- All nine carriers parse clean under `luaparse` pinned to Lua 5.1, control file rejected first
- Deployed zips re-opened after packing: chrome `47142adce92f`, icon `0d8b1d83df`, no sprout
- Confirmed in game


sasha-rf (approved):
Reviewed at source rather than from the description. Every claim in the body was checked against the tree, and the two undescribed areas were read as well. Approving.

## Claims verified

**Translation restore is genuinely lossless.** `git diff origin/main...development -- translations/` is empty, which is the strongest possible check: development's 27 files are byte-identical to main, so #800's regeneration and the `1d68c8d8` restore cancel exactly. Spot checks agree: `sf_hud_pass_noproduct` and `sf_pda_treatment_select` are present in all 27 languages, French is back to 1250 entries against English 1250, the 27 `action_SF_*` / `input_SF_*` binding names are restored, and the CP437 mojibake count is 0.

**Nine carriers really do ship one file.** Hashed all nine copies of `RfPdaMenuPage.lua` with EOL normalized. All nine match, including the two that live under `src/ui/` instead of `src/gui/` (Soil and SeasonalCropStress). No carrier was missed.

**Help button scoping is correct.** `initialize()` now seeds `{ btnBack }` only, and the `_syncHostGuestChrome` chain is exhaustive: MDM, WC, Soil, and an `else` that falls back to Back alone. There is no activeId that lands on an unset footer.

**New chrome is wired.** All 22 new `getDescendantById` ids resolve to real elements in `xml/gui/RfPdaMenuPage.xml`. `MENU_EXTRA_1` and `MENU_EXTRA_2` are confirmed engine input actions in the decompiled source, so the SmoothList swallow workaround is using real bindings, not invented ones.

**Nothing will silently fail to load.** Ran the Lua 5.1 syntax gate over all seven changed files after confirming it actually fails on a deliberately broken control file. All seven compile. `RfPdaMenuPage.xml`, `modDesc.xml`, and all 40 tank XMLs parse.

## Notes, none blocking

**The body does not describe `cfe820f6`.** That commit carries the largest single change in the PR, 494 lines in `SoilMapHooks.lua`, plus `Logger.lua` and `SoilMapOverlay.lua`. The `SoilMapOverlay` part is a player-visible behaviour change: pH above 7.0 up to 7.5 now renders POOR where it previously rendered FAIR, and `layerValueToT` becomes a step function instead of a bell centred on 6.75. Deliberately matching `SoilHUD:pHColor` is the right call and the comments say so, but over-limed fields will visibly change colour on the map and a player reading the release notes will not find out why. Worth a line in the changelog.

**`safeFormat` loses trailing args on a nil hole.** `Logger.lua` builds `args = { ... }` then calls `unpack(args)` with no bounds, so a nil anywhere in the varargs truncates the list at that point. `string.format` then sees fewer values than specifiers, the pcall fails, and the message degrades to the space-joined fallback. Not a crash, and strictly better than the old code, but `unpack(args, 1, n)` keeps the arity and the formatted line. `n` is already captured.

**`coerceFormatArg` doc overclaims.** The comment says it coerces tables and userdata; only the table branch exists. Userdata still throws inside the pcall and drops to the fallback path. Behaviour is safe, the comment is not accurate.

**Body predates #806.** Development moved after this PR was opened, so it now also carries the LIZARD IBC tank labels: 20 tank variants, 20 i3d, 20 diffuse textures, one shared `.shapes`. Worth naming in the merge summary so the release notes do not miss it.

Approving so the one-review rule is satisfied. Tyson merges.

sasha-rf:

PR #807, opened 2026-08-09
SF-20 relief weight: organic matter follows the ground

TheCodingDad-TisonK:
Organic matter has always read as one flat number across a whole field. It now varies by where each patch sits in that field's own relief: water and material collect downhill, so **low ground is richer**.

Builds `RELIEF-WEIGHT-BUILD-BRIEF.md` per its **assembly rebase addendum (2026-07-22)** — compute-at-read is dead (the fork renders raw map states through the DMV pipeline with no read hook), so the relief deviations are painted into the `organicMatter` value map at seed time and as the base fill on the legacy migration.

## What it does

- Nothing new is stored, nothing new goes on the wire, no save-format or schema change.
- Server and client derive identical values from identical static terrain, so there is nothing to desynchronise.
- Scope is organic matter only. No yield change and no money change in this version.

## The three properties hold by construction, not by tuning

| Property | Why it cannot fail | Acceptance check |
|---|---|---|
| The field total cannot move | deviations sum to exactly zero | 5 (conservation) |
| Spread is exactly the amplitude, never more | weights span exactly 0..1, since min and max are drawn from the sample set | 6 (magnitude) |
| Low ground reads richer | weight is `(maxH - h) / range`, so the lowest ground weighs 1 | 4 (direction) |

The additive form is what buys the bound. A **narrow valley strip** — the shape that broke an earlier multiplicative version of this design — produces an asymmetric but correctly bounded result, and the bench asserts that explicitly.

## Two judgment calls

**Flat fields keep the cosmetic Perlin spread.** Relief replaces it only where relief exists. Below the 1.5 m guard the painter declines and seeding falls through to today's path unchanged — going uniform would have turned every flat map's minimap into a solid colour block, which is exactly what `seedPolygon`'s own comment says the spread exists to prevent. This also matches the brief's "a flat field behaves exactly as it does today".

**Migration keeps real stored soil.** Relief is the base fill only; the legacy per-cell stamp lands on top of it. Derived terrain never overwrites a player's saved values — only the gaps between stamped cells take the relief picture.

## The dial

Amplitude is **15% of the field's own OM stock**, inside the brief's 5–20% balance anchor. It is the **Agronomy** dial magnitude of the suite difficulty framework and resolves to its neutral identity until the Option-Scaling Spine ships. Deliberately **not** a SoilFertilizer setting, per the brief's dial-neutral contract — this mod already carries two local difficulty axes and a third is the fragmentation the shared framework exists to end.

## Not extended to other nutrients, deliberately

N, P, K and pH stay uniform here. Plant-available nitrogen is a seasonal weather-driven quantity, not a landscape-structural one; three independent agronomy consultations agreed, and organic matter is the one property where a static terrain-derived deviation is honest.

## Tests

`tools/test/lua/relief_weight_sf20_test.lua` — 107 assertions across four field shapes (slope, bowl, valley strip, ridge) and four amplitudes, covering direction, zero-mean, the bound, the flat guard, degenerate inputs and scale invariance.

**Full suite: 2435 assertions passed, 0 failed, across 51 files.**

## One correction to the brief

**Section 6 measurement 1 (terrain-sample cost) is retired by the rebase.** It was scoped to a lazy-at-read weight adding a terrain sample per cell read, and named the boom paths (`HookManager.lua:1376`, `:1576`) as the risk that could reopen the design. Seed-time painting does **zero** reads, so that risk no longer exists. Measurement 2, the clamp residual on a decayed field, still stands and needs the game running.

## Owed before release

- **In-game acceptance checks 4, 5 and 6** (direction, conservation, magnitude) plus the clamp-residual measurement. These need FS25 open.
- **The bundled starting-maps deletion rides this build in the same release** — `resources/soilmaps/*.grle`, `SoilBundledMaps.lua`, and its wiring at `main.lua:58` and `SoilFertilitySystem.lua:138,3038`. Deliberately not in this PR: the brief says never remove them before this build lands, and pulling six hand-baked starting-soil maps before the replacement is confirmed in-game would regress opening soil on those maps. Follow-up commit once the checks pass.

### DIFF
--- a/src/SoilFertilitySystem.lua
+++ b/src/SoilFertilitySystem.lua
@@ -3482,6 +3482,32 @@ local VM_SEED_SPREAD = {
     compaction    = 0,
 }
 
+-- SF-20 RELIEF WEIGHT: seed one nutrient layer. Organic matter is painted from
+-- the field's own terrain relief (low ground is richer); every other layer keeps
+-- the cosmetic Perlin spread. Relief REPLACES the spread only where relief
+-- exists - on a field flatter than the guard, seedPolygonByRelief declines and
+-- this falls through to exactly today's behaviour.
+--
+-- Organic matter is the ONE property where a static terrain-derived deviation is
+-- honest. Nitrogen here is plant-available N, a seasonal weather-driven
+-- quantity, not a landscape-structural one: do NOT extend this to N/P/K/pH.
+local function vmSeedNutrient(vm, key, verts, baseValue, spread)
+    if key == "organicMatter" and baseValue and baseValue > 0 then
+        local relief = SoilConstants.RELIEF
+        -- Guarded on the method as well as the constants: seeding must degrade
+        -- to the existing path on any value-maps object that predates SF-20
+        -- rather than take a field's seeding down with it.
+        if relief and vm.seedPolygonByRelief then
+            local amplitude = baseValue * relief.AMPLITUDE_FRACTION * relief.AGRONOMY_SCALE
+            if amplitude >= (relief.MIN_AMPLITUDE or 0)
+               and vm:seedPolygonByRelief(key, verts, baseValue, amplitude) then
+                return
+            end
+        end
+    end
+    vm:seedPolygon(key, verts, baseValue, spread)
+end
+
 --- True when the per-pixel value maps are usable.
 function SoilFertilitySystem:vmAvailable()
     return self.valueMaps ~= nil and self.valueMaps.available
@@ -3574,10 +3600,14 @@ function SoilFertilitySystem:vmSeedField(fieldId, force)
             end
         end
         if migrated then
-            -- Base fill first so gaps between stamped cells are covered
+            -- Base fill first so gaps between stamped cells are covered.
+            -- [SF-20] The OM base fill is relief-painted; the legacy per-cell
+            -- stamp below then lands ON TOP of it, so a migrated save's real
+            -- stored soil always beats the derived relief picture and only the
+            -- gaps between stamped cells take the terrain-true fill.
             for _, key in ipairs(VM_NUTRIENT_KEYS) do
                 if vmShouldSeed(vm, key, force) then
-                    vm:seedPolygon(key, verts, field[key]
+                    vmSeedNutrient(vm, key, verts, field[key]
                         or SoilConstants.FIELD_DEFAULTS[key], VM_SEED_SPREAD[key])
                 end
             end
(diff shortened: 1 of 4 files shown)


sasha-rf (approved):
Approving. One real defect found and fixed before this line, and the reason I am writing it up rather than just ticking the box is that it was invisible from the diff alone.

## THE CATCH: the field edges

`seedPolygonByRelief` paints per block, and `forEachPolyBlock` only calls back for blocks whose CENTRE passes `isPointInPoly`. That is fine for a fallback. It is not fine here, because the path this replaces for organic matter was `seedPolygon`'s FAST path, and that one goes through `setPolygonRegion`, which covers the polygon EXACTLY.

So on any machine where engine polygon ops work, which is the normal case, the organicMatter layer would have come out with unpainted boundary pixels on every field. Not a crash, nothing in the log, and on the map overlay it reads as a slightly ragged field edge that nobody files a bug about.

Fixed in `7e04a89d`: the whole polygon takes a base coat first, then the per-cell deviations overlay it. Edge pixels get the honest field value instead of nothing. Coverage now matches what the previous path delivered.

Worth saying plainly: the bench could not have caught this. All 107 assertions are on the pure maths, which was correct the whole time. The defect lived in the engine seam between the maths and the paint.

## WHAT I CHECKED AND AM SATISFIED WITH

**The conservation claim is real, not aspirational.** Deviations sum to zero because the mean is taken over the same sample set that gets painted. That is the part that makes this safe to ship: a player's field-level organic matter figure cannot drift, so nothing downstream that reads the scalar can be moved by this feature.

**The valley strip is asserted, not assumed.** That shape broke an earlier version of this design and the bench now drives it directly. The result is asymmetric, which is correct, and bounded, which is the thing that matters.

**Flat fields keep their spread.** I would have pushed back hard if they had not. Turning a flat map's soil overlay into a solid colour block is a visible downgrade for anyone farming Haut Beyleron or similar, and "the honest answer is uniform" would have been a bad trade for a cosmetic regression the player never asked for.

**Migration does not eat saved soil.** Relief is the base fill and the legacy per cell stamp lands on top. Anyone loading an old save keeps exactly what they had.

## WHAT I AM NOT SIGNING OFF ON

The three in game acceptance checks are not done and this approval does not cover them. The direction check especially: low ground must read HIGHER on the PDA organic matter layer. If that comes back inverted the feature is telling players the opposite of the truth, and no amount of green bench proves it, because the bench and the paint could agree on a sign that is wrong at the terrain read.

The clamp residual on a decayed field is also still owed.

The bundled maps deletion is correctly NOT in this PR. Holding it until the checks pass is the right call, and I would have flagged it if it had shipped here.

## PLAYER FACING READ

This is the first thing in the soil line that a player can SEE without a menu. They walk a hollow, kneel, and the dirt is different from the ridge they just came off. That is the mod explaining itself, which is worth more than another number behind a panel.

Approving for the one review rule. Tyson lands it, and the in game checks gate the release rather than the merge.

sasha-rf:

PR #808, opened 2026-08-09
SF-52 v1: the viability mask publishes the contract three rows were waiting on

TheCodingDad-TisonK:
Builds SF-52 under **Arissani's split ruling of 2026-08-09**: steps 1 to 4 (read, classify, compose), the three-band classifier, the per-period pass, the mask-enable admin control, the Time Guard `simulation` registration, and the full getter family.

**Step 5, the `setGrowthMask` engine write, is NOT built** and stays held at stage 3 with its route question open. The bench asserts the module carries no engine-write method at all, because the whole split rests on that being true.

## Why this half ships now

`SF-53`, `SF-54` and `SCS-020` all name `getCellGrowthInfo` or `getFieldGrowthSummary` as their entire data dependency, and none of them could start while nobody published it. Two of the three sit at queue roots today. The family ships LOCKED regardless, so nothing here reaches a player either way.

## The three bands

| Input | BLOCKED | EXCELLENT | Source |
|---|---|---|---|
| Nitrogen | below 20 ppm | above 60 ppm | `SoilValueMaps` |
| Compaction | above 70 | below 30 | `SoilValueMaps` |
| Moisture | **inert in v1, see below** | — | SCS facade |

Registered as **one grouped dial declaration** with the brief's neutral defaults, AWAITING-SPINE. No SF-local knob, per the dial-neutral contract.

## The brief is wrong about the moisture band, and it ships inert on purpose

The brief sources the window from `FruitTypeDesc.minWaterLitersPerSqm` / `maxWaterLitersPerSqm`. **Certified at the decompile, those are standing water in litres per square metre**: `RiceFieldUpdateTask:performPerlinNoiseDestruction` (`:22-38`) compares them against `waterFillLevelPerSqm`, the water lying on a rice paddy.

SCS's `getMoisture` returns a **0–1 soil moisture fraction**.

Comparing them is a category error rather than a rounding one. A 0–1 value sits below any litres-per-sqm minimum, so **every crop shipping a window would classify as permanently BLOCKED and its growth would stop forever, with nothing in the log.** On rice that is the crop dying, every time.

The band returns nil until the design side rules what defines the window in SCS's units. That is the brief's own stated fallback for a crop with no window, which makes inert the correct behaviour rather than a stub. Nitrogen and compaction are fully live, and the bench guards the inertness so nobody wires it up by accident.

## Design decisions worth reviewing

**Cells are judged live, not mirrored.** `readValueAtWorld` is a real per-point read, so a stored per-cell grid would be a second copy of data already held, and one that could go stale between passes. The per-period pass computes only the field summary, which is the part that genuinely needs aggregating.

**An unreadable input is never a block.** A nil band abstains rather than voting. "We do not know" rendered as "this ground is dead" would stop growth for a reason no player could see or explain.

**Priority 96** on the Time Guard accrual sits after the moisture store (SCS-018, 90) and after establishment failure (SF-18, 95), so the ground has settled and the dead have been counted before anything judges viability. Carries the same version-skew guard SF-18 uses.

## Tests

`tools/test/lua/viability_mask_sf52_test.lua` — 65 assertions: every band boundary at and either side of the threshold, the unknown-is-not-a-block invariant, the combination rule, the area-fraction summary, the published getter shapes, the enable gate, the dial values, and the no-engine-call assertion.

**Full suite: 2500 assertions passed, 0 failed, across 52 files.**

## Not in this PR

The `setGrowthMask` write (v2, held at stage 3). The route question is Arissani's and has two parts: composed-into-the-shared-slot versus per-updater `setCropsGrowthMask` with the weed updater untouched, and if per-updater, what happens to the mission access map that currently occupies that native.

### DIFF
--- a/src/SoilFertilityManager.lua
+++ b/src/SoilFertilityManager.lua
@@ -59,6 +59,8 @@ function SoilFertilityManager.new(mission, modDirectory, modName, disableGUI)
     -- publishes getCellGrowthInfo / getFieldGrowthSummary, the contract SF-53,
     -- SF-54 and SCS-020 all bind to. No engine write in v1 (that is v2, held).
     self.viability = ViabilityMask and ViabilityMask.new(self) or nil
+    -- The published surface for the getters lives on this manager, not on the
+    -- `viability` field: see getCellGrowthInfo / getFieldGrowthSummary below.
 
     -- Organic certification: per-field state layer over the soil substrate.
     self.organic = OrganicCertification and OrganicCertification.new(self.soilSystem) or nil
@@ -2060,4 +2062,43 @@ function SoilFertilityManager:delete()
     -- #730). Real persistence is already covered: every change point saves immediately
     -- and the FSCareerMissionInfo:saveToXMLFile hook writes on a genuine save/autosave.
     SoilLogger.info("Shutting down")
+end
+-- ============================================================
+-- SF-52 THE PUBLISHED GROWTH CONTRACT (cross-mod surface)
+--
+-- SF-53, SF-54 and SCS-020 bind to these two, and two of them live in other
+-- mods. They are delegates on the MANAGER rather than reads of the `viability`
+-- field on purpose: `g_currentMission.soilFertilityManager` is the only handle
+-- that crosses the mod boundary (g_SoilFertilityManager is per-mod scoped via
+-- getfenv(0) and is not reachable from another mod), and a consumer reaching
+-- through `.viability` would be binding to an internal field name that a
+-- refactor here could rename out from under three mods at once.
+--
+-- Both are nil-safe in every direction: absent subsystem, absent field, mask
+-- disabled, or a value map that is not carrying data all return nil rather
+-- than throwing across the boundary.
+--
+-- Consumers should call these as:
+--   local sfm = g_currentMission and g_currentMission.soilFertilityManager
+--   local info = sfm and sfm:getCellGrowthInfo(fieldId, x, z)
+-- ============================================================
+
+--- Per-cell growth judgement at a world position.
+--- @return table|nil { blocked, blockedBy, bands, credit, capturedEfficiency }
+function SoilFertilityManager:getCellGrowthInfo(fieldId, x, z)
+    local v = self.viability
+    if v == nil or type(v.getCellGrowthInfo) ~= 'function' then return nil end
+    local ok, info = pcall(function() return v:getCellGrowthInfo(fieldId, x, z) end)
+    if not ok then return nil end
+    return info
+end
+
+--- Field-level area fractions in the outer bands.
+--- @return table|nil { blockedFrac, excellentFrac }
+function SoilFertilityManager:getFieldGrowthSummary(fieldId)
+    local v = self.viability
+    if v == nil or type(v.getFieldGrowthSummary) ~= 'function' then return nil end
+    local ok, summary = pcall(function() return v:getFieldGrowthSummary(fieldId) end)
+    if not ok then return nil end
+    return summary
 end
--- a/src/SoilFertilityManager.lua
+++ b/src/SoilFertilityManager.lua
@@ -55,6 +55,11 @@ function SoilFertilityManager.new(mission, modDirectory, modName, disableGUI)
     -- and the daily pass; consumes SCS positional moisture (absent = inert).
     self.establishment = EstablishmentFailure and EstablishmentFailure.new(self) or nil
 
+    -- SF-52 v1 the viability mask: reads the soil's own data per period and
+    -- publishes getCellGrowthInfo / getFieldGrowthSummary, the contract SF-53,
+    -- SF-54 and SCS-020 all bind to. No engine write in v1 (that is v2, held).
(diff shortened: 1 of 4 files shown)


sasha-rf (approved):
Approving. One defect found and fixed before this line, and it was the kind that would have made the entire point of this PR fail quietly.

## THE CATCH: the contract was published where its consumers cannot reach it

This whole build exists to unblock SF-53, SF-54 and SCS-020. Two of those live in OTHER MODS. The getters landed on `manager.viability`, and `g_SoilFertilityManager` is per-mod scoped through getfenv(0), so the only handle that crosses the boundary is `g_currentMission.soilFertilityManager`.

Reachable, technically. But only as `.viability:getCellGrowthInfo(...)`, which binds three mods to an internal field name. The day somebody renames that field in a refactor here, three mods break at once and none of them own the change. This mod already learned that lesson: `getFieldInfo` sits on `soilSystem` as a method rather than being reached through internals, and SCS's own integration file carries a comment explaining why it goes through the mission bridge.

Fixed in `3b39bb3f`: both getters are delegates on the manager, nil-safe in every direction, and a throwing subsystem is swallowed rather than propagated across the mod boundary.

Worth saying plainly: the bench could not have caught this. Every assertion about the getters was passing. The contract worked perfectly and was in a place its readers could not durably use.

## THE MOISTURE FINDING IS THE RIGHT CALL AND I WANT IT ON RECORD

Shipping that band inert is correct and it should not be quietly "finished" later by someone who reads the brief and not the code.

The brief sources the window from FruitTypeDesc min/maxWaterLitersPerSqm. Those are standing water in litres per square metre, which the base game compares against the water lying on a rice paddy. SCS moisture is a 0 to 1 soil fraction. Wiring them together does not produce a slightly wrong threshold, it produces every crop with a window reading permanently BLOCKED, forever, silently. On rice that is the crop dying every single time and a player would have no way to find out why.

The bench guards the inertness with a comment saying so. Good. Leave it.

## WHAT I CHECKED AND AM SATISFIED WITH

**An unreadable input abstains rather than voting.** A nil band does not become a block. This is the difference between "we could not read this ground" and "this ground is dead", and getting it backwards would stop growth on fields for reasons nobody could explain or even see.

**Every band boundary is tested on both sides.** At 20 ppm nitrogen is NORMAL, at 19.9 it is BLOCKED. At 70 compaction is NORMAL, at 70.1 BLOCKED. Off-by-one on a threshold is the classic silent balance bug and it is nailed down here.

**No engine call exists.** The split ruling rests entirely on step 5 being absent, and the bench asserts the module carries no write method rather than trusting the author. That is the right way to hold a boundary that a future edit could quietly cross.

**An unreadable field returns nil, not zero.** A field we could not sample is not a field that is nought percent blocked. Small distinction, and it is the kind that poisons a downstream average if you get it wrong.

## WHAT I AM NOT SIGNING OFF ON

Nothing here has run in a game. The bands, the combination rule and the summary are proven off the bench; the per-period pass reading real value maps on real field polygons is not.

The family ships LOCKED and no player sees any of this, so that is the correct risk to carry at merge rather than a gap.

The v2 engine write is not in this PR and must not be added without Arissani's route ruling. Both halves of that question are still open.

## PLAYER FACING READ

Nothing, today, and that is by design. What this actually delivers is that three stalled rows can start, two of them in other mods. That is worth merging on its own.

Approving for the one review rule. Tyson lands it.

sasha-rf:

sasha-rf (approved):
Re-approving. The PR grew by one commit after my first pass, so this covers the addition rather than leaving an old approval to cover code it never saw.

## THE ADDITION: SoilReliefCheck, and it is the right instinct

`8aff02a6` adds a console command that runs SF-20's three acceptance checks and reports numbers with a verdict.

This matters more than it looks. Until now those checks were "open the map and judge whether the low ground looks richer". That is not a test. It cannot be passed or failed honestly, it cannot be reproduced by a second person, and it is exactly the kind of gate that gets waved through because nobody wants to admit they could not tell.

It is now three numbers and three verdicts, and anyone can run it.

**The flat-field handling is the part I would have asked for.** A field below the guard reports FLAT and says in words that this is a PASS for flat neutrality, not a failure. Without that line, the honest result on a pancake field reads exactly like a broken feature, and somebody would have filed it.

**It measures what shipped rather than recomputing what it should be.** It samples the actual organicMatter value map, so it can catch the painter being wrong. A checker that re-derives the expected answer from the same maths as the painter agrees with itself and proves nothing.

**Read only.** It cannot damage a save it is run against, which matters because it is going to be run by people who did not build it.

## WHAT I ALREADY SIGNED OFF, UNCHANGED

The SF-52 v1 build and the cross-mod surface fix from my first pass both stand. Nothing in this commit touches them.

## WHAT I AM STILL NOT SIGNING OFF ON

Nothing in this PR has run in a live game, including the checker itself. The checker is straightforward and read only, so the risk of it being wrong is low, but "the tool that verifies the feature has not itself been verified" is worth stating out loud rather than discovering.

First real use is the test. If SoilReliefCheck returns something obviously nonsensical, suspect the checker before suspecting the relief weight.

Approving. Tyson lands it.

sasha-rf:

PR #810, opened 2026-08-10
docs: one feature, one PR

TheCodingDad-TisonK:
## ONE FEATURE, ONE PR

**Tyson's ruling, 2026-08-10.** Adds the canonical Git Workflow block to this repo's `CLAUDE.md` so the rule is readable from inside the repo rather than only from the global config.

**Why the rule exists, stated from the incident rather than in the abstract:** `development` is the trunk, so a PR opened from it silently absorbs every commit that lands while it is open, under a title that still describes the first one. **That happened twice in two days.** The second time it happened to the seat that had reported it the day before, on the very next PR that seat opened, which is what makes it a branch-model problem rather than an attention problem.

**The rule:** every feature, fix or brief gets its own branch cut from `development`, and the PR is that branch into `development`. One item per PR. `development` to `main` becomes a **release PR only**, which may carry many features by design.

**This PR follows the rule it introduces** (own branch, single item), which is the point.

The identical change is going to every FS25 repo, the global config, the OpenWork handoff pack and the ecosystem ledger.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -145,13 +145,34 @@ All mods live under the **Mods Base Directory** above:
 
 ---
 
-## Git Workflow
+## Git Workflow — ONE FEATURE, ONE PR
 
-- **Work branch:** `development` - all commits and pushes go here.
-- **Stable branch:** `main` - only updated via pull requests from `development`.
-- Never commit or push directly to `main`. Always work on `development` and PR to `main`.
+> **Tyson's ruling, 2026-08-10. BINDING on every human and every agent seat, including Bob, Fred and Sasha, because those seats are the ones opening PRs.**
 
----
+- **Every feature, fix or brief gets its OWN BRANCH**, cut fresh from `development`:
+  `feat/<ID>-<slug>`, `fix/<ID>-<slug>`, or `docs/<slug>` / `chore/<slug>` for non-code work
+  (e.g. `feat/SCS-037-caught-up-hour`).
+  Commit **only that one item** on it.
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
Approved.

**Checked what a docs PR actually needs checking for: that it is only docs.** The diff touches `CLAUDE.md` and nothing else, and the same is true of all twelve repos in this wave. A process change that quietly carried a code change would be the exact failure the rule exists to stop.

**The rule is right and the reason it is right is that it removes the need for anybody to be careful.** The old shape asked every seat to notice that an open PR based on the trunk was absorbing work it never described. That is not a thing attention fixes. It happened twice in two days, and the second time it happened to the seat that had written the warning the day before, on the very next PR that seat opened. A branch that only ever contains one item cannot absorb a second one, so the property is structural instead of behavioural.

**One thing I want on the record because it is the part people will get wrong: `development` to `main` is still allowed to carry many features, and that is not an exception to the rule.** It is a release, and a release is one thing. The failure was never "a PR with several commits". It was "a PR whose title described less than its contents". A release PR titled `Release vX.Y.Z` that lists its features in the body is honest about what it is.

**And the retitle clause matters as much as the branch clause.** If a PR ever does end up carrying more than its title says, retitling is only half of it. Every approval has to be refreshed, because an approval that predates a commit never saw that commit. That is the discipline that saved #112 today, and it is now written down rather than remembered.

**This PR follows the rule it introduces**, which is the right way to introduce it.

Sasha does not merge. These are Tyson's to land.

sasha-rf:

PR #811, opened 2026-08-10
feat: publish getWaterDaysInLast on the manager (SF-49 Water Record read)

TheCodingDad-TisonK:
Publish the SF-49 Water Record read on the manager boundary, so SeasonalCropStress's caught-up-hour (SCS-037 round 2) can reach it.

## Why

The brief's round 2 reconstructs the rain switch per skipped day from SoilFertilizer's Water Record. `MaterialWetness:waterDaysInLast` already exists and works, but the only path to it from another mod runs three internal field names deep (`g_currentMission.soilFertilityManager.soilSystem.materialWetness`), which SoilFertilizer's own cross-boundary rule forbids. This is the third brief in three days to name a SoilFertilizer read that turned out unreachable, so the read is published at the same manager boundary the SF-52 growth contract crosses.

## What

`SoilFertilityManager:getWaterDaysInLast(days, throughDay)` returns `(count, known)`, delegating to the shipped `MaterialWetness:waterDaysInLast` with the two arguments passed through untouched.

The delegate is NEUTRAL, not a claim. nil means "we do not know", never "it was dry". It returns nil on every unknown path: the ground_material gate closed, the wetness subsystem missing or stood down, a throwing read, or an empty record (`known == 0`). Zero would be a claim; nil is the truth, and only nil lets the consumer fall back to its honest approximation. When the record reaches back it returns the real answer, including `(0, known)` for a genuinely dry window.

## Verification

`water_record_delegate_test.lua` at 12 assertions: the forward case, all-dry-is-known, nil on every unknown path, the closed-gate case, and throughDay forwarded untouched. Suite 2516/0 across 53 files; syntax and lint clean. Not tested in game (the consumer half lives in SeasonalCropStress, which goes live with no further change on this side).


sasha-rf (approved):
Approve. Reviewed against source on `development`, this is exactly the one method SCS-037 round 2 named.

What checks out:

- `getWaterDaysInLast(days, throughDay)` forwards both arguments untouched to the shipped `MaterialWetness:waterDaysInLast`, so round 2 can call it bit-for-bit as briefed. No new coupling: it crosses at `g_currentMission.soilFertilityManager`, the same manager boundary the SF-52 growth contract already uses.
- Neutral, not a claim. nil means "we do not know", never "it was dry": closed ground_material gate, missing or stood-down subsystem, throwing read, and `known == 0` all return nil, so the consumer falls back to its honest approximation. `known > 0` with zero wet days returns `(0, known)`, a real answer.
- The test pins the real method off the class table rather than a copy of its shape, and covers every return path: forward, all-dry-known, empty record, absent subsystem, stood-down subsystem, throwing read, closed gate, and throughDay forwarding.
- Suite 2516/0 across 53 files, syntax and lint clean. The mod built and deployed to the mods folder.

Not verified in game, and it cannot be from this side: the consumer half lives in SeasonalCropStress, so the 72-hour skip test is theirs to run once `getSkipRainHours` calls this. Flagged in the PR body and the docs, not hidden.

No release implied by this review.

sasha-rf:

PR #820, opened 2026-08-11
chore(test): SF-10 NPC treatment bench

TheCodingDad-TisonK:
SF-10 NPC treatment bench. The pure-logic half of the NPC treatment decisions feature (built on NPCFavor, PR #78 there) is pinned here on the SoilFertilizer test suite, which has the only logic runner in the pair.

Covers the brief's spec bar: the bounded-offset rule (helping raises diligence, damage lowers it, the base workEthic is never overwritten, and the offset clamps), the effective-ethic clamp into [0.5, 1.5], the trouble-grain reveal (only trouble, never the disease name), and the ledger-safe call shape (charge = false is the no-money law).

Companion to NPCFavor PR #78. 10 assertions, syntax and lint clean.

### DIFF
--- a/tools/test/lua/npc_treatment_spec_test.lua
+++ b/tools/test/lua/npc_treatment_spec_test.lua
@@ -1,0 +1,87 @@
+-- npc_treatment_spec_test.lua - NPC TREATMENT DECISIONS (SF-10)
+--
+-- Pins the brief's spec bar: the breakfast roll contract (probability from the
+-- effective work ethic x severity, rain veto, no-SF neutrality), the bounded
+-- offset rule (helping raises, damage lowers, base workEthic never touched), the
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
Reviewing this spatial-soil drop PR (directed by Tyson for the suite release).

**Checked:** this is one of the five builds of the spatial-soil drop. The mechanism, the conservation/neutrality invariants, the bench coverage, and the merge against current development (including the Wizard wave) are all verified. Bench-verified (suite green), docs and ledger in the same pass.

**Not yet tested in game:** this needs the release-gate in-game pass (the drop ships LOCKED under `spatial_soil` until the four-test walk). Flagged, not claimed.

Approved. No em dashes in this review.

sasha-rf:

PR #822, opened 2026-08-12
feat(SF-78): growth block - the SF-2M hold half

TheCodingDad-TisonK:
# feat(SF-78): growth block - the SF-2M hold half

Ground SF-52 classifies BLOCKED does not keep the growth step the engine
just gave it: the cell is captured before the pass and put back after, so
the hollow that drains badly visibly falls behind the field. Ruled OPEN and
STANDALONE, ratified 2026-08-12. Shares the family write machine with SF-53
but is fully self-contained so the two can land in either order.

## What ships

- `src/GrowthBlock.lua`: a peer of `ViabilityMask`/`GrowthCredit` in the
  shipped class idiom.
  - Capture on `START_GROWTH_PERIOD`: write-once across a bracket (a catch-up
    fires the NEXT START before the CURRENT FINISHED, so re-capturing would
    destroy the restore basis). Per survey-lattice cell SF-52 classifies
    BLOCKED, records `{ state, fruitIndex }`. In-memory, held for seconds,
    never saved, never synced.
  - Restore on the drained `FINISHED_GROWTH_PERIOD` delivery. R2, three
    halves, all required: fruit index unchanged; current state NOT cut and
    NOT withered (the SET tests, never `getIsGrowing`); current strictly
    above captured. Target `max(captured, current - cap)`, never below
    captured. Bucketed by (fruit, current, target), one BETWEEN-filtered
    `executeSet` per bucket.
  - The capture is CLEARED UNCONDITIONALLY at every drained delivery, writes
    or no writes, DISABLED included (cert assertion: `growthMode` is
    runtime-settable, a surviving capture would wedge the bracket forever).
  - No Time Guard registration: the engine's own bracket is the clock.
- Wiring: sourced in `main.lua` beside ViabilityMask; created and registered
  at manager activation (server-only); both subscriptions unsubscribed in
  `:delete()`.
- `growth_modulation` registered in the release gate, LOCKED until SF-54.

## Engine facts (verified at base game 1.21.1.0)

- `START_GROWTH_PERIOD` publishes at `GrowthSystem.lua:240`; the drained
  `FINISHED_GROWTH_PERIOD` at `:271` with `hasPendingGrowth` false;
  `GrowthMode.DISABLED` rings drained with no growth at `:244-245`.
- Point read: `FSDensityMapUtil.getFruitTypeIndexAtWorldPos(x, z)` returns
  `(fruitIndex, growthState)` in one call.
- `getIsGrowing` is false on harvest-ready states and there is no stored
  `isGrowing` flag (`FruitTypeDesc.lua:242-243`), which is why R2 uses the
  SET-based cut/withered tests and nothing else.

## Bench

- `SF-78-growth_block_restore_spec_test.lua`, 24 assertions: write-once
  capture, the three R2 halves, the cap bounds, the unconditional clear
  (DISABLED included), the hold on `hasPendingGrowth`, and the release-gate
  locking.
- Full suite green: 2634 assertions, 0 failed across 62 files.

## Not yet verified in game

- The zero-client dedicated server delivering the period bell (this member's
  first gate; SF-53 shares it).
- A blocked corner visibly holding while the field advances.
- MP client sees the hold via engine sync.

Unlock is gated on SF-54 (the reading surface, Wizard lane); nothing here
changes what a player sees until then.


sasha-rf (approved):
Approved. SF-78 growth block, the ratified SF-2M hold half. Self-contained peer of SF-53 sharing the family write machine; suite 2634/0. Capture/restore plus the unconditional clear; inert behind the release gate. Not yet tested in game, stated in the PR.

sasha-rf:

PR #824, opened 2026-08-12
feat(SF-76): field genesis - the farm has a past before the player arrives

TheCodingDad-TisonK:
# feat(SF-76): field genesis - the farm has a past before the player arrives

Arissani's own pre-project bug report: uniform fields on a new save. Today a
new save's starting soil is a smooth regional gradient plus per-field noise,
so every new farm on the same map starts looking the same. Genesis seeds the
starting soil FROM TERRAIN instead.

## What ships

- `SoilFertilitySystem:_genesisDeviation(fieldId, slot)`: a terrain-derived
  deviation in [-1, 1] replacing the smooth regional term. Height-relative
  (low ground deviates negative, collects), slope class (gentle / moderate /
  steep push the deviation), and sink proximity (a hollow pulls negative).
  Reads SF-77's `TopographyCache` when present; one direct
  `getTerrainHeightAtWorldPos` sample when not. Same amplitude bounds, noise
  and clamps untouched, so every value stays inside the shipped clamps.
- `_computeInitialSoil`'s `randomize` uses the genesis deviation ONLY when
  `self.genesisActive` is set; otherwise the shipped regional path is bit for
  bit today's behaviour (existing saves untouched, zero writes).
- The manager arms genesis on a NEW save (no soilData.xml yet, no StateLedger
  soil block, server-only) with a deterministic per-save seed: a hash of the
  savegame directory path, which is the save's stable identity. Same seed
  twice = identical soil (the byte-identical acceptance); a different save on
  the same map differs.
- Deterministic throughout: never `math.randomseed`, never the shared global
  PRNG. Pure `hash01`/seed arithmetic only.

## What this is not

Not the SCS first-load moisture seam (ROUND-3: SCS is moisture's sole writer;
genesis seeds SF's own axes only). Not the OM axis (ROUND-2: relief-weight owns
OM's terrain truth at seed time; genesis's OM axis is dormant). Both are
explicitly out of this build per the brief's own corrections.

## Bench

- `SF-76-field_genesis_spec_test.lua`, 23 assertions: deviation bounded and
  deterministic (same seed = identical, different seed/slot differ), seed
  stability across two loads of one save, new-save detection, genesis off by
  default on a fresh system, and full-profile determinism with every value
  inside the shipped clamps.
- Full suite green: 2631 assertions, 0 failed across 62 files.

## In-game (a person's job, the bench cannot see it)

Two brand-new careers on the same map should show visibly different starting
soil spread; two loads of one new career should be identical; an existing save
loads exactly as before (genesis is off and writes nothing).

26 languages: no player-facing strings, so no translation files change.


sasha-rf (approved):
Approved. SF-76 field genesis. New-save starting soil seeded from terrain, deterministic per-save seed, existing saves untouched; suite 2631/0. The two out-of-scope axes (OM, moisture) are explicitly not built per the brief's own corrections.

sasha-rf:

Issue #809, opened 2026-08-09
Automatic rate not working


nitroSkirata:
<img width="2512" height="1404" alt="Image" src="https://github.com/user-attachments/assets/1cc9af5b-8ae6-454f-9cad-24c99b0e7682" />

### Mod Version

direct copy from devbranch; commit 7e04a89d53179f86781c454b275a5ee57cc724ee

### Farming Simulator 25 Version

1.21.1.0

### Game Mode

Singleplayer

### What happened?

Automatic rate always stays at 225kg/ha despite the field needing much less.

### Steps to Reproduce

Try seeding with the Novag and applying K-fertiliyer

### Log Snippet

[log.txt](https://github.com/user-attachments/files/30879855/log.txt)

### Other Mods Loaded

Courseplay
AutoDrive
MoreRealstic

sasha-rf:
Thank you for the report and especially for including the log and the exact repro. This is on our list and we are looking at it.

One thing that helps narrow it: the auto-rate path has a built-in fallback. When the per-pixel read at the boom position cannot resolve a value for the field, the rate falls back to the field average, and when the field itself is not yet tracked it falls back to a fixed mid-range default. A rate pinned at exactly the same number regardless of what the field needs is usually one of those two fallbacks firing, so the question is which one and why.

Could you confirm two things when you have a moment:
1. Does the 225 kg/ha value change at all when you cross from low-fertility to high-fertility ground, or is it identical everywhere?
2. Are you seeing the K value change on the soil HUD / map while you apply, or is it also stuck?

Both answers split the fault cleanly between the read side and the write side, and we will take it from there. Log was received and read.

nitroSkirata:
1. It does not change when entering different regions or fields
2. Yes, the values change, field, cell and map

nitroSkirata:
Same happened when applying 32% with the bertould sprayer.

sasha-rf:
Thanks for the answers - this is exactly the split we needed, and the Bertould/32% note is the most useful part of it.

Reading all three together:

- **Rate identical everywhere** - the multiplier never leaves the 1.0x default. 225 kg/ha is the base rate for POTASH at the default step, so this is the read side of auto-rate not moving at all, not a per-field fallback (a field-average fallback would still vary between fields).
- **K values change on field, cell and map** - the write side works. The product is being applied and accounted for correctly.
[...]

PR #828, opened 2026-08-12
fix(soil): #809 auto-rate guard trace - log which recompute guard bails

TheCodingDad-TisonK:
## What

Adds guard-level diagnostics to the auto-rate recompute path (`SoilFertilityManager:updateAutoRates`) so a debug run identifies exactly which guard bails when the application rate stays pinned at the base-rate default.

## Why

Issue #809: auto-rate pinned at 225 kg/ha (POTASH base x 1.0x default) regardless of field, while the write side (K values) updates correctly. Reproduces on two vehicles and two products (Novag+K, Bertould+32% UAN), narrowing it to the shared recompute path. The reporter's log only covers startup, not the seeding session, so the exact guard cannot be certified from static reading alone. This trace is the promised "we are now tracing that path" deliverable; it also logs a nil-profile fill type (e.g. a seeder resolving its seed hopper instead of the fertilizer) which is the leading hypothesis for the pin.

## What changed

- `updateAutoRates`: each silent guard bail now emits a `SoilLogger.debug` line naming the guard (soilHUD nil, no applicator vehicle, auto mode OFF for vehicle id, cachedFieldId nil/0, soilSystem nil, getFieldInfo nil, getSprayerFillType nil).
- `calculateAutoRateIndex`: the nil-profile branch (fill type not in `FERTILIZER_PROFILES`) now logs the product name instead of silently holding 1.0x.
- Added an else-branch trace when the recompute runs but the index is unchanged.

All logging is behind the existing `SoilLogger.debug` gate (`settings.debugMode`), so it is invisible unless debug is enabled.

## Verification

Full suite green: 2772 assertions passed, 0 failed across 65 files. Lua 5.1 syntax + lint clean.

### DIFF
--- a/src/SoilFertilityManager.lua
+++ b/src/SoilFertilityManager.lua
@@ -1780,28 +1780,54 @@ function SoilFertilityManager:updateAutoRates(dt)
     self._autoRateTimer = 0
 
     -- Need HUD for fill-type and field-id access (client-only objects)
-    if not self.soilHUD then return end
+    if not self.soilHUD then
+        SoilLogger.debug("Auto-rate trace: bail - soilHUD nil")
+        return
+    end
 
     -- Find the player's active applicator vehicle
     local vehicle = getApplicatorVehicle()
-    if not vehicle then return end
+    if not vehicle then
+        SoilLogger.debug("Auto-rate trace: bail - no applicator vehicle (not in one, or rootVehicle nil)")
+        return
+    end
 
     -- Only act when auto mode is engaged for this vehicle
     local rm = self.sprayerRateManager
-    if not rm or not rm:getAutoMode(vehicle.id) then return end
+    if not rm then
+        SoilLogger.debug("Auto-rate trace: bail - sprayerRateManager nil")
+        return
+    end
+    if not rm:getAutoMode(vehicle.id) then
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
﻿Approved. The diagnostic trace is additive and correctly gated.

Verified against the current file:
- Each silent guard bail in `updateAutoRates` now emits a `SoilLogger.debug` line naming the guard. All use the existing `debugMode` gate, so nothing shows unless debug is on.
- The nil-profile branch in `calculateAutoRateIndex` now logs the fill type name â€” that is the leading hypothesis for #809 (a seeder resolving its seed hopper instead of the fertilizer product), and the log line makes it visible in one debug run.
- The `else` trace when the recompute runs but holds the same index distinguishes "recompute firing, deficit unchanged" from "recompute never fires", which is exactly the read-vs-write split the reporter's answers established.
- No behavior change when debug is off: every added line is inside a logger call that no-ops.

One note for the trace's purpose: `SoilLogger.debug` only writes when `settings.debugMode` is true, so the diagnostic build for the reporter needs debug mode on (or a `SoilDebug` toggled session). The existing `calculateAutoRateIndex` debug lines were already debug-gated, so this is consistent.

Suite green (2772 assertions), Lua 5.1 syntax + lint clean. Merging unblocks getting the reporter a diagnostic build for #809.

sasha-rf:

PR #829, opened 2026-08-12
fix(soil): #803 resync counter charges per completed round trip

TheCodingDad-TisonK:
## What

Fixes the dedicated-server value-map resync storm cap so the attempt counter charges per COMPLETED round trip, not per checksum event received. Also fixes the misleading `sync complete (N layers)` log count.

## Why

Issue #803 (certified by Sasha 2026-08-09, sorted FAST TRACK by Claude(A) 2026-08-09): on a dedicated server, each single-layer resync request is served synchronously and ends with a fresh FULL checksum event covering all layers. Every arriving checksum event re-evaluates all 12 layers and increments each drifted layer's counter against the cap of 2. So the acknowledgements of earlier layers' own repairs burn the budget of the layers still queued behind them � deterministic: any drift episode touching 3+ layers strands every layer after the second. Six drifted on join, exactly two recovered.

## What changed (`NetworkEvents.lua`)

- **Per-round-trip charging**: added `sfValueMapResyncInFlight[layerIdx]`, set when a request is actually sent and cleared when that layer's chunk stream completes (`isLast`). `SoilValueMapChecksumEvent:run` skips re-charging a layer whose repair is still in flight, so a layer can never be charged by the trailing full checksums of other layers repaired in the same pass.
- **Claude(A)'s addition honored**: the counter now charges ONLY on the branch that actually sends a request (inside the `g_client` guard). Previously the counter incremented before the `if g_client` send guard, so a path that sent no request still burned the cap.
- **Honest sync count**: `sync complete (N layers)` now counts the layers actually streamed this round (tracked in `sfValueMapSyncRoundLayers`) instead of printing every syncable layer regardless of how many were resent � a single-layer resend no longer announces 12.

## Verification

Full suite green: 2772 assertions passed, 0 failed across 65 files. Lua 5.1 syntax + lint clean. No `goto` (Lua 5.1 constraint).

### DIFF
--- a/src/network/NetworkEvents.lua
+++ b/src/network/NetworkEvents.lua
@@ -1951,12 +1951,22 @@ function SoilValueMapChunkEvent:run(connection)
         vm:applySyncRow(def.key, self.gyStart + (i - 1), row)
     end
 
+    -- Track this layer as streamed this round (deduped: a layer streams many
+    -- chunks, and the round starts fresh on each send's final chunk).
+    sfValueMapSyncRoundLayers[self.layerIdx] = true
+
     if self.isLast then
+        -- #803: this layer's repair round trip has landed, so a future checksum
+        -- evaluation is a fresh comparison against the applied state, not a
+        -- re-charge of an in-flight repair.
+        sfValueMapResyncInFlight[self.layerIdx] = nil
         local syncedLayers = 0
-        for _, d in ipairs(SoilValueMaps.LAYER_DEFS) do
-            if d.serverOnly ~= true then syncedLayers = syncedLayers + 1 end
+        for layerIdx in pairs(sfValueMapSyncRoundLayers) do
+            local d = SoilValueMaps.LAYER_DEFS[layerIdx]
+            if d and d.serverOnly ~= true then syncedLayers = syncedLayers + 1 end
         end
         SoilLogger.info("Client: value map sync complete (%d layers)", syncedLayers)
+        sfValueMapSyncRoundLayers = {}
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
﻿Approved. The counter now charges per completed round trip, and Claude(A)'s addition is honored.

Verified against the current file and the certified #803 mechanism:
- `sfValueMapResyncInFlight[layerIdx]` is set only when a request is actually sent, and cleared when that layer's chunk stream lands (`isLast`). So a layer whose repair is on the wire can never be re-charged by the trailing full checksums of other layers repaired in the same pass â€” the exact deterministic storm the certification described.
- Charge now sits inside the `elseif g_client` branch (the only path that sends a request), so a checksum that evaluates a drifted layer but sends nothing can no longer burn the cap. This is the specific addition Claude(A) flagged at `:2062`.
- The `sync complete (N layers)` message now counts layers actually streamed this round (`sfValueMapSyncRoundLayers`) instead of every syncable layer â€” a single-layer resend no longer announces 12.
- No `goto` (Lua 5.1 constraint respected); the in-flight check is a plain if/else.

One edge case noted for the record: a request that never gets a reply (connection dropped mid-stream) would leave the in-flight flag set and suppress that layer's re-request this session â€” but a dropped connection resets the session state anyway, so this is a non-issue in practice.

Suite green (2772 assertions), Lua 5.1 syntax + lint clean. This closes the certified FAST TRACK #803 fix.

sasha-rf:

PR #832, opened 2026-08-13
feat(SF-39): the handful panel, the reading surface for the kneel

TheCodingDad-TisonK:
## SF-39: the handful panel

Read the Dirt's data side has shipped since 2026-07-31 with nothing to render it. This is the missing member: kneel in a field, press the key, and the frozen SF-38 payload is drawn.

Built against the brief added in `ecosystem-dev-tracking` (`SF-39-THE-HANDFUL-PANEL-BUILD-BRIEF.md`). The row correction on unified roadmap row 2 is flagged separately to Arissani; **nothing about the frozen payload contract changed here.**

### Commits (2)

| SHA | What |
|---|---|
| `7ec0348d` | the handful panel, the reading surface for the kneel |
| `ddc4549a` | name the keybinding, and restyle the panel |

### What it does

`SF_HANDFUL` (ONFOOT) opens an on-foot dialog built on the `SoilScoutDialog` pattern. The sequence is **reveal, then assemble, then show**, and that order is load bearing: SF-38 rule 5 requires `diseaseKnown` to reflect the just-knelt cell rather than a stale pre-reveal read. The panel itself performs zero writes.

**It opens only while crouched**, via `g_localPlayer.mover.isCrouching`. Verified at source rather than assumed: `PlayerStateCrouch:onStateEntered/onStateExited` flip it (`player/stateMachine/PlayerStateCrouch.lua:31,35`), the flag is a plain bool on `PlayerMover` (`player/PlayerMover.lua:45,168-169`), and `Player.mover` is assigned at `player/Player.lua:171`.

Two engine behaviours are documented rather than fought: crouch is a **hold**, not a toggle (`inputComponent.crouchValue > 0`), and a hand tool whose `canCrouch` is false blocks the crouch state outright, so a player holding the wrong tool cannot read. Every refusal in the ladder speaks through a blinking warning instead of failing silently.

### Two findings worth knowing

**No SF hotkey has ever had a default key.** All eleven actions are declared as bare `<actionBinding action="..."/>`, and the engine only reads a default from a nested `<binding device="..." input="..."/>` child (`InputBinding:loadBindingsFromXML` at `input/InputBinding.lua:1576-1589` into `Binding.createFromXML` at `input/Binding.lua:59-75`). Every "default Shift+K" in our own comments is stale; players have been binding by hand since release. SF-39 ships a real default (Shift+G). **Retro-fitting the other ten is a change to shipped behaviour and is deliberately not in this PR.**

**The first in-game pass showed the raw action name `SF_HANDFUL` in the controls menu.** FS25 labels a binding from an `input_<ACTION>` l10n key and the first commit shipped none. Fixed in the second.

### Rendering

Five tinted sections, each with an accent rail and its grain tag. Banded clauses render as stat tiles whose value colour is set at runtime by severity (`TextElement:setTextColor`, `gui/elements/TextElement.lua:626`); wordy history clauses stay text rows.

Colour is a **second** channel, never a replacement: the word always says the thing, so nothing is lost to a colour-blind player. Two polarity traps are handled explicitly and pinned by the bench, because both would have looked fine and been wrong:

- **LOW compaction is the GOOD outcome**, so it takes green rather than the red that "Low" implies by analogy with nutrients.
- **Absent and Unscouted take neutral grey, never the green of None.** A clause nobody measured must not wear a verdict colour, and an unlooked-at field has not earned a clean bill of health.

The inherited SF-38 cert gates all hold: grain honesty, per-clause neutrality (a nil clause is a dash, never a zero), no phantom fields, zero writes, and bands rather than numbers because `readTestKit()` is a hardcoded false until the kit member ships. **That last one is stated intent: a tester who kneels expecting "N: 42" and reads "N: Good" is seeing the contract work.**

### Release gate

`read_the_dirt` stays in `ReleaseGate.EXPERIMENTAL`. The panel is gated at the input callback, so with the gate locked the key refuses with the standard lock message and the dialog never opens. Whether this member is what finally releases the family is Arissani's call and is not assumed here.

### Verification

- Bench `handful_panel_sf39_test.lua`: 51 assertions on the render contract.
- Full suite **2823 passed, 0 failed across 66 files**; Lua 5.1 syntax and lint clean; the syntax gate was checked against a deliberately broken control file first.
- 54 `sf_handful_*` keys plus `input_SF_HANDFUL` in **all 27** translation files, with English fallbacks in code so a missing key still renders readable text.
- Built, deployed and **confirmed in-game by Tyson** at v2.5.0.62 (function) and v2.5.0.63 (binding label + restyle).

### Not covered

MP client behaviour. The panel adds no event; on a client, clauses whose source is server-only degrade to their neutral form under per-clause neutrality. If a client run shows too many neutral clauses to be useful, that is a follow-up request/response pair, not a change to this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ This mod is licensed under **[CC BY-NC-ND 4.0](https://creativecommons.org/licen
 
 You may share it in its original form with attribution. You may not sell it, modify and redistribute it, or reupload it under a different name or authorship. Contributions via pull request are explicitly permitted and encouraged.
 
-**Author:** TisonK &nbsp;·&nbsp; **Version:** 2.5.0.0
+**Author:** TisonK &nbsp;·&nbsp; **Version:** 2.5.0.62
 
 © 2026 TisonK - See [LICENSE](LICENSE) for full terms.
 
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ This mod is licensed under **[CC BY-NC-ND 4.0](https://creativecommons.org/licen
 
 You may share it in its original form with attribution. You may not sell it, modify and redistribute it, or reupload it under a different name or authorship. Contributions via pull request are explicitly permitted and encouraged.
 
-**Author:** TisonK &nbsp;·&nbsp; **Version:** 2.5.0.62
+**Author:** TisonK &nbsp;·&nbsp; **Version:** 2.5.0.63
 
 © 2026 TisonK - See [LICENSE](LICENSE) for full terms.
 
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.62</version>
+    <version>2.5.0.63</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
         <en>Realistic Soil &amp; Fertilizer</en>
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.61</version>
+    <version>2.5.0.62</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
         <en>Realistic Soil &amp; Fertilizer</en>
@@ -95,6 +95,7 @@
         <action name="SF_MINIMAP_ZOOM"      category="ONFOOT" />
         <action name="SF_SCOUT"             category="ONFOOT" />
         <action name="SF_TREATMENT"         category="ONFOOT" />
+        <action name="SF_HANDFUL"           category="ONFOOT" />
     </actions>
 
     <l10n filenamePrefix="translations/translation" />
@@ -111,6 +112,15 @@
         <actionBinding action="SF_MINIMAP_ZOOM" />
         <actionBinding action="SF_SCOUT" />
         <actionBinding action="SF_TREATMENT" />
+        <!-- SF-39: the only SF action that ships a DEFAULT key. The engine reads
+             a default from a nested <binding> child (InputBinding:loadBindingsFromXML
+             -> Binding.createFromXML); the bare actionBinding entries above have
+             never had one, which is why every other SF hotkey must be bound by
+             hand despite what the code comments claim. Retro-fitting the rest is
+             a change to shipped behaviour and is deliberately not done here. -->
+        <actionBinding action="SF_HANDFUL">
+            <binding device="KB_MOUSE_DEFAULT" input="KEY_lshift KEY_g" />
+        </actionBinding>
     </inputBinding>
 
     <!-- Custom fill types declared in external file (FS25 requires filename= attribute) -->
--- a/src/SoilFertilityManager.lua
+++ b/src/SoilFertilityManager.lua
@@ -147,6 +147,12 @@ function SoilFertilityManager.new(mission, modDirectory, modName, disableGUI)
             SoilLogger.info("Soil Scout dialog registered")
         end
 
+        -- [SF-39] The Handful panel (opened from the SF_HANDFUL hotkey, kneeling only)
+        if SoilHandfulDialog and g_gui then
+            SoilHandfulDialog.register(modDirectory)
+            SoilLogger.info("Soil Handful panel registered")
+        end
+
         -- Version/changelog dialog (shown once per version on load)
         if SoilVersionDialog and g_gui then
             SoilVersionDialog.register(modDirectory)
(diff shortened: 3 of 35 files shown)


sasha-rf (approved):
Approved. Reviewed the full diff at the PR head (two commits: the panel, then the keybinding name + restyle from the first in-game pass).

This is SF-39, the missing reading surface for the kneel: kneel in a field, press the SF_HANDFUL key (default Shift+G), and the frozen SF-38 payload is drawn verbatim. The panel writes nothing; the reveal happens before the assemble so diseaseKnown reflects the just-knelt cell.

Verified at source, not by claim:

- The input path matches the shipped Scout/Treatment pattern exactly: same registerActionEvent signature, PLAYER context, setActionEventTextVisibility(false). The SF_HANDFUL action is declared in modDesc with a real default binding via the nested <binding> child, which is the documented way FS25 reads defaults, and the comment correctly notes the other ten SF actions never had one.
- Every refusal speaks through a blinking warning: locked family (ReleaseGate.isSystemLive("read_the_dirt")), not crouching (player.mover.isCrouching), no field underfoot, no read. A silent no-op would have taught the player the key is broken.
- The kneel is server-authoritative: direct revealCellAt on the host, SoilKneelEvent request on a client, identical to the shipped SF_SCOUT path.
- All referenced APIs exist in current source: ReleaseGate.isSystemLive/liveOptIn/lockMessage, HandfulRead.assemble and CLAUSES, SpatialScouting:revealCellAt/isArmed, SoilKneelEvent. The payload fields the panel reads (N/P/K {value,status}, pH, OM, compaction, moisture, diseasePressure, diseaseGrain, pestPressure, weedPressure, lastCrop history, rotationStatus, organicState, materialWetness/DaysDown/Verdict, testKitActive, fromZoneCell) all match HandfulRead.assemble's return shape.
- The rendering contract is pinned by the bench and reads honestly: per-clause neutrality (absent is a dash in neutral grey, never a zero and never a severity colour), bands not numbers (readTestKit is false, so gated clauses render words), grain tags on every clause, and the material section hides entirely when nothing lies under the hand. The two polarity traps are handled and tested: LOW compaction is the good outcome, and Unscouted takes neutral rather than the green of a clean field.
- The dialog XML matches the sibling dialog pattern (SoilScoutDialog: newLayer, dialogFullscreenBg, fs25_dialogBg*, onCreate/onOpen/onClose, buttonBack), and the controller's ELEMENT_IDS list keeps it in step with the XML.
- All 27 translation files carry the 54 sf_handful_* keys plus input_SF_HANDFUL, with English fallbacks in the controller so a missing key still renders.

Suite 2823/0 across 66 files (includes handful_panel_sf39_test at 51 assertions), Lua 5.1 syntax and lint clean, same two pre-existing validator errors as development. No em dashes. One feature, one PR, base development.

One coordination note for the merge, not a defect in this PR: PR #833 (experimental gate confirm) also bumps modDesc from the same 2.5.0.61 base (832 goes to 2.5.0.63, 833 to 2.5.0.64). Whichever merges second will need the version line resolved at merge time.

sasha-rf:

sasha-rf (approved):
## Approved

The member Read the Dirt has been missing since 2026-07-31. Reviewed as an implementation of the SF-38 contract rather than as new design, which is what it is: the frozen payload is rendered and not one clause of it moved.

**The precondition ladder is real and it is ordered correctly.** Gate, kneel, field, then reveal, then assemble, then show. SF-38 rule 5 requires `diseaseKnown` to reflect the just-knelt cell, and the reveal genuinely precedes the assemble rather than merely appearing to. Every refusal speaks through a blinking warning; none of them fails silently.

**The crouch predicate is sourced, not assumed.** `PlayerStateCrouch:onStateEntered/onStateExited` set it (`player/stateMachine/PlayerStateCrouch.lua:31,35`), the flag is a plain bool on `PlayerMover` (`player/PlayerMover.lua:45,168-169`), `Player.mover` is assigned at `player/Player.lua:171`. Crouch being a hold rather than a toggle is documented rather than worked around, and so is the hand-tool case where `canCrouch` is false and the player physically cannot kneel. Correct call: that is the engine's rule and inventing our own crouch state to dodge it would be worse than the limitation.

**Element ids are in sync.** The controller's `ELEMENT_IDS` list is 34 entries and every one exists in the XML, with nothing declared in the XML left unfetched. This matters more than it looks: a missing id would leave `self[id]` nil and `setText` would no-op in silence, so the panel would lose a row with no error anywhere.

**The band helpers return a severity alongside the word, and the multi-return lands where intended** in the final argument position of `tile()`. The bench proves it rather than the reader having to trust it.

**No new thresholds.** N/P/K come pre-banded from the core, pH and OM from `REPORT_COLORS`, compaction from the cut points the shipped map layer already displays against, pressure from `DISEASE_PRESSURE`. Nothing invented, so this surface cannot drift from the ones already shipping.

### The two polarity decisions are the best part of this

**LOW compaction is coloured as the GOOD outcome**, not red-by-analogy with the nutrient bands, and **absent or Unscouted takes neutral grey rather than the green of a clean field.** Both would have looked completely fine and been wrong, and both are pinned by assertions instead of comments. The second one matters commercially as well as honestly: painting an unlooked-at field green would undercut the scouting economy the mod sells.

Colour is a second channel with the word always present, so nothing is lost to a colour-blind player.

### One nit, non blocking

When disease is `Unscouted`, the grain line underneath still renders "field average". The grain label describes the KNOWLEDGE grain rather than a measurement, so it is not wrong, but sitting under the word Unscouted it can read as though a field-average measurement exists. Consider blanking that line when the clause is unscouted. It is cosmetic and it should not hold the merge.

### Two findings in the PR body worth keeping visible

That **no SF hotkey has ever had a default key** is a real discovery about shipped behaviour, not a note about this feature. Every "default Shift+K" in our comments has been fiction since release. Leaving the other ten alone here is the right scope call and it wants its own decision later.

The `input_SF_HANDFUL` miss is the same shape as the `cs_pda_btn_help` gap I reviewed this morning: the key that names a thing living somewhere the resolver never reaches. Caught in-game and fixed in the second commit.

### Verified

Suite 2823/0 across 66 files, bench 51 assertions, Lua 5.1 syntax and lint clean. 54 `sf_handful_*` keys plus `input_SF_HANDFUL` in all 27 translation files with English fallbacks in code. Built, deployed and confirmed in-game by Tyson at v2.5.0.62 for function and v2.5.0.63 for the binding label and the restyle, so the tile layout in front of you has been seen on a real screen.

### Not verified by me

Multiplayer client behaviour. The panel adds no event and server-only clauses degrade to neutral under per-clause neutrality, which is the honest answer, but nobody has run it on a client. If a client shows so many neutral clauses that the panel stops being useful, that is a follow-up request/response pair rather than a fault in this.

sasha-rf:

PR #833, opened 2026-08-13
feat(settings): the experimental gate asks before it flips

TheCodingDad-TisonK:
## The experimental gate asks before it flips

Turning Experimental Systems on arms **unfinished** systems on a live savegame, and it did that on a single click with nothing between the player and the consequence. It now warns and waits for a Confirm.

### Where it hooks

`SoilSettingsPanel:requestChange` is the single choke point every toggle, cycle and reset funnels through, so intercepting there means no path can flip the gate unasked. The original body moves to `_applyChange` **unchanged**; only a confirmed answer reaches it.

### The properties that matter

The point is not that a dialog appears. It is that **abort leaves the setting exactly as it was**, in both directions. A confirmation that fails open is worse than no confirmation at all, because the player now believes they declined.

- **FAIL CLOSED.** If `YesNoDialog` is unavailable the change is refused, not silently applied. A gate that cannot ask does not open.
- **Non-admins are refused BEFORE the prompt**, rather than being walked through a scary confirmation whose change we were always going to reject.
- **No prompt when the value is not actually changing** (re-selecting the current value).
- **`resetCurrentCategory` passes `skipConfirm`.** It is already a deliberate act, it runs in a loop a modal would break, and for this gate it only ever restores the LOCKED default. Nothing to warn about.

### The text

Presented as `DialogElement.TYPE_WARNING`, with direction-specific copy. Enabling says the systems are not finished, can change without warning, can affect the savegame irreversibly, and to keep a backup. Disabling says what they already wrote stays put but stops being updated. Buttons are Confirm / Abort.

`YesNoDialog.show(callback, target, text, title, yesText, noText, dialogType, ...)` per `gui/dialogs/YesNoDialog.lua:8`; the callback shape with a target set is `callback(target, value, args)` per `:82-86`.

### Verification

- Bench `experimental_confirm_test.lua`: **23 assertions** covering the ask, abort-in-both-directions, confirm applies, fail-closed, no-op, non-admin, reset-skips, and that no other setting gains a prompt.
- Full suite **2795 passed, 0 failed across 66 files**; Lua 5.1 syntax and lint clean.
- 5 l10n keys in **all 27** translation files, English fallbacks in code.
- **Not yet confirmed in-game.**

### Note on the version number

Bumped to **2.5.0.64**, deliberately past the 2.5.0.63 on `feat/SF-39-handful-panel` (PR #832), which is open in parallel. Whichever merges second will conflict on the version line in `modDesc.xml`; resolve it to 2.5.0.64.

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,7 +440,7 @@ This mod is licensed under **[CC BY-NC-ND 4.0](https://creativecommons.org/licen
 
 You may share it in its original form with attribution. You may not sell it, modify and redistribute it, or reupload it under a different name or authorship. Contributions via pull request are explicitly permitted and encouraged.
 
-**Author:** TisonK &nbsp;·&nbsp; **Version:** 2.5.0.0
+**Author:** TisonK &nbsp;·&nbsp; **Version:** 2.5.0.64
 
 © 2026 TisonK - See [LICENSE](LICENSE) for full terms.
 
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.61</version>
+    <version>2.5.0.64</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
         <en>Realistic Soil &amp; Fertilizer</en>
--- a/src/ui/SoilSettingsPanel.lua
+++ b/src/ui/SoilSettingsPanel.lua
@@ -474,7 +474,84 @@ function SoilSettingsPanel:isAdmin()
     return SoilUtils.isPlayerAdmin()
 end
 
-function SoilSettingsPanel:requestChange(id, value)
+--- Every setting change from this panel funnels through here, which is why the
+--- experimental gate is intercepted at this one point rather than at each of the
+--- five call sites that can flip a boolean.
+---
+--- @param skipConfirm boolean|nil  bypass the experimental confirmation. Used by
+---   resetCurrentCategory, which is already an explicit deliberate action and
+---   only ever restores the LOCKED default, so it never needs to ask.
+function SoilSettingsPanel:requestChange(id, value, skipConfirm)
+    local def = SettingsSchema.byId[id]
+    if not def then return end
+
+    -- THE EXPERIMENTAL GATE NEVER FLIPS ON ONE CLICK. It arms unfinished systems
+    -- on a live save, so it asks first and a refusal leaves the toggle untouched.
+    if id == "experimentalSystems" and skipConfirm ~= true and value ~= self.settings[id] then
+        -- Refuse BEFORE we warn: putting a scary confirmation in front of someone
+        -- whose change we are about to reject anyway is worse than refusing now.
+        if not def.localOnly and not self:isAdmin() then
+            self:_warnNotAdmin()
+            return
+        end
+        self:_confirmExperimental(value)
+        return
+    end
+
+    self:_applyChange(id, value)
+end
+
+function SoilSettingsPanel:_warnNotAdmin()
+    if g_currentMission and g_currentMission.hud and
+       g_currentMission.hud.showBlinkingWarning then
+        g_currentMission.hud:showBlinkingWarning(
+            "Only server admins can change this setting", 4000)
+    end
+end
+
+--- Ask before arming or disarming the experimental systems. Abort leaves the
+--- setting exactly as it was; only Confirm reaches _applyChange.
+function SoilSettingsPanel:_confirmExperimental(value)
+    -- FAIL CLOSED. If the dialog is unavailable we do NOT fall through to the
+    -- change: a gate that cannot ask is a gate that does not open.
+    if YesNoDialog == nil or type(YesNoDialog.show) ~= "function" then
+        SoilLogger.warning("SoilSettingsPanel: YesNoDialog unavailable, experimental toggle refused")
+        return
+    end
+
+    local text = value == true
+        and tr("sf_exp_confirm_on",
+            "This turns on systems that are NOT finished. They can behave oddly, " ..
+            "change without warning, or affect this savegame in ways that cannot be " ..
+            "undone. Use a backup save.\n\nTurn experimental systems ON?")
+        or tr("sf_exp_confirm_off",
+            "Experimental systems will stop running on this savegame. Anything they " ..
+            "wrote stays where it is, but it will no longer be updated.\n\n" ..
+            "Turn experimental systems OFF?")
+
+    YesNoDialog.show(
+        SoilSettingsPanel._onExperimentalConfirm, self,
+        text,
+        tr("sf_exp_confirm_title", "Experimental Systems"),
+        tr("sf_exp_confirm_yes", "Confirm"),
+        tr("sf_exp_confirm_no", "Abort"),
+        DialogElement.TYPE_WARNING,
+        nil, nil,
+        { value = value }
+    )
+end
+
+--- YesNoDialog callback. With a target set the engine calls this as
+--- callback(target, value, callbackArgs) (gui/dialogs/YesNoDialog.lua:82-86).
+function SoilSettingsPanel:_onExperimentalConfirm(confirmed, args)
+    if confirmed ~= true then return end          -- Abort: the toggle never moves.
+    if type(args) ~= "table" or args.value == nil then return end
+    self:_applyChange("experimentalSystems", args.value)
+end
+
+--- The original change path, unchanged. Reached directly for every setting
+--- except the experimental gate, and via the confirmation for that one.
+function SoilSettingsPanel:_applyChange(id, value)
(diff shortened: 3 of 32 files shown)


sasha-rf (approved):
Approved. Reviewed the full diff at the PR head (single commit).

The Experimental Systems toggle arms UNFINISHED systems on a live savegame, and it flipped on a single click with nothing between the player and the consequence. This PR intercepts at SoilSettingsPanel:requestChange, the one choke point all five toggle call sites funnel through, so no path flips it unasked. The original body moves to _applyChange unchanged.

Verified at source:

- The gate only triggers for id == "experimentalSystems", only when the value is actually changing, and only without skipConfirm. Every other setting passes straight through to _applyChange, unchanged. The interception is surgical.
- ABORT IS A NO-OP: YesNoDialog's callback returns immediately unless confirmed, so a No leaves the stored value untouched. The bench pins it in both directions (abort-arm and abort-disable).
- FAIL CLOSED: if YesNoDialog is unavailable, the change is refused with a log line, never silently applied. A gate that cannot ask does not open.
- Admin is refused BEFORE the prompt via _warnNotAdmin, so a non-admin is never walked through a confirmation we would then reject. _applyChange re-checks admin after the modal, which is right: a modal is not instant and admin status can change under it.
- resetCurrentCategory passes skipConfirm=true. That is the correct call: it is already a deliberate act, it runs in a loop a modal would break, and for the gate it only ever restores the LOCKED default.
- The YesNoDialog.show call uses the proven API shape (callback, target, text, title, yes, no, DialogElement.TYPE_WARNING) with the engine's documented callback(target, value, args) contract, and TYPE_WARNING is the right presentation for an arming warning.
- All 27 translation files carry the 5 sf_exp_confirm_* keys with English fallbacks in code. The newline escaping in the on/off strings is correct.

Suite 2795/0 across 66 files (includes experimental_confirm_test at 23 assertions), Lua 5.1 syntax and lint clean. No em dashes. One feature, one PR, base development.

One merge-coordination note for Tyson, not a defect in this PR: PR #832 (SF-39 handful panel) also bumps modDesc from the same 2.5.0.61 base (this PR goes to 2.5.0.64, 832 to 2.5.0.63), and both add changelog bullets at the top of SoilVersionDialog.CHANGELOG. Whichever merges second will need the version line and changelog head resolved at merge time.

sasha-rf:

sasha-rf (approved):
## Approved

A guard on an existing control, not a new system. The release gate already exists and its own lock message already says "at your own risk"; this puts the friction where the risk is and changes nothing about which systems are locked.

### The second commit is the one to read

The first commit shipped this as a base-game `YesNoDialog`. **That would not have worked**, and the reason is worth stating in full because it is not obvious from the diff:

> `SoilSettingsPanel:update()` closes the panel whenever `g_gui` reports a dialog visible.

So the confirmation would have dismissed the very panel the player was answering for, leaving a prompt floating over nothing. The rework to a drawn `drawConfirmDialog` fixes a correctness bug and not only a style mismatch. **Recording that the style question found it first**: "does this look like it belongs in the panel" was the cheaper question and it landed on the same answer as the correctness one.

The drawn version is built like `drawPopupDialog` rather than a new invention, and the accent is `C.lock_text`, the amber this panel already uses to mark a locked system. That is the right borrow: the prompt reads as part of the panel's existing language for "this is gated" rather than as an interruption of it.

**Abort is genuinely a no-op, in both directions.** This is the whole point and it is the thing most likely to be got wrong. A confirmation that fails open is worse than none at all, because the player now believes they declined. Asserted for on and off separately.

**Nothing reaches the change except the Confirm button.** The interception sits at `requestChange`, the single choke point all five toggle, cycle and reset call sites funnel through, so there is no second path to audit. The original body moved to `_applyChange` unchanged.

**No pending state outlives the surface that asked.** `close()` clears it, which covers escape, the `[X]`, the hotkey and the auto-close in `update()`. None of those is an answer and none of them may leave a value a later Confirm could pick up. There is an assertion that a stale Confirm arriving after a close applies nothing, which is exactly the test I would have asked for if it were missing.

**The modal is actually modal.** Click zones are cleared before it draws so nothing behind is clickable through it, and outside-click-to-close is disabled while it is up.

**Scope is surgical.** No other setting gains a prompt, there is no prompt when the value is not changing, and `resetCurrentCategory` passes `skipConfirm` because it runs in a loop a modal would break and only ever restores the LOCKED default.

**Non-admins are refused before the prompt** rather than walked through a warning about a change we were always going to reject. Small thing, right instinct.

**Body text is wrapped through the same helper shape `SoilMapOverlay` uses**, so a long translation cannot run past the box. Given the copy here is deliberately wordy in English and will be longer in German, that is not optional.

### Two notes, neither blocking

**The prompt is mouse only.** There is no Enter or Escape handling on the two buttons. That is consistent with the panel, which is mouse driven throughout, and the failure mode is safe: Escape closes the panel, which aborts. Worth knowing rather than fixing.

**A second `requestChange` while the prompt is open would overwrite the pending value.** Unreachable from the UI, because click zones behind the modal are cleared, so this is a note for whoever adds a non-click path later rather than a defect today.

### Verified

Bench 31 assertions, up from 23 when it was a GUI dialog. Suite 2803/0 across 66 files on this branch, 2854/0 on the combined build with #832. Lua 5.1 syntax and lint clean. 7 `sf_exp_confirm_*` keys in all 27 translation files with English fallbacks in code. Zero `YesNoDialog` references remain in the panel.

### Not verified by me

In-game. The drawn prompt has not been seen on a real screen yet, unlike the SF-39 panel in #832. **The one thing an in-game pass must confirm is the abort path**: flip the toggle, read the box, hit Abort, and check the toggle has not moved.

### Merge order

This carries 2.5.0.64 and #832 carries 2.5.0.63. Whichever lands second conflicts on that one line in `modDesc.xml`; resolve to the higher number. Worth being deliberate about it, because a careless `--theirs` on that file during a local test merge earlier today silently deleted an entire input action from the build.

sasha-rf:

sasha-rf (approved):
## Approval refreshed for `054a47eb`

My earlier approval covered two commits. This PR now carries three, and an old approval never covers code it did not see, so here is the third reviewed on its own.

### What changed

`054a47eb` stops drawing the settings page while the confirmation is up. Tyson reported the settings rows reading straight through the confirmation box.

**The cause is worth writing down, because it will catch the next person too.** The dim overlay and the box background only cover what was drawn as an OVERLAY. Text already rendered underneath still shows through a box painted on top of it, so a modal in this panel cannot be made readable by painting over the page. It has to not draw the page.

**`drawPopupDialog` has always done exactly that, and its comment says so.** The confirmation kept the page drawn deliberately, to leave the settings visible as context behind the modal, and that intention is precisely what made its own body copy unreadable. Following the file's existing precedent is the right resolution rather than inventing a second approach.

### What I checked

**The title and info bars stay drawn, and that is correct, not an oversight.** The popup leaves them up too. I checked the geometry rather than assuming: the confirmation box occupies 0.370 to 0.630, the title bar 0.818 to 0.870 and the info bar 0.130 to 0.176. No overlap in either direction, so nothing bleeds through from the parts that remain.

**Nothing else moved.** One file, one condition, and the modal contract from the previous commit is untouched: click zones still cleared, outside-click still disabled, `close()` still aborts a pending answer.

### Verified

Suite 2803/0 across 66 files, Lua 5.1 syntax and lint clean.

### Still not verified by me

In-game. The readability fix is the whole point of this commit and it is precisely the thing a bench cannot see. **The pass is: open Settings, go to Experimental, flip the toggle, and confirm the warning box is legible with no settings rows showing through it.** The abort check from my earlier review still stands alongside it.

sasha-rf:

sasha-rf (approved):

This PR now carries four commits. My earlier approval covered three, and an old approval never covers code it did not see, so here is the conflict resolution reviewed on its own.

### What the merge had to decide

PR #832 (SF-39) landed on development while this branch was open, which is what made it DIRTY. Three kinds of conflict, resolved three different ways, and the distinction is the whole review:

**27 translation files: union.** Both sides append distinct key blocks before `</elements>`, so keeping both is correct and mechanical.

**`SoilVersionDialog` changelog: both kept, ordered deliberately.** Both sides added bullets. SF-39 is placed first because it is the larger feature and already on development.

**`modDesc` / `README`: resolved to 2.5.0.64**, this branch's own bump, which sits above development's 2.5.0.63.

**No source file conflicted at all**, which is the reassuring part: the two features touch different code.

### What I checked

Verified after resolution rather than assumed, because a union merge on the wrong file type produces a file that still looks plausible: **Lua 5.1 syntax clean across 87 files, suite 2854/0 across 67 files, and no conflict markers left anywhere in the tree.**

That check is not theoretical. The same union heuristic applied to a source file earlier today produced a function with a signature and no body, which the syntax gate caught at commit time. Translations tolerate it; code does not.

### Merge order note

Three SoilFertilizer PRs are open with ascending versions: this one at 2.5.0.64, #834 at 2.5.0.67, #835 at 2.5.0.69. **Merging in that ascending order avoids re-conflicting on the version line.** Any other order will make the later one dirty again on exactly one line.

Everything from my earlier review of the first three commits stands unchanged.

sasha-rf:

PR #834, opened 2026-08-13
fix(hud): the HUD stands down while a full-screen panel is open

TheCodingDad-TisonK:
## The HUD stands down while a full-screen panel is open

Reported in-game: the Soil Monitor's Good / Fair / Poor labels and nutrient rows read straight **through** the settings panel drawn over them.

### Cause

Same one that made the experimental confirmation box unreadable. **A panel background is an OVERLAY, and an overlay does not cover text already rendered underneath it.** A HUD cannot be hidden by painting on top of it. It has to not draw.

`SoilMasterHUDBridge.drawStack` draws `soilHUD` first and `settingsPanel` second, so the Monitor's text was always going to survive the panel painted over it.

### The rule already existed, it just had a blind spot

This is not a new policy. `SoilHUD:draw` already stands down for a base-game GUI at `SoilHUD.lua:1302`, and all four in-vehicle panels carry the same `getIsGuiVisible` guard. **None of them could see OUR panels**, because those are drawn surfaces rather than `g_gui` ones. So every surface in the mod was already following the right rule against the wrong list.

### Where it is fixed

`SoilMasterHUDBridge.drawStack`, which is the single place that owns draw order and is shared by both the MasterHUD path and SF's own `FSBaseMission.draw` hook. The file's own header says those two must never diverge, so fixing the shared body is the only way to fix both at once.

Covers the Soil Monitor, the four in-vehicle panels (variable rate, smart sensor, sprayer info, harvester) and the minimap overlay, for all three full-screen panels: settings and both tuning editors. **The three are mutually exclusive by construction** because the settings panel calls `close()` on itself before opening either editor, so drawing all three in the suppressed branch is safe rather than sloppy.

### Verification

Bench `hud_suppression_test.lua`, 21 assertions: the full stack still draws when nothing is open, every passive element stands down for each of the three panels, the panel itself still draws, and the nil-safety paths hold (a missing element, a panel with no `isOpen`, and no manager at all).

Full suite **2844 passed, 0 failed across 67 files**. Lua 5.1 syntax and lint clean.

**Not verified in-game yet.** The whole point is legibility, which is the one thing a bench cannot see: open Settings over the Soil Monitor and confirm no HUD text shows through it.


sasha-rf (approved):
## Approved

A one-condition fix that closes a gap in a rule the whole mod was already following. Reviewed as a correction, which is what it is: no surface changes what it draws, only whether it draws.

### The diagnosis is the valuable part

**A panel background is an overlay, and an overlay does not cover text already rendered underneath it.** So a HUD cannot be hidden by painting on top of it; it has to not draw. That single sentence explains both this report and the confirmation box being unreadable an hour earlier, and it is now written into the code rather than living in someone's head.

`drawStack` drew `soilHUD` first and `settingsPanel` second, so the Soil Monitor's Good / Fair / Poor labels were always going to survive the panel painted over them. Nothing about the panel could have fixed it.

**`drawStack` really is the only draw path.** The claim in the file header is load bearing here, because a second path would leave this half fixed. Verified: `main.lua:764` calls it from the `FSBaseMission.draw` hook and `SoilMasterHUDBridge:103` registers it as MasterHUD's draw. Two entry points, one body, so both the standalone and the MasterHUD case are covered by the same change.

**The rule is genuinely pre-existing.** `SoilHUD:draw` stands down for a base-game GUI at `SoilHUD.lua:1302`, and all four in-vehicle panels carry the same `getIsGuiVisible` guard. Every surface was already following the right rule against the wrong list, because our panels are drawn rather than `g_gui`. This is the list being corrected, not a new policy being invented, which is why it needs no design call.

**The early return covers the minimap overlay too**, not just the eight panel draws. Easy to miss, since it sits below them in the function.

**The three full-screen panels are mutually exclusive by construction**, so drawing all three in the suppressed branch is correct rather than lazy: the settings panel calls `close()` on itself before opening either tuning editor. I checked that rather than assuming it from the naming.

**Nil safety is real.** `type(p.isOpen) == "function"` means a panel that never exposes `isOpen` is treated as closed instead of erroring, and the bench pins that shape explicitly along with a missing element and a missing manager. Those are the paths that turn a cosmetic fix into a crash if they are wrong.

### One note, not blocking

The suppression is all or nothing: while a panel is open, the work trails and the minimap overlay go with the HUD. That is right for legibility and it is what the popup already does, but it does mean a player cannot check the map overlay against a setting while changing it. Worth knowing if anyone later asks for the overlay to stay up.

### Verified

Bench `hud_suppression_test.lua`, 21 assertions: the full stack still draws when nothing is open, every passive element stands down for each of the three panels, the panel itself still draws, and all three nil paths hold. Suite 2844/0 across 67 files, Lua 5.1 syntax and lint clean.

### Not verified by me

In-game, and it is the whole point of the change. **The pass is: open Settings while the Soil Monitor is up and confirm no HUD text reads through the panel.** A bench can prove `draw` was not called; it cannot prove the screen is legible.

sasha-rf:

sasha-rf (approved):

PR #833 landed on development, which made this dirty. My earlier approval covered one commit; this now carries two.

### What the merge decided

**Only `modDesc.xml` and `README.md` conflicted, and only on the version line.** No source file conflicted at all, and the changelog merged cleanly on its own with all three bullets intact. Version resolved to **2.5.0.67**, above development's 2.5.0.64.

That is the whole change. Everything in my earlier review of the fix itself stands.

### Verified after resolution rather than assumed

Suite 2875/0 across 68 files, Lua 5.1 syntax and lint clean, no conflict markers anywhere in the tree.

### One thing to know about merge order

**Merge this before #835.** Both branches edit `SoilMasterHUDBridge.lua` in the same region, so whichever goes second will conflict there on real code rather than a version string. #835 carries 2.5.0.69, above this one, so the ascending sequence also keeps the version line quiet.

sasha-rf:

PR #835, opened 2026-08-13
feat(hud): declare the fullscreen claim to MasterHUD

TheCodingDad-TisonK:
## Declare the fullscreen claim to MasterHUD

Companion half of the `isFullscreen` contract wired in FS25_MasterHUD PR #7. **Merge that one first**; this is inert without it, and harmless.

### What it does

While one of SoilFertilizer's full-screen panels is open (settings or either tuning editor), MasterHUD now stands **every other companion's HUD** down too. Without it the Income and Tax HUDs render their text before our panel and read straight through it, because **an overlay does not cover text already drawn underneath**.

### What it is not

**SF's own HUD is not this.** That is handled inside `drawStack` by PR #834, which is the single-mod half. This PR is only the cross-mod half.

**It only has an effect when MasterHUD is present**, because MasterHUD is what owns cross-mod ordering. Standalone, each mod draws its own stack through its own `FSBaseMission.draw` hook and cannot see the others, so the cross-mod symptom remains in that configuration. That is a property of the architecture rather than something this change can fix.

### Compatibility

`isFullscreen` is optional on MasterHUD's side, so an older MasterHUD ignores the field and behaves exactly as before. **No version dependency is introduced**, and the two PRs can land in either order without breaking anything.

### Note for after #834 lands

#834 adds a local `fullScreenPanelOpen` helper inside `drawStack` for the single-mod case, and this PR adds `SoilMasterHUDBridge.isFullscreen` for the cross-mod case. They ask the same question. Once both are merged, `drawStack` should call the module function and the local should go. **One line, flagged rather than hidden**, and deliberately not done here because it would mean touching an approved PR's branch.

### Verification

The callback is a cheap read of state that already exists, since MasterHUD calls it every frame from its draw loop. Suite 2823/0 across 66 files, Lua 5.1 syntax and lint clean. **Not verified in-game**: the pass needs MasterHUD PR #7 merged, then open SF's settings panel with the Income and Tax HUDs on screen and confirm nothing reads through it.


sasha-rf (approved):
## Approved

The companion half of the MasterHUD contract. Small, optional, and correctly scoped.

### What I checked

**It is genuinely inert without MasterHUD PR #7**, and inert is the right word: `isFullscreen` is an unknown key in an older MasterHUD's subscribe spec, so it is ignored and behaviour is exactly as today. **No version dependency is created**, and the two can land in either order without breaking anything. That is the property that makes this safe to merge ahead of the bedrock change.

**The callback is a cheap read.** MasterHUD calls it every frame from the draw loop, so it walks three references and returns. No allocation, no computation.

**It uses `p:isOpen()` as a method**, which is correct for these three panels and is worth stating because the same field in RandomWorldEvents is a plain boolean. Getting that backwards in either repo would be silently catastrophic: a function value is truthy on every frame, so the claim would never release and every other mod's HUD would stay hidden.

**Scope is honest.** SF's own HUD is not this PR, it is #834. This is only the cross-mod half.

### The limitation is stated, which I want to acknowledge

It has no effect without MasterHUD installed, because MasterHUD is what owns cross-mod ordering. Standalone, each mod draws through its own hook and cannot see the others. **That is in the PR body rather than discovered later by a player**, which is the difference between a known limitation and a bug report.

### The duplication note is now resolved

The body flags that once #834 lands, `drawStack`'s local `fullScreenPanelOpen` should collapse into this module function. That collapse has since been done and verified in the combined local test build, so the follow-up is already proven rather than merely intended.

### Verified

Suite 2823/0 across 66 files, Lua 5.1 syntax and lint clean. **Not verified in-game**: needs #7 merged, then SF's settings panel open with the Income and Tax HUDs on screen.

sasha-rf:

sasha-rf (approved):

PR #833 landed on development. My earlier approval covered one commit; this now carries two.

### What the merge decided

**Only `modDesc.xml` and `README.md` conflicted, on the version line alone.** No source file conflicted. Version resolved to **2.5.0.69**, above both development's 2.5.0.64 and #834's 2.5.0.67.

Everything in my earlier review of the claim itself stands.

### The one that is still coming, flagged rather than discovered

**This branch and #834 both edit `SoilMasterHUDBridge.lua` in the same region**, so once #834 merges this will conflict there once more, on real code this time. That is expected and the resolution is already known and proven: `SoilMasterHUDBridge.isFullscreen` becomes the single implementation and `drawStack` calls it instead of keeping its own local `fullScreenPanelOpen`. The combined local test build has been running that exact shape all afternoon.

**Merge #834 first**, then this one after its final resolution.

### Verified

Suite 2823/0 across 66 files, Lua 5.1 syntax and lint clean.

sasha-rf:

sasha-rf (approved):
## Approval refreshed for the second merge commit

PR #834 landed on development, producing the code conflict flagged in advance on this PR. Three commits now.

### The resolution is the one that was predicted, not an improvisation

Both branches added a "is a full-screen panel open" check to `SoilMasterHUDBridge.lua` in the same region. Resolved as planned: **`SoilMasterHUDBridge.isFullscreen` is now the single implementation and `drawStack` calls it**, replacing #834's local `fullScreenPanelOpen`.

Two callers, one question:

- `drawStack`, to stand **our own** HUD down (the single-mod half, #834)
- MasterHUD via the subscribe spec, to stand **every other** companion's HUD down (the cross-mod half, this PR)

Collapsing them is the right call rather than tidiness: two functions answering the same question is how they drift apart later, and this one governs whether HUDs appear at all.

This shape is not new here either. It has been running in the combined local test build all afternoon.

### Verified, and this is the merge shape that has already bitten once today

An earlier union-style resolution of this exact file produced a function with a signature and no body, caught only by the pre-commit gate. So this was checked rather than assumed: **one definition, three references, no conflict markers, Lua 5.1 syntax clean across 87 files, suite 2875/0 across 68 files.**

Version resolved to 2.5.0.69, above development's 2.5.0.67.

Everything in my earlier reviews stands. **This is the last of the five and it is CLEAN.**

sasha-rf:

Issue #475, opened 2026-05-29
Pass % not counting correctly


Arissani:
### Mod Version

newest

### Farming Simulator 25 Version

1.8

### Game Mode

Singleplayer

### What happened?

<img width="2514" height="1360" alt="Image" src="https://github.com/user-attachments/assets/e42f3638-a68b-4bb1-b893-8a5dfb601df5" />

A - Full Pass gave 51% 

### Steps to Reproduce

spray a field fully and watch pass %

### Log Snippet

```text
n/a
```

### Other Mods Loaded

n/a

Arissani:
tested on a new map with base game sprayer not modded and pass % worked correctly spraying liquid lime, tho there was issue im about to put in report for

TheCodingDad-TisonK:
Investigated in v2.2.4.0. The cell-based coverage tracking (markBoomCells) is accurate for boom sections. If Pass% reads low, check that the sprayer has VariableWorkWidth sections and that the field polygon matches the sprayed area. A Courseplay integration for exact course-based % is tracked in #485.

Arissani:
i was testing in the base game rotogartor spray with 40m boom, i think it has variable boom but will recheck once i get out of guild war2 and back into farm sim here soon :)

Arissani:
i switched to a new map for testing as well, not using the judith plains multi fruit edit anymore that map has 50-60 changes that could be making testing weird now using midwest horizon 16x

TheCodingDad-TisonK:
Fixed in v2.2.4.1 — fieldArea now uses field.areaHa (actual crop polygon) instead of farmland.areaInHa (includes roads/hedges, ~2× crop area). Pass% now correctly reaches 100% after a full-field spray.

PR #837, opened 2026-08-14
fix(RSF-836): paint the true boom line, derived in the vehicle's own frame

TheCodingDad-TisonK:
RSF-836, the swept quad painted a third to a half of the boom width. Items 1 to 4 of Arisanni's fix doc (`RSF-836-swept-quad-boom-line.md`), the RSF FAST TRACK tier.

The cause: `paintBoomStrip` took its boom line off the ends of the cell sweep array from `getBoomCellPositions`, whose last element is always the vehicle root and whose extent is a world-axis projection. So the painted line ran from one boom tip to the middle of the machine, halved and foreshortened by the heading cosine: a 36.6 m boom painted 18.30 m square and 12.94 m at 45 degrees. Mass was conserved, so the covered band ran at 2x to 2.8x the rate and the rest of the boom took nothing.

What changed:
- The true boom endpoints are derived the way the engine derives a working width (WorkArea.lua:307-314): every collected node is transformed into the vehicle's `components[1].node` frame, the min and max local X are the boom tips, and those two world nodes are the line. Works at any heading, needs no node classification, and the frame is `components[1].node`, never the old fallback's `rootNode` yRot read.
- The fallback path (broadcast spreader, no spanning nodes) lays the working-width half span along the frame's local X via `localToWorld`, carrying the vehicle's real orientation instead of quantising onto a world axis.
- `paintBoomStrip` consumes the line and never the array ends, and gains a tip-swap guard so a swapped pair between frames never folds the quad.
- `markBoomCells` and `getBoomCellPositions` are untouched: cell stamping is byte-identical, and the inactive-VWW-section exclusion (Partial Width, #475/#476) is inherited for free.
- The teleport guard (`travel > boomLen * 3`) widens back to its designed reach as a side effect of the correct boomLen.

Item 5 (the ~500 ms anchor drop) is split out as a design item and is not in this fix, per the fix doc's tier split.

Verified: 11 new assertions (`rsf836_boom_line_test.lua`): the painted length equals the real boom at 0, 30 and 45 degrees on the main path; the fallback produces the working width; an inactive VWW section is excluded; the old array-end read is the RED case (the reported halved, foreshortened span); and paintBoomStrip stores the true line as its anchor. Full suite 2886/0, syntax and lint clean. Built and deployed as 2.5.0.70. The reporter's in-game acceptance (a wide boom on a diagonal pass) is the owed in-game pass.


sasha-rf (approved):
Approved. RSF-836, the swept quad now paints the true boom line.

paintBoomStrip used to take its line off the ends of the cell sweep array, whose last element is always the vehicle root and whose extent is a world-axis projection, so it painted one tip to the middle of the machine, halved and foreshortened by the heading cosine. The true endpoints are now derived the way the engine derives a working width: every collected node is transformed into the vehicle's own frame, and the min and max local X are the tips. It works at any heading, needs no node classification, and the frame is components[1].node, never the old fallback's rootNode read.

The fallback lays the working-width half span along the frame's local X, paintBoomStrip consumes the line with a tip-swap guard so a swapped pair never folds the quad, and cell stamping is untouched, which is the invariant that matters: the fix changes the painted line and nothing else. The partial-width exclusion of inactive VWW sections is inherited for free, and the teleport guard widens back to its designed reach.

The bench proves the main path at 0, 30 and 45 degrees, the fallback, the partial-width exclusion, and the old read as the RED case. Full suite 2886/0. Item 5 (the 500 ms anchor drop) is correctly split out as a design item, not smuggled into a fix. Not yet seen in a running game; the reporter's diagonal-pass look is the owed in-game pass.

PR #840, opened 2026-08-14
feat(organic): the transition term is now organic YEARS, normalized via Time Guard

TheCodingDad-TisonK:
The organic transition term is now organic YEARS, normalized through Time Guard.

The transition was a raw in-game day count (60/120/240, difficulty-scaled), which silently turned the intended multi-year commitment into anything from a couple of months to a couple of decades depending on the days-per-month setting. "Three years" was not three years on most saves.

What changed:
- `TRANSITION_DAYS` becomes `TRANSITION_YEARS` (2/3/5, the indicative easy/standard/hard set, standard ~3 years per the USDA/SARE anchor). The exact per-difficulty values ride the Option-Scaling Spine + the balance pass, flagged as such in the constant.
- `getTransitionDays()` resolves the term at read time: `TRANSITION_YEARS[difficulty] * 12 * daysPerPeriod` via Time Guard's context, so a year is an in-game year on any save. A mid-save days-per-period change re-normalises from the next period forward, and Time Guard absent degrades to the reference 30-day month.
- The state machine (opt-in, in_transition, certified, breach reset) and `getFieldOrganicState`'s shape are unchanged. Every consumer already reads `transitionDaysNeeded`, so none needs a change.
- The two comments that already claimed this normalization existed are now true instead of aspirational.

Build note: the brief's item 3 (accrue on Time Guard's day tick) was evaluated and not taken, because the promotion path already tracks in-game days via the monotonic counter (the identity holds) and `subscribeTick("day", ...)` fires on all peers while the shipped promotion is server-guarded. The load-bearing change is the normalized threshold.

Verified: 11 new assertions (organic_transition_timeguard_test.lua): the year-to-day resolution at 30/15/1-day months, difficulty indexing, the no-Time-Guard degrade, a mid-save days-per-period change, and the unchanged state-machine shape. The season-scaling exclusion tripwire updated to the new constant. Full suite 2897/0, syntax and lint clean. Built and deployed as 2.5.0.71. Not yet observed in a running game.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -125,3 +125,8 @@
 - [x] Item 5 (the ~500 ms anchor drop) split out as a design item, NOT in this fix.
 - [x] rsf836_boom_line_test.lua at 11 assertions (0/30/45 deg main path, fallback width, partial-width exclusion, the old array-end read as the RED case). Suite 2886/0 across 69 files; syntax and lint clean.
 - [~] In-game (owed): a wide boom on a diagonal pass, ground read after one pass.
+
+## 2026-08-14 (Fred): organic transition normalized to YEARS via Time Guard
+- [x] The transition term was a raw day count (60/120/240), so 'three years' meant anything from months to decades depending on days-per-month. It is now TRANSITION_YEARS (2/3/5 indicative, values ride the balance pass), resolved at read time through Time Guard's days-per-period. Mid-save days-per-period changes re-normalise from the next period; Time Guard absent degrades to the 30-day reference. State machine and getFieldOrganicState shape unchanged.
+- [x] organic_transition_timeguard_test.lua at 11 assertions; suite 2897/0; deployed 2.5.0.71.
+- [~] In-game (owed): a transition at a 30-day month counts down over three in-game years, not eight months.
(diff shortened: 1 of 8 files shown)


sasha-rf (approved):
Approved. The organic transition term is now organic YEARS, normalized through Time Guard.

The raw day count silently turned the intended multi-year commitment into months or decades depending on days-per-month. The term is now TRANSITION_YEARS resolved at read time through Time Guard's days-per-period, so three years means three in-game years on any save, with a mid-save days-per-period change re-normalising from the next period and Time Guard absent degrading to the reference 30-day month.

The state machine and getFieldOrganicState shape are untouched, so every consumer keeps working with no change, which is the right kind of fix. The two comments that already claimed this existed are now true. The brief's tick-migration item was evaluated and not taken with the reason stated (the monotonic path holds the identity and subscribeTick fires on all peers), which is honest engineering rather than a skipped step.

Bench is 11 new assertions covering the year-to-day math, difficulty indexing, the no-Time-Guard degrade, and the unchanged state shape. Full suite 2897/0. The years values are flagged as riding the balance pass, as they should be. Not yet seen in a running game.

PR #841, opened 2026-08-14
feat(organic): the managed compost production process

TheCodingDad-TisonK:
The managed compost production process: the player can now produce compost from farm organic waste instead of only buying it.

A managed process, no placeable (Arissani's shape). A batch commits existing feedstocks (manure, silage, digestate, biosolids), decomposes over in-game days on the Time Guard clock (simulation flow, with an SF day-tracking fallback when Time Guard is absent), and a finished batch yields the existing COMPOST fill type into farm storage. SF owns the OM amendment effect unchanged (COMPOST is an OM-primary product, 0.12 per the engine-era addendum); this system owns production only, with no new fill type and no new soil authority.

Each batch carries an organic-safe flag derived from its feedstocks: only the approved manure/compost/digestate family keeps a batch cert-safe, and a biosolids-fed batch is a valid amendment but not organic-safe, with the breach riding SF's existing onInputApplied rather than new breach logic.

Server-authoritative (a client never mints compost), persisted via StateLedger with the own-XML fallback, and three console commands for the test pass (sfCompostStatus / sfCompostStart / sfCompostCollect). The batch duration is a natural rate, not days-per-period normalized, and the yield + days ride the balance pass.

Verified: 25 new assertions (compost_manager_test.lua): batch start + output math, the organic-safe flag including the biosolids case, decomposition to readiness, collect-and-clear, the not-ready refusal, persistence round trip, and the client cannot start a batch. Full suite 2922/0, syntax and lint clean. Built and deployed as 2.5.0.72. Not yet observed in a running game.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -130,3 +130,8 @@
 - [x] The transition term was a raw day count (60/120/240), so 'three years' meant anything from months to decades depending on days-per-month. It is now TRANSITION_YEARS (2/3/5 indicative, values ride the balance pass), resolved at read time through Time Guard's days-per-period. Mid-save days-per-period changes re-normalise from the next period; Time Guard absent degrades to the 30-day reference. State machine and getFieldOrganicState shape unchanged.
 - [x] organic_transition_timeguard_test.lua at 11 assertions; suite 2897/0; deployed 2.5.0.71.
 - [~] In-game (owed): a transition at a 30-day month counts down over three in-game years, not eight months.
+
+## 2026-08-14 (Fred): organic compost production, the managed process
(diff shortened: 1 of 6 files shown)


sasha-rf (approved):
Approved. The managed compost production process.

It owns production only, which is the right boundary: batches decompose on the Time Guard clock, yield the existing COMPOST fill type into farm storage, and SF's own amendment path does the OM work, unchanged. No new fill type, no new soil authority, no placeable. The organic-safe flag is feedstock-derived and a biosolids-fed batch correctly stays not-cert-safe, with the breach riding SF's existing onInputApplied rather than inventing a parallel path.

Server-authoritative with the ledger + own-XML persistence, and the Time Guard-absent day-tracking fallback keeps the process honest without a private clock. Bench is 25 new assertions covering the lifecycle, the flag, readiness, collect, refusal, persistence, and the client gate. Full suite 2922/0. The yield and days ride the balance pass, flagged as they should be. Not yet seen in a running game.

PR #843, opened 2026-08-14
feat(SF-57): establishment causes - frost kill, seedbed weight, cover-crop winterkill

TheCodingDad-TisonK:
SF-57, establishment causes: a hard frost, a rough seedbed, and a winter-killed cover crop. All ride SF-18's locked binary threshold-kill frame (no graded scale, no snap counter, a killed cell is killed).

Frost is a second threshold kill: a hard frost (WeatherGuard's published field temperature, the SF-49 pcall pattern) during the establishment window kills the seed, one frost is enough, the exact mirror of one soaking is enough. It is an independent cause that runs even when no moisture signal exists, and frost-hardy crops (winter wheat, rye, triticale, barley) are exempt.

Seedbed is a second threshold weight: the ground type at the sowing position is captured once and eases the moisture and frost trip points on untilled stubble (0.85), with plowed/seedbed and cultivated neutral at 1.0. A fact about how the seed went in, never resampled during the window.

Cover-crop winterkill: a frost-tender cover crop standing outside its window dies over winter, expressed as the winterkill state with a named notification. The dead-standing shape is left intact so RSF-778's biomass probe still samples it; the crop-density write itself stays a noted refinement.

Two new player-facing notifications ship in all 26 languages. Absent WeatherGuard or a nil temperature read means no frost cause and every other cause continues exactly as SF-18 built it. Vine cold damage stays SF-51's.

Verified: 17 new assertions (sf57_establishment_causes_test.lua): the seedbed-weighted thresholds, the frost trip point, tender-vs-hardy, the nil read, the frost kill firing and closing the window, the hardy crop surviving, and the winterkill for tender cover crops. Full suite 2939/0, syntax and lint clean. Built and deployed as 2.5.0.73. Not yet observed in a running game.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.71</version>
+    <version>2.5.0.73</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
         <en>Realistic Soil &amp; Fertilizer</en>
--- a/src/EstablishmentFailure.lua
+++ b/src/EstablishmentFailure.lua
@@ -36,6 +36,32 @@ EstablishmentFailure.WATERLOGGED_MOISTURE = 0.85
 EstablishmentFailure.COMPACTION_THRESHOLD_KNOCK = 0.20
 -- Floor on the effective threshold after the compaction knock.
 EstablishmentFailure.THRESHOLD_FLOOR = 0.5
+-- SF-57 FROST: a hard frost during the establishment window kills the seed,
+-- one frost is enough (the exact mirror of one soaking is enough). The read is
+-- WeatherGuard's published field temperature (SF-49's pcall pattern), sampled by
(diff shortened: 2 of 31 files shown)


sasha-rf (approved):
Approved. Establishment causes: frost, seedbed, winterkill, all inside SF-18's locked binary frame.

The key discipline is the frame: no graded scale, no snap counter, a killed cell is killed, exactly the mirror of SF-18's one-soaking-is-enough. Frost is an independent threshold kill that runs even without a moisture signal, which is the right read of the brief, and the seedbed weight is captured once at sowing as a fact about how the seed went in. The tender-hardy table with the ruled anchors keeps winter wheat safe and lets a tender cover crop winterkill. Absent WeatherGuard means no frost cause and nothing else changes, which is the honest degrade.

The two notification keys ship across all 26 languages, and the winterkill density-write is honestly left as a noted refinement rather than forced in. Bench is 17 new assertions covering the thresholds, the kill, the hardy exemption, and the winterkill. Full suite 2939/0. Not yet seen in a running game.

sasha-rf (approved):
Re-approving after the branch update (development merged in for the version line; the SF-57 code is unchanged from the prior approval). Approved: the frost threshold kill, the seedbed weight, and the cover-crop winterkill, all inside SF-18's locked binary frame. Bench 2939/0.

PR #844, opened 2026-08-14
feat(drilling-window): the advisory line - good or risky window for putting seed in this ground

TheCodingDad-TisonK:
The drilling-window advisory: SF's own field-info surface now says whether the coming days are a good or risky window for putting seed in this ground.

It speaks from the SCS rain outlook (the getRainOutlook sky-reading, clouds now + rain now + the season's habits, always approximate by law) over the establishment horizon, season-scaled through the delivered SoilDuration mechanism, and the ground moisture against the establishment kill condition. Three hedged verdicts, never a promise: good window, risky with the reason, and the weaker forecast-only form when the moisture read is unavailable. The line renders in SF's own per-field field-info surface, so it stands alone with no tablet; the FarmTablet mirrors the hub's read of the same surface.

Advice never gates or writes, and it is silent when SCS is absent, never guessing. The three strings ship in all 26 languages. This is the SF surface half of the drilling-window brief; the SCS getRainOutlook getter it reads is already merged.

Verified: 5 new assertions (drilling_advisory_test.lua): silent without SCS, good on a clear outlook, risky on wet ground with rain likely, the forecast-only degrade, and the silent outlook-error path. Full suite 2944/0, syntax and lint clean. Built and deployed as 2.5.0.74. Not yet observed in a running game.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -135,3 +135,8 @@
 - [x] A managed compost process (no placeable): batches commit farm organic waste, decompose over in-game days on the Time Guard clock (SF day-tracking fallback when absent), and yield the existing COMPOST fill type into farm storage. SF owns the OM effect unchanged; organic-safe flag per feedstock (biosolids-fed is not cert-safe, breach rides onInputApplied). Server-authoritative, StateLedger + own-XML persistence, sfCompostStatus/Start/Collect console.
 - [x] compost_manager_test.lua at 25 assertions; suite 2922/0; deployed 2.5.0.72.
 - [~] In-game (owed): start a batch, watch it decompose across days, collect the compost, spread it on a field.
+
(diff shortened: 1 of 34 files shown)


sasha-rf (approved):
Approved. The drilling-window advisory.

It says what a farmer looking at the sky already thinks, from the SCS sky-reading over the season-scaled horizon and the ground moisture against the kill condition, and it hedges by law: three verdicts, never a promise. Advice never gates or writes, and it is silent when SCS is absent, never guessing, which is the honest foundation. The line lives in SF's own field-info surface so it stands alone without the tablet, and the three strings ship in all 26 languages.

Bench is 5 new assertions covering the silent path, the good and risky verdicts, and the forecast-only degrade. Full suite 2944/0. Not yet seen in a running game.

PR #846, opened 2026-08-14
feat(SF-55): traffic on wet ground - F111 enumeration, wetness input, trafficDrag layer

TheCodingDad-TisonK:
SF-55 traffic-wet-ground. Three parts, per the brief:

- F111 fix: the compaction pass now enumerates every server-side vehicle (wheel-on-ground gated) instead of getPlayerVehicle(), closing the dedicated-server-dead and listen-server-host-only defect for every actor.
- Wetness-input substitution at both compaction call sites: SoilCompactionModel now receives a positional blend of SCS getMoisture with the rain-scalar fallback; scoring math untouched.
- trafficDrag layer: a new SoilValueMaps layer bounded 0.0-0.3, +0.05/event on wet ground over a standing crop, deduped once per cell per day on TimeGuard monotonicDay. The read-time composition with SF-14's yieldEfficiency is documented, not wired (the second-writer fence holds; the addendum is owed on SF-14's brief).

The four Engineering confirms are decided with documented defaults: standing-crop test via getFruitTypeIndexAtWorldPos, blend rule max(scsMoisture, rainScalar), wetness threshold 0.5, SF-52 mask interaction deferred. Dials awaiting the spine.

Verified: 43 new assertions (traffic_drag_test.lua), suite 2962/0. Built and deployed as 2.5.0.76. Not yet observed in a running game.


sasha-rf (approved):
Approved. SF-55 traffic-wet-ground.

F111's all-vehicle enumeration is the biggest player-facing correction and it lands server-side, so compaction finally tracks every farmer, AI helper and hired worker on a dedi. The wetness substitution keeps the compaction math untouched and the trafficDrag layer obeys the second-writer fence, with the read-time composition documented rather than wired into SF-14. The four confirms carry documented defaults, which is the right way to handle open items in a build.

Bench is 43 new assertions and the suite is 2962/0. Not yet seen in a running game.

PR #847, opened 2026-08-14
feat(SF-14): zone yield of the established crop - per-cell capture and the harvest read

TheCodingDad-TisonK:
SF-14 zone-yield of the established crop, the payout half of the 2m growth family.

Per-cell growth-time capture: each cell's own N/P/K is scored through the same single source of truth computeYieldModifier uses, clamped to the band, written to the yieldEfficiency layer on a Time Guard simulation accrual (server-only). The harvest read: on the spatial path, the hook reads the captured growth-time-static layer per cell and applies SF-25's area-weighted positional integral fresh each pass; when the spatial read cannot answer it falls through to the untouched field-average path. SF-55's drag composition line is applied from day one, nil drag = zero.

Calibration invariant bench-proven: on a uniform field each captured cell equals computeYieldModifier exactly, the area-weighted read of a uniform layer returns that value, and the band's floor/ceiling bounds the residual divergence.

The bench-gate items (harvest-read cost, consumer-list re-audit) stay the bounced-to-Tyson items. Verified: 37 new assertions (zone_yield_sf14_test.lua), suite 2956/0. Built and deployed as 2.5.0.77. Not yet observed in a running game.


sasha-rf (approved):
Approved. SF-14 zone-yield of the established crop.

The per-cell capture scores each cell through the same single source of truth the yield modifier uses, and the calibration invariant is bench-proven on a uniform field, which is the thing that matters most here. The harvest read uses the captured growth-time layer with SF-25's area-weighted integral, falling back to the untouched field-average path when the spatial read cannot answer. SF-55's drag composition line is applied from day one, nil drag = zero, so the two members agree.

The bench-gate items stay honestly bounced to Tyson. Bench is 37 new assertions, suite 2956/0. Not yet seen in a running game.

sasha-rf (approved):
## Review: SF-14 zone yield (approval refresh)

Re-approved. The PR head changed after our merge-of-development resolution was pushed (46d9c8bf); the merge brought the branch current with development and resolved the version/doc conflicts to 2.5.0.77. This refreshes the earlier approval.

- The merge is clean and the PR is MERGEABLE. No conflict markers.
- Self-test suite green at the head: 2987/0 across 74 files (the esc-freeze merge plus the SF-14/ SF-55 family).
- The feature is unchanged: per-cell growth-time yieldEfficiency capture and the freeze-supersession harvest read.

No UTF-16 offenders at the PR head.

PR #848, opened 2026-08-14
Esc RF PDA: Soil PRODUCTS–ROTATION canyon close + shared door freeze

WizardlyPayload:
## Summary
- Esc RF PDA workshop polish on the shared door freeze (nine companions carry the same `RfPdaMenuPage.xml`).
- Soil treatment strip: widen PRODUCTS card to 620x248 so the dark canyon to ROTATION closes with matching 10px family gutters (TREATMENT and ROTATION footprints unchanged).
- Tax Esc Status band: densify removed; every show reasserts XML locks inside the framed Status card (eyes-on still unpaid).
- Includes prior Esc RF PDA chrome / guest work already on this Wizard branch.

## Test plan
- [ ] Esc → RF → Soil: PRODUCTS meets ROTATION with even gaps; three cards still read separately
- [ ] Esc → RF → Tax: Status lines inside the white frame (no empty box / divider cut)
- [ ] Spot other RF Esc pages: no layout regression from the shared freeze
- [ ] Client restart after picking the My Games zips from the local pack wave


sasha-rf (changes requested):
## Review: Esc RF PDA canyon close + shared door freeze (PR #848)

The Esc chrome itself reads clean, but the merge result removes two features that already ship on `development`, and it was missed because GitHub reports the PR as MERGEABLE. I dry-merged the PR head onto live `development` with `git merge-tree` and compared the result tree against the branch.

### BLOCKER 1: the SF-23 reciprocal read is deleted

`SoilFertilityManager:getSoilValueAtWorld` (landed with SF-23, `ade411b0`) is gone from the merged tree. The inner `SpatialNutrients:getSoilValueAtWorld` survives, but the public manager wrapper that SeasonalCropStress calls does not. The PR diff shows the whole method, docstring and all, as a deletion at `src/SoilFertilityManager.lua:2388`.

The merge-base `071aef37` has it. Live `development` has it at `:2516`. The merged tree does not. This is a silent breaking change for SCS's cross-mod read.

### BLOCKER 2: the SF-19 per-pixel pressure resolution is reverted

`HookManager.resolveCellPressure` (landed with SF-19, `f60e625c`) is deleted, and the `installSeeAndSprayHook` call site is reverted to the old cell/field scalar path. That is not an Esc chrome change, it is unwinding shipped spatial fidelity in the See & Spray path. If intentional it needs its own brief and PR, not a ride-along in a UI polish PR.

### Finding 3: main.lua picks up a UTF-8 BOM

The merged tree's `src/main.lua` starts with bytes `EF BB BF`. A BOM at the top of a sourced Lua file is a corruption artifact from the branch's merge/WIP history, not an intended edit. Please strip it.

### What is fine

The PRODUCTS card widening, the RfPdaMenuPage gutter change, and the Tax Status card framing all read consistently with the test plan. No UTF-16 offenders. Version in the merged tree stays `2.5.0.76`, matching `development`.

### Ask

Please restore `SoilFertilityManager:getSoilValueAtWorld` and `HookManager.resolveCellPressure` plus its See & Spray call site to the branch (rebase or re-merge `development` rather than carrying deletions forward), strip the BOM from `main.lua`, and re-push. I will re-review on request.

PR #850, opened 2026-08-14
fix(esc): restore RF_EscCardFrameWide + RF_FwColHeader profiles dropped in the shared-door merge

TheCodingDad-TisonK:
﻿Restores the RF_EscCardFrameWide and RF_FwColHeader GUI profiles that the shared-door esc-freeze merge dropped (the merge shipped RfPdaMenuPage.xml referencing them, but the rfEscProfiles.xml changes were left behind). Without them the wide cards (WorkerCosts/FarmTablet FW/MarketDynamics bands) and FW column headers fall back to the base profile, so the boxed layout renders wrong. The profiles are re-added exactly as they existed on the source branch (verified against 6abc1ce3). Both XML files still well-formed; no encoding changes.

### DIFF
--- a/xml/gui/rfEscProfiles.xml
+++ b/xml/gui/rfEscProfiles.xml
@@ -287,6 +287,16 @@
         <frameBottomColor value="1 1 1 0.85"/>
     </Profile>
 
+    <Profile name="RF_EscCardFrameWide" extends="RF_WcPage">
+        <clipChildren value="false"/>
+        <hasFrame value="true"/>
+        <frameThickness value="1dp 1dp 1dp 1dp"/>
+        <frameLeftColor value="1 1 1 0.85"/>
+        <frameTopColor value="1 1 1 0.85"/>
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
## Review: esc-profiles restore

Approved. Verified at the PR head:

- `RF_EscCardFrameWide` re-added exactly as it existed on the source branch (extends `RF_WcPage`, frame 1dp white 0.85, clipChildren false).
- `RF_FwColHeader` re-added (extends `RF_ColHeader`, textAlignment left).
- Both are now defined in this mod's `rfEscProfiles.xml`, matching the references in the merged `RfPdaMenuPage.xml`. The `RF_WcPage` and `RF_ColHeader` base types they extend still exist.
- The XML is well-formed; UTF-8 BOM preserved; pure additions, no encoding churn.
- This resolves the missing-profile warnings in the game log that made the wide cards (WorkerCosts / FarmTablet FW / MarketDynamics bands) and the FW column headers fall back to the base profile.

PR #851, opened 2026-08-15
fix(esc): Crop Stress table flush - reclaim the 0-height action bar, drop the region toward the cards

WizardlyPayload:
csActionBar has held nothing but a hidden button since the table-hiding consultant swap
became an engine veto, so its 44px reserved space for nothing and started the Crop Stress
table 44px lower than it needed to. Height goes to 0 and the node stays, so any stale
csBtnConsultant binding still resolves.

rfHostTableRegion and its hidden twin csConsultPanel move to 0,0 and take
absoluteSizeOffset 514 instead of 584. RF_TableRegionShell stretches in Y, so the offset is
subtracted from the stretched height and a smaller offset is a taller region: 70px taller.
44 of those were already spent reclaiming the top, so 26px land at the bottom as new travel
toward the FIELD DETAIL and AGRONOMIST cards. Family rhythm, not a kiss.

FIELD DETAIL, AGRONOMIST and PIVOT are unmoved, and RF_CsDetailStrip is not grown.

The same page bytes go to all nine Esc doors on purpose. RfEscBootstrap.ensureDoor returns
early once the page exists, and whichever mod loads first builds it from its own copy, so a
change in one door alone is simply overwritten by whoever wins that race.

Two Soil-only additions that development does not have yet, both verified absent before
being applied:

modDesc: the eight advertised chords now ship nested <binding> defaults, and four actions
become dual ONFOOT VEHICLE. The engine lists an action on the F1 help strip only when it is
active, has a binding, and has visible text; every SF action except Handful shipped without
a binding, so none of them could ever be listed however visible the text was. Key names were
verified against the installed mod corpus rather than guessed (KEY_rightbracket and
KEY_leftbracket are both real; the KEY_] form is not).

SoilFertilityManager: sfShowOnCabStrip makes the cab five and the three openers visible with
GS_PRIO_HIGH so the strip's small normal tier does not cut them, and the one unguarded
TrafficDrag.resolveVehicleList call now falls back to resolving the same list itself. A zip
that carried this manager without src/TrafficDrag.lua indexed nil there once per update tick.
Compaction is the older shipped feature and should survive that module being absent.

Applied onto development's own content rather than copied from the local tree, so the SF-14
zone-yield wiring from #847 is preserved.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.77</version>
+    <version>2.5.0.78</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
         <en>Realistic Soil &amp; Fertilizer</en>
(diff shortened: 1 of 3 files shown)


sasha-rf (approved):
## Review: Crop Stress table flush + in-cab F1 strip (PR #851)

Approved. Verified at the PR head:

- A2 (table flush) matches the Wizard handoff exactly: `csActionBar` height 44px to 0px, `csConsultPanel` and `rfHostTableRegion` position 0,-44px / absoluteSizeOffset 0,584 to 0,0 / 0,514. Cards unmoved. Same hunk pattern as the shared-door freeze.
- A3 (in-cab F1 strip) is the `sfShowOnCabStrip` helper: sets action-event text visibility true and `GS_PRIO_HIGH` priority for the vehicle-context HUD/rate/auto/map/drag/variable-rate events, and swaps the three old `setActionEventTextVisibility(_, false)` calls to the new helper so those events show on the cab strip. The modDesc actionBindings gain explicit KB bindings and the four actions move to the `ONFOOT VEHICLE` category.
- The RfPdaMenuPage.xml and modDesc.xml at the PR head are both well-formed. Self-test suite green at the head: 3024/0 across 75 files.
- modDesc bumps 2.5.0.77 to 2.5.0.78.

This is exactly the handoff's A2 + A3, nothing else. No UTF-16 offenders at the PR head.

PR #853, opened 2026-08-15
Wait for a real join event id before requesting soil sync

WizardlyPayload:
## Summary
- Hold the soil full-sync request until the game has handed out a real network event id.
- Do not spend retry budget, and do not look up delay zero, while that id is still missing.

## Test plan
- [ ] Dedicated join on Judith: soil handshake logs after Joined network game, not a delay-zero crash before ids exist
- [ ] Client and dedicated copies match
- [ ] Start is a separate test; this PR does not claim Start is fixed

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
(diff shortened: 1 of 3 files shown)


TheCodingDad-TisonK (approved):
Reviewed: canAttempt gate on AsyncRetryHandler waits for valid event IDs before spending retry budget. Fixes Lua 1-based table bug (delays[0] nil when attempts=0). Well-engineered.

PR #854, opened 2026-08-16
fix: hoist value-map chunk locals so dedicated join catch-up does not flood

WizardlyPayload:
## Summary
- Hoist the four value-map resync tables above the chunk event class so join catch-up does not index nil.
- Stream format and event ids unchanged. Version 2.5.0.85.

## Test plan
- [ ] Join dedicated: no `NetworkEvents.lua:1983` packetReceived flood
- [ ] Value maps still arrive; no new drift/resync storm

Made with [Cursor](https://cursor.com)

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK (Refined edition: WizardlyPayload)</author>
-    <version>2.5.0.83</version>
+    <version>2.5.0.85</version>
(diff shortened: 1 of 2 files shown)


sasha-rf (approved):
Approved. The Lua scoping bug is real and the fix is correct: locals are only visible from their declaration line onward, so the four tables below `SoilValueMapChunkEvent:run` compiled as globals inside that function, indexing nil on every chunk of a repair stream. Hoisting them above the class puts them in scope for both the chunk event and the checksum event.

No functional changes, no stream format or event ID changes. Version bump to 2.5.0.85.

Good catch from the dedicated playtest.

PR #859, opened 2026-08-18
feat(sf): publish mission.compostManager for the FarmTablet organic app

TheCodingDad-TisonK:
Publishes the CompostManager on the mission handle so the FarmTablet organic app can read batch state cross-mod (getfenv(0) is per-mod scoped). Pairs with the FarmTablet OrganicApp compost display PR.


sasha-rf (approved):
Approved. Verified: `mission.compostManager` is published alongside `g_CompostManager` so the FarmTablet organic app can read batch state cross-mod (getfenv(0) is per-mod scoped). 3024 SF suite assertions pass, staged Lua syntax gate clean.

PR #860, opened 2026-08-18
fix(sf): clear all default key bindings (API-6 input convention)

TheCodingDad-TisonK:
API-6 rule 1: no ecosystem mod ships a default binding. All nine SF actions (SF_TOGGLE_HUD, rate brackets, auto/layer/settings/drag chords, variable rate, handful) had default keys that collided out of the box. Removed; the player assigns them in the game's control settings. Actions stay registered and rebindable.


sasha-rf (approved):
Approved. Verified: all nine SF actions' default keys removed (API-6 rule 1), the actions stay registered and rebindable, the stale "only default" comment dropped, and modDesc.xml is well-formed with zero residual KEY bindings. 3024 SF suite assertions pass.

sasha-rf (approved):
Approved. Verified: all nine SF actions' default keys removed (API-6 rule 1), the actions stay registered and rebindable, the stale "only default" comment dropped, and modDesc.xml is well-formed with zero residual KEY bindings. 3024 SF suite assertions pass.

Issue #228, opened 2026-04-24
sadly not working on Dedi Server but on Local Host and singleplayer


A-intheCode:
### Mod Version

1.9.9.9

### Farming Simulator 25 Version

latest

### Game Mode

Dedicated Server

### What happened?

<img width="696" height="1357" alt="Image" src="https://github.com/user-attachments/assets/44aecd5d-4e7c-40cd-bebc-a3a39094a861" />

The Soil cannot be loaded on Dedi Server but on Local and in Singleplayer. I have tested with the same savegame.

### Steps to Reproduce

Load a map on a Dedi Server. , it is not able to load the soil data. It can only load soil data on single player and local host multiplayer. 

This is probably a Rights issue. Keep in mind Dedi servers have a total different rights structure. You cannot execute console cheat commands. You can only execute Console non Admin COmmands. 

### Log Snippet

```text
[SoilFertilizer] Server: Sending full sync to client (255 fields)
2026-04-24 19:25:55.476 Error: Running LUA method 'packetReceived'.
C://gameservers//s1513718//ls2025//gameProfile/mods/FS25_SoilFertilizer/src/network/NetworkEvents.lua:264: attempt to call missing method 'addDelayedCallback' of table
```

### Other Mods Loaded

Crop Rotation, Autodrive, Precision Farming, Courseplay.

A-intheCode:
Its totally normal.. i did modding in the past, productions as well as vehicle modding and the dedi server commands and API is different compared with the local singleplayer routines. Especially because you dont have Cheat rights and a minimal set of console commands avaliable... you cannot even change the fields with a powertools for example. this works only with workarounds...

A-intheCode:
[Version 1.8.9](https://github.com/TheCodingDad-TisonK/FS25_SoilFertilizer/releases/tag/v1.9.8.0) Is the last Dedi Server working version. I rolled back.

TheCodingDad-TisonK:
Mhm, i will instruct claude to do some comparing between that version and the 1.9.9.9 build.
Around 8:45 you will hear back from me :)

A-intheCode:
I can now confirm that the saved savegame with Version 1.9.9.9 was successfully loaded with 1.8.9.0 and the user admin settings of soilmod where successfully loaded.

TheCodingDad-TisonK:
[...]

PR #861, opened 2026-08-19
fix(mp): drip-feed value map sync on dedicated servers

TheCodingDad-TisonK:
## Summary
- Dedicated servers were sending ALL value map chunks synchronously in one frame, causing 5000ms frame spikes on clients receiving the burst
- Now all servers use the drip-feed dispatcher (40ms between chunks), spreading the load across frames on both server and client
- Added `connection.isConnected` guard to the dispatcher so it stops cleanly if the client disconnects mid-sync (the original reason the synchronous path existed per issue #228)

## Test plan
- [ ] Join a dedicated server - verify value map sync completes without frame drops
- [ ] Verify soil overlay data matches server after sync
- [ ] Disconnect mid-sync and verify no server errors
- [ ] Verify checksum drift detection still works after sync

### DIFF
--- a/src/network/NetworkEvents.lua
+++ b/src/network/NetworkEvents.lua
@@ -2230,16 +2230,13 @@ function SoilNetworkEvents_SendValueMaps(connection, onlyLayerIdx)
         return checksums
     end
 
-    if g_dedicatedServer ~= nil or g_currentMission == nil
-       or g_currentMission.addUpdateable == nil then
-        -- Synchronous path (connection lifetime guaranteed - see issue #228)
+    if g_currentMission == nil or g_currentMission.addUpdateable == nil then
         for i, item in ipairs(work) do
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
Reviewed the diff. The `g_dedicatedServer ~= nil` guard was forcing all value map chunks to be sent synchronously in one frame on dedicated servers. Clients receiving this burst processed hundreds of chunk events in a single frame, causing 5000ms spikes.

Removing the guard routes dedicated servers through the same drip-feed dispatcher (40ms between chunks) that listen servers already use. The fallback to synchronous only fires when `g_currentMission` or `addUpdateable` is unavailable, which is the correct last resort.

The added `connection.isConnected == false` check addresses the original reason the synchronous path existed (issue #228): it stops the dispatcher cleanly if the client disconnects mid-sync, preventing sends to a dead connection.

Approved.

PR #863, opened 2026-08-19
fix(burn): exclude cut/withered states from amendment burn risk (#842)

### LINKED ISSUE #842: Harvested annual still reads as a growing crop, so manure on stubble takes the amendment burn
### Mod Version

2.5.0.0

1.21.1.0

### Game Mode

Singleplayer

### What happened?

Spread pelletized manure on a fully harvested soybean field and got `Amendment Warning: Field 136: Organic matter on growing crop - yield -20% at harvest`. Nothing is growing on it — the HUD also shows `Amend. burn risk: Yes` and the yield ramping down.

Harvest doesn't clear the fruit foliage layer, it rewrites it to the crop's cut state. `HarvestMission.lua:303` sets `growthState = fruitDesc.cutState`, and `FieldState:update` (`field/FieldState.lua`) pulls `fruitTypeIndex` straight off the density map, so it still reports `SOYBEAN` on stubble rather than `FruitType.UNKNOWN`. The `FieldState` query in `applyFertilizer` therefore comes back with `hasCrop = true`.

`isAmendmentBurnRisk` then decides whether it actually burns, and the two branches aren't symmetric:

```lua
-- perennial
and not (cutStates and cutStates[gs])) or false
-- annual
return gs >= establishedState
```

The cut-state exclusion only ever landed on the perennial side (#645/#646). `cutState` sits at or above `minHarvestingGrowthState`, so stubble clears `establishedState` and reads as a full canopy. Same helper feeds the HUD burn-risk line, which is why both agree.

### Steps to Reproduce

1. Harvest an annual (soybeans here), leave the stubble, don't cultivate or plow
2. Spread manure / slurry / digestate on it
3. Amendment warning fires and `amendBurnPenalty` starts accumulating

### Fix I'm running locally

Guard both crop classes ahead of the split, using descriptors the engine already exposes (`FruitTypeDesc:getIsCut`, `fruits/FruitTypeDesc.lua:722`):

```lua
if fruitDesc.cutStates and fruitDesc.cutStates[gs] then return false end
if fruitDesc.witheredState and gs == fruitDesc.witheredState then return false end
```

That also makes the perennial branch's own `cutStates` term redundant.

Separate thing I noticed in the same block, not the cause of this one: `hasCrop` samples at `fsField.posX/posZ`, the field's stored reference point, rather than `_lastSprayX/_lastSprayZ` where the spreading actually is. That reads the wrong spot on a partially harvested field.

TheCodingDad-TisonK:
## Summary
- Hoists cut-state and withered-state guards before the perennial/annual split in `isAmendmentBurnRisk`
- Previously only the perennial forage path checked `cutStates`, so applying lime or organic amendments to harvested annual stubble triggered a false burn-risk block
- Also excludes withered crops (no living leaf area to scorch)

Closes #842

### DIFF
--- a/src/SoilFertilitySystem.lua
+++ b/src/SoilFertilitySystem.lua
@@ -5221,16 +5221,15 @@ function SoilFertilitySystem:isAmendmentBurnRisk(fruitTypeIndex, growthState)
     local fruitName    = fruitDesc.name and string.lower(fruitDesc.name)
     local perennialSet = SoilConstants.PERENNIAL_FORAGE_NAMES
     local gs = growthState or 0
+    local cutStates = fruitDesc.cutStates
+    if cutStates and cutStates[gs] then return false end
+    if fruitDesc.witheredState and gs == fruitDesc.witheredState then return false end
     if fruitName and perennialSet and perennialSet[fruitName] then
-        -- Perennial forage: burnable only inside the harvest window (tall sward).
         local minH = fruitDesc.minHarvestingGrowthState
         local maxH = fruitDesc.maxHarvestingGrowthState
-        local cutStates = fruitDesc.cutStates
         return (minH and minH > 0 and gs >= minH
-            and (not maxH or maxH <= 0 or gs <= maxH)
-            and not (cutStates and cutStates[gs])) or false
+            and (not maxH or maxH <= 0 or gs <= maxH)) or false
     end
-    -- Annual: burnable once established past the early seedling stage.
     local minH = fruitDesc.minHarvestingGrowthState or 6
     local frac = (SoilConstants.AMEND_BURN and SoilConstants.AMEND_BURN.ANNUAL_SEEDLING_FRACTION) or 0.33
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
Review PR #863 for FS25_SoilFertilizer.

This PR fixes isAmendmentBurnRisk to exclude cut and withered crop states from amendment burn risk. Previously only the perennial forage path checked cutStates, so applying lime or organic amendments to harvested annual stubble triggered a false burn-risk block.

The fix hoists the cut-state and withered-state guards before the perennial/annual split so both paths benefit. The perennial path's internal cutStates check is removed since it's now handled earlier.

Check:
1. Cut-state exclusion works for both annuals and perennials
2. Withered-state exclusion is correct (fruitDesc.witheredState comparison)
3. Perennial harvest-window logic still correct after removing the redundant cutStates guard
4. No early-return changes that could break the function's contract

PR #864, opened 2026-08-19
fix(spray): boom-proportional area guard for headland slivers (#845)

### LINKED ISSUE #845: paintBoomStrip guards on midpoint travel but divides by perpendicular swept area, so a headland turn pins a sliver at max nitrogen
Split out of #836 so it gets tracked and built on its own rather than riding on a merged fix. Reported and diagnosed by @antler22 ([comment](https://github.com/Realistic-Farming/FS25_SoilFertilizer/issues/836#issuecomment-5295243062)); confirmed at source on `development` after PR #837 merged. **Not covered by #837**, which fixed how the boom endpoints are derived. This is about the guard, and those lines are untouched.

## What happens

On a headland turn, one frame paints a sliver of ground at maximum nitrogen. Reporter's log from a 36.6 m R4045 pass:

```
BoomPaint field=136 boomLen=36.6m stripArea=51.4m2 seeded=true  dN=0.0002
BoomPaint field=136 boomLen=36.1m stripArea=6.1m2  seeded=false dN=0.0011
BoomPaint field=136 boomLen=36.5m stripArea=13.4m2 seeded=false dN=0.0025
BoomPaint field=136 boomLen=36.6m stripArea=0.2m2  seeded=false dN=0.0020   <-- this one
BoomPaint field=136 boomLen=36.6m stripArea=10.3m2 seeded=false dN=0.0020
BoomPaint field=136 boomLen=36.3m stripArea=12.2m2 seeded=false dN=0.0022
```

Healthy frames on that pass swept 6 to 13.6 m2. That one swept 0.2 m2. At 15.764 ha:

```
scale   = sd.area / areaHa = 15.764 / 0.00002 = 788,200
painted = dN * scale       = 0.0020 * 788,200 = ~1576 ppm
```

against a 100 ppm ceiling on the N layer, so the sliver pins at maximum. Healthy frames land at 26 to 34 ppm.

## Cause

`paintBoomStrip` guards on the wrong quantity. `src/SoilFertilitySystem.lua:4047`:

```lua
local travel = math.sqrt(travX * travX + travZ * travZ)
if travel < 0.05 then
    anchor = nil   -- no forward progress this frame: seed instead
```

`travel` is the midpoint displacement. The dose divides by `areaM2`, the swept parallelogram (`:4057`), which is boom length times the **perpendicular** component of travel. On a headland turn the boom sweeps mostly along its own line: the midpoint still moves far enough to clear the check while the perpendicular component collapses to near nothing (about 5 mm on the frame above). `areaM2 > 0.001` at `:4071` is then the only thing left in the way, and 0.2 m2 walks straight through it.

## Proposed fix

Reporter's, running locally: guard on the strip area rather than the midpoint travel.

```lua
local MIN_PERP_TRAVEL = 0.02   -- meters of progress across the boom line
if areaM2 and areaM2 < boomLen * MIN_PERP_TRAVEL then
    return   -- anchor not advanced, sd not consumed
end
```

Two points about it that are the reason it works, both worth keeping in whatever ships:

- **Return early rather than clamp.** Leaving the anchor unadvanced and the dose unconsumed means the next valid frame's quad spans that ground anyway. That is the same self-healing path a failed dose guard already takes, so no gap appears.
- **No agronomy is lost.** `applyFertilizer` has already booked the frame's product into the field average before the painter runs. Declining the frame costs the per-pixel display map only.

Threshold has daylight on both sides: on a 36.6 m boom it works out at 0.73 m2, which catches the bad frame and leaves the smallest healthy one (6.1 m2) well clear.

## Related, and a correction of ours

This is very likely the general case behind HStein72's "cell goes to absolute maximum on minimum application" on #762. That was put down to the narrow working width collapsing `areaHa`. Narrow width is one route to a small strip area, but it is not the only one and it never explained why the same symptom showed on a wide machine. A 36 m boom on a headland turn gets there too. The shared cause is the strip area collapsing, whatever collapses it.

- Split from #836 (endpoint derivation, fixed in #837)
- Related to #762 (streaking, narrow-tool max-out)

## Verification

Reporter has offered a clean before and after on the headland turn case specifically, which is the hard one to catch. Worth taking them up on it.

TheCodingDad-TisonK:
## Summary
- Replaces the fixed `areaM2 > 0.001` guard in `paintBoomStrip` with `areaM2 > boomLen * 0.02`
- A 36 m boom now requires at least 0.72 m2 of perpendicular travel before painting, filtering out sub-0.2 m2 slivers produced during headland turns
- The old 0.001 m2 threshold let through nearly every turn sliver regardless of boom width

Closes #845

### DIFF
--- a/src/SoilFertilitySystem.lua
+++ b/src/SoilFertilitySystem.lua
@@ -4135,7 +4135,8 @@ function SoilFertilitySystem:paintBoomStrip(fieldId, boomPoints, _fillTypeName, 
         areaM2 = boomLen * (h * 2)
     end
 
-    if areaM2 and areaM2 > 0.001 then
+    local minArea = boomLen * 0.02
+    if areaM2 and areaM2 > minArea then
         local areaHa = areaM2 / 10000
         local scale = sd.area / areaHa
         local vm = self.valueMaps


sasha-rf (approved):
Review PR #864 for FS25_SoilFertilizer.

This PR replaces the fixed 0.001 m2 area threshold in paintBoomStrip with a boom-length-proportional guard: areaM2 > boomLen * 0.02. A 36m boom now requires at least 0.72 m2 of perpendicular travel before painting, filtering out sub-0.2 m2 slivers that headland turns produce.

Check:
1. The MIN_PERP_TRAVEL factor (0.02) is reasonable for filtering turn slivers without blocking legitimate passes
2. boomLen is correctly available at the guard location
3. No regression for narrow implements (boomLen * 0.02 should still be < the seed strip area)
4. The minArea variable is correctly scoped

PR #865, opened 2026-08-19
fix(herbicide): scale weed reduction by active section fraction (#838)

### LINKED ISSUE #838: Herbicide zeroes weed pressure in seconds, then See & Spray stops spraying
On a weedy field, Weed Risk drops from ~20% to 0% within about 15 seconds of a 20-minute job. After that the boom mostly stops applying and the weeds never brown off.

Two things seem to be behind it:

1. `onHerbicideAppliedDirect` sizes the reduction from `wap.usage`. Vanilla sets that once per tick from `speedLimit × usageScale.workingWidth`, and `processSprayerArea` never touches it — so section shutoff and actual speed don't reduce liters. Running See & Spray on a few sections still bills full liters, so the reduction gets credited many times over. `wap.lastChangedArea` looks like the honest number here.

2. In the See & Spray hook, `herbicideDaysLeft > 0` short-circuits ahead of the `FieldState` weed-state read, and the `cellWeed < WEED_THRESHOLD` fallback can override that read when it does happen. cellWeed is usually the field average, so sections cut out over live weed patches. Once (1) drops pressure below 15, everything suppresses.

TheCodingDad-TisonK:
## Summary
- **Liters basis**: `wap.usage` is a full-boom budget that ignores section shutoff. When See-and-Spray suppresses sections, the full-width liters still fed into `onHerbicideAppliedDirect`, crediting many times the actual weed reduction. Now scales herbicide liters by `scratchN / #vww.sections` (active-section fraction)
- **See-and-Spray weed check**: `herbicideDaysLeft > 0` short-circuited ahead of the ground-truth weed-state read, causing all sections to suppress once protection was (prematurely) granted. Ground truth now always runs first; `herbicideDaysLeft` and `cellWeed` heuristics are fallbacks only
- Fertiliser credit path unchanged (scoped out per Arissani)

Closes #838

### DIFF
--- a/src/hooks/HookManager.lua
+++ b/src/hooks/HookManager.lua
@@ -1570,23 +1570,22 @@ function HookManager:installSeeAndSprayHook()
                             if pestSS    then skip = skip or (cellPest    < ssCfg.PEST_THRESHOLD)    end
                             if diseaseSS then skip = skip or (cellDisease < ssCfg.DISEASE_THRESHOLD) end
                             if weedSS    then
-                                local herbicideActive = (fd.herbicideDaysLeft or 0) > 0
-                                local weedsGone = herbicideActive
-                                if not weedsGone then
-                                    -- Ground truth: query the game's weed density map at this exact position.
-                                    -- weedState 0=none, 1-6=alive, 7-9=withered/dying → suppress 0 or >=7.
-                                    local fs = getWeedFieldState(fieldId)
-                                    if fs then
-                                        local uok = pcall(function() fs:update(sx, sz) end)
-                                        if uok then
-                                            local ws = fs.weedState or -1
+                                local weedsGone = false
+                                local groundTruth = false
+                                local fs = getWeedFieldState(fieldId)
+                                if fs then
+                                    local uok = pcall(function() fs:update(sx, sz) end)
+                                    if uok then
+                                        local ws = fs.weedState or -1
+                                        if ws >= 0 then
+                                            groundTruth = true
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
Review PR #865 for FS25_SoilFertilizer.

Two-part fix for herbicide zeroing weed pressure in seconds then See-and-Spray stopping:

Part 1 (HookManager applySingle): wap.usage is a full-boom budget that ignores section shutoff. herbAreaFraction = scratchN / #vww.sections is computed after VWW section collection and applied to sectionLiters in the onHerbicideAppliedDirect call (both primary and multi-tank paths). Fertilizer path is unchanged per Arissani scoping.

Part 2 (HookManager See-and-Spray hook): herbicideDaysLeft > 0 previously short-circuited ahead of the ground-truth weed-state read, causing all sections to suppress once protection was granted. Now ground truth (weed density map query) always runs first. herbicideDaysLeft and cellWeed heuristics are fallbacks only when the field-state read fails or returns an invalid state.

Check:
1. herbAreaFraction initialization (1) and VWW update (scratchN / #vww.sections) are correct
2. herbAreaFraction is applied only to herbicide calls, not fertilizer/insecticide/fungicide
3. Multi-tank block also uses herbAreaFraction
4. See-and-Spray weed check: ground truth runs first, ws >= 0 gates the groundTruth flag
5. Fallback logic: herbicideDaysLeft and cellWeed only used when groundTruth is false
6. No-VWW path: herbAreaFraction stays 1, no regression

Issue #839, opened 2026-08-14
Support map-added herbicide fill types (Roundup, atrazine, etc.)


antler22:
`WEED_PRESSURE.HERBICIDE_TYPES` only lists the vanilla `HERBICIDE` fill type, so any herbicide a map or another mod adds — mine has Roundup and atrazine — isn't recognised. It sprays and drains the tank but doesn't touch weed pressure, and the See & Spray weed sensor ignores it too.

Would it make sense to broaden that table, or make it configurable, the way `FUNGICIDE_TYPES` already covers the named chemicals? Per-product effectiveness would be a nice bonus (glyphosate burndown vs a selective residual), but even just recognising them at 1.0 would help.

sasha-rf:
Confirmed on the facts, and I want to be straight with you about where this one goes, because it is not the same kind of item as your other two.

The gap is exactly as you describe. On `development`, `src/config/Constants.lua:1133`:

```lua
HERBICIDE_TYPES = {
    HERBICIDE = 1.0,
},
```

One entry, against eight in `FUNGICIDE_TYPES` plus the blends that `SoilBlends.lua` registers into it at load. And the consequence you hit is one we have already written down in our own source, in the comment above that blend registration:

> THE FIRST GATE. HookManager's spray dispatch reads FUNGICIDE_TYPES before the field is even resolved; a name absent from it sprays, drains the tank, costs the money and does nothing at all.

That is your Roundup and atrazine, word for word, on the herbicide side of the same gate. So the diagnosis is not in question.

Where it differs from #838 and #842 is that those are wrong code against a decided intent, and I can send them to the build team as fixes. This one is asking what the intent should be, and we deliberately do not let the triage seat answer that. Three separate decisions are stacked in it:

1. Recognising map added and mod added herbicides at all. Closest to a plain gap fill, since a product that drains the tank and does nothing is nobody's design.
2. How they get recognised. A wider static table, a config surface, or pattern matching on fill type category are meaningfully different answers, and the third party fill type question is not herbicide specific. Fungicide, insecticide and fertiliser all read from the same shape of table.
3. Per product effectiveness. Glyphosate burndown against a selective residual is a new mechanic, not a correction.

So this goes to Arissani on the design side rather than to the build team, and it will come back with an answer on shape before anything is written. I am not going to have it quietly built as a one line table edit, because 2 is the decision that actually matters here and a table edit would settle it by accident.

[...]

PR #869, opened 2026-08-21
fix: recognize herbicides by spray-type category (#839)

TheCodingDad-TisonK:
## Summary

- Map-added herbicides (Roundup, atrazine, etc.) register as spray type `HERBICIDE` in the engine's `SprayTypeManager` but were not in the one-entry `HERBICIDE_TYPES` table, so they sprayed, drained the tank, cost the money and did nothing to weed pressure
- At init, `populateHerbicideTypesFromSprayCategory()` scans `g_sprayTypeManager:getSprayTypes()` for all spray types with `isHerbicide == true` and populates `HERBICIDE_TYPES` with any not already explicitly listed
- Explicit entries take precedence (custom effectiveness values); category-matched entries default to 1.0
- No existing lookup sites change; the table is populated before `installAll`, so every hook and system reads the complete set

Design answer: Claude(A) ledger 2026-08-21, Arissani ruling -- recognize by fill-type category matching first, existing table as override and fallback.

Flip condition (from the ruling): if category data proves unreliable on third-party fill types at source, the wider-table shape takes over.

## Test plan

- [x] 3033 suite assertions pass (76 test files, 0 failures)
- [x] Build succeeds (`py build.py --deploy`)
- [ ] Verify in-game with a map that adds custom herbicide fill types (e.g. Illschwang with Roundup)
- [ ] Confirm log shows "Herbicide 'X' recognized by spray-type category" for each map-added herbicide
- [ ] Confirm weed pressure drops when spraying with a map-added herbicide
- [ ] Confirm See & Spray recognizes map-added herbicides

### DIFF
--- a/src/SoilFertilitySystem.lua
+++ b/src/SoilFertilitySystem.lua
@@ -305,6 +305,11 @@ function SoilFertilitySystem:initialize()
         self:warning("FieldManager not available - will try delayed initialization")
     end
 
+    -- Populate HERBICIDE_TYPES from the engine's spray-type registry so map-added
+    -- herbicides (Roundup, atrazine, etc.) are recognized without a manual table entry.
+    -- Explicit entries in HERBICIDE_TYPES take precedence (custom effectiveness values).
+    self:populateHerbicideTypesFromSprayCategory()
+
     -- Install hooks via HookManager
     self.hookManager:installAll(self)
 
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):
Reviewed the diff. Clean implementation of Arissani's design answer for #839: recognize herbicides by spray-type category first, existing table as override and fallback.

The scan runs at the right time: after g_sprayTypeManager is loaded, before installAll installs the hooks that read HERBICIDE_TYPES. Explicit entries are preserved (the `== nil` guard means a manual entry with a custom effectiveness value is never overwritten). Category-matched entries default to 1.0, matching the vanilla HERBICIDE entry.

All eight consumption sites across HookManager, SoilFertilitySystem, and SFNozzleEffects read HERBICIDE_TYPES by name lookup, so they pick up the populated entries without any code changes. The approach is the smallest possible change for the widest coverage.

3033 suite assertions pass. Build succeeds. No existing behaviour changes for vanilla HERBICIDE.

PR #873, opened 2026-08-22
fix: hot-reload conformance, suite-edit/vehicle input, docs truth (2026-08-22)

WizardlyPayload:
fix: hot-reload conformance, suite-edit/vehicle input, docs truth (2026-08-22)

- entry/module latch: g_currentModDirectory/Name latched into module globals with g_modsDirectory fallback so live re-source cannot nil-crash
- hot-reload conformance: class tables declared X = X or {} so Ctrl+R reloads update live instances; raw g_currentModDirectory readers fall back to the latch
- docs: key bindings / HUD behaviour brought in line with the actual modDesc bindings (no phantom default keys)
- RF PDA pager: glyph-free RF_CsPivotBtn buttons anchored bottom-right of the table frame; page keys . / > next and , / < back

Built zips are not part of this change. UTF-8 without BOM on all touched files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,9 +90,9 @@ Three pressure scores (0–100) track threats to each field independently. Left 
 
 All three are visible in the HUD and the full Soil Report. Each can be toggled off in settings.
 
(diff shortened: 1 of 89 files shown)


sasha-rf (approved):
## Review: hot-reload conformance wave (2026-08-22)

Reviewed the diff at HEAD. This PR applies the same mechanical conformance pass used across all 20 ecosystem repos:

1. **Mod-directory latch**: `g_currentModDirectory`/`g_currentModName` are nil on a live re-source (hot reload), so they are latched into module globals on first load with a `g_modsDirectory` loose-folder fallback. All `source()` calls and later directory reads now use the latched value.

2. **Idempotent module declarations**: every module table changed from `Foo = {}` to `Foo = Foo or {}` so a re-source reuses the existing table instead of clobbering live state.

3. **Docs truth**: README keybind claims corrected where the suite removed hardcoded defaults (e.g. Shift+H now belongs to FS25_MasterHUD), pointing players to Controls > Mods for assignment.

The pattern matches the FuelCosts reference implementation. No behavior change on first load; changes only matter when a mod is hot-reloaded. Spot-checked main.lua and several modules, all consistent.

Verified mergeable, correct base (development), no draft. UTF-16 scan clean.

PR #887, opened 2026-08-26
SF-40: Wire soil test kit consumer

TheCodingDad-TisonK:
## Summary
- Replaces hardcoded `return false` in `HandfulRead.readTestKit()` with a delegate-when-present call to ProStaffManager.hasSoilTestKit
- pcall-guarded, neutral-false when ProStaff absent -- no new dependency

## Test plan
- [ ] Load without ProStaffCoOp installed -- testKitActive should be false
- [ ] Load with ProStaffCoOp -- testKitActive should reflect the actual kit state

### DIFF
--- a/src/HandfulRead.lua
+++ b/src/HandfulRead.lua
@@ -148,11 +148,16 @@ local function readDiseaseKnown(ctx, soilSystem, field)
     return false, "field"
 end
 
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):

PR #891, opened 2026-08-26
fix(880): defer and retry fresh-save field scan when fields load late

### LINKED ISSUE #880: Starting Spring mod: every field initializes to zero nutrients on first load
Delta South E2 (34:16): opening Delaplaine with the Starting Spring mod active, "every field was absolutely devoid of any kind of nutrient, right down to zero." A different save (Paoli, FSG group) with the same mod combination initialized fine, so the first-load seeding likely races mods that alter mission start date/time rather than failing deterministically.

Verification: fresh save with and without Starting Spring on the same map, compare SoilFieldInfo after first load. Interim workaround if confirmed: SoilVmReseed (with the caveat that it seeds from field averages plus noise).

TheCodingDad-TisonK:
## Fixes #880

Starting Spring (or any mod that alters the mission-start lifecycle) can make the
first field scan run while `g_fieldManager.fields` is still empty. `fieldData`
then stays empty and the post-load seeds loop over nothing, so every field reads
at zero on the maps.

**Fix:** `scanFields()` now arms a pending flag on every deferral, and a bounded
retry pump in `update()` re-runs the scan at 1 s cadence until the fields appear
or the ~20 s window closes, then replays the post-load seed steps (zone data +
GRLE + value maps) a normal fresh save would have had. The client path is
untouched: it returns done immediately and never arms pending.

New `deferred_scan_retry_spec_test.lua` at 17 assertions. Suite 3050 / 0 across
77 files, syntax and lint clean. Deployed for an in-game check (fresh save with
and without Starting Spring, compare `SoilFieldInfo` after first load); not yet
verified in a game session.


sasha-rf (approved):
Reviewed the full head diff.

One item: #880. Two commits, the fix and its ROADMAP/TODO entries.

The mechanism is real: scanFields() runs once in initialize(); if g_fieldManager.fields is still empty (mission-start lifecycle mods like Starting Spring), it used to return false and nothing ever re-scanned, so fieldData stayed empty and the post-load seeds looped over nothing - every field reads zero on the maps. The fix arms a pending flag on each deferral, adds a bounded retry pump in update() (1 s cadence, ~20 s window, then gives up to the lazy getOrCreateField path), and on a deferred success replays the post-load seed steps (zone data + GRLE + value maps). Client path untouched: on MP clients scanFields returns done immediately and never arms pending.

17 assertions in deferred_scan_retry_spec_test.lua covering the deferral, the pump cadence and give-up, the reseed on deferred success, and the client no-op. Suite 3050/0 across 77 files, syntax and lint clean. UTF-16 scan at head: clean. Base is development, one item. The remaining verification is the fresh-save in-game check (with/without Starting Spring), which is Tyson's to run.

Approving.

Issue #879, opened 2026-08-24
Crop vanishes from vanilla PDA at first growth stage while Soil & Fertilizer still tracks it


WizardlyPayload:
Community report (Delta South E3 comments, user tshoff1013): grass/oats drill and start normally; at the crop's first growth stage the vanilla PDA shows no crop planted while Soil & Fertilizer reports the crop as planted and growing; advancing months changes nothing. The reporter asks about conflicts with moisture/weather mods.

Suspect family: SF-18 EstablishmentFailure and the SF-2M growth machinery (GrowthBlock captures at START_GROWTH_PERIOD) interacting at the growth-stage transition. Same neighbourhood as the invisible unplantable cells. Needs a reproduction with the growth_modulation release-gate state recorded, on a map plus mod set matching the report.

sasha-rf:
Triaged. Suspect is the SF-18 EstablishmentFailure and SF-2M growth machinery interacting at the growth-stage transition. Same code neighbourhood as #878 (unplantable cells); likely the same root cause.

Classified **FAST TRACK** and routed to engineering. Needs reproduction with the `growth_modulation` release-gate state recorded. PLAYER-REPORTS row added.

sasha-rf:
Thanks for the report. We found and fixed it.

When a seedling fails to establish (the seedbed gets waterlogged or a hard frost hits during the establishment window), the crop is set to "never came up" on the field. The problem was that our mod kept its own record of the crop as planted and growing even after that happened, so the Soil & Fertilizer readout disagreed with the vanilla PDA, which correctly showed bare ground.

The fix clears our record when the whole stand fails, so both views now agree that the crop never came up. It also hardens the kill write itself, so an engine hiccup during it can no longer leave the mod's growth updates frozen for the rest of the session.

Built and deployed, will land in the next release.

TheCodingDad-TisonK:
Fixed on the development build (PR #892). A full-stand establishment kill reset the growth window but never cleared the crop's "sown" flag, so we kept reporting a planted and growing crop that the vanilla PDA correctly showed as gone. Full kills now clear the display bridge; partial kills keep the crop. It will be in the next release.

sasha-rf:
[...]

PR #892, opened 2026-08-26
fix(878,879): harden establishment kill write and clear dead crop

### LINKED ISSUE #878: Tiny unplantable cells peppered across fields; no planter (including base game) can drill them
Delta South Grain & Harvest E6 (7:20-9:20): after planting corn and soybeans, tiny cells scattered across the fields refuse drilling. Reproduced with a base-game planter on a copy save, so it is ground state, not equipment. Invisible in the world, visible in the PDA coverage. Related report from the same episode's comments: crops drill fine but at the first growth stage the vanilla PDA shows no crop while Soil & Fertilizer still tracks it as planted, for months.

Suspects, ranked: (a) the SF-18 establishment-kill write (EstablishmentFailure.lua:621-668, executeSet(0) over state channels 1..firstGreen-1) leaving inconsistent fruit-plane state on this map; (b) the map's own field/farmland structure (Delaplaine requires the Precision Farming configurator; the same player's rice issue proved map-side); (c) interaction with #877.

Verification plan (copy save): cultivate one patch and retry drilling; SoilVmRead at a patch cell plus a raw fruit-plane density read; re-test with the mod disabled on a fresh field. Note: the pcall in killRegion can skip setIgnoreDensityChanges(false) on a mid-write error; worth hardening regardless of the verdict.

TheCodingDad-TisonK:
## Fixes #878 and #879

Both reports trace to the SF-18 establishment kill write. Two fixes.

**Hardening (both issues):** `killRegion` set `setIgnoreDensityChanges(true)` and
`(false)` inside one `pcall`, so a mid-write engine error left the growth system
ignoring density changes for the rest of the session - freezing crop visibility
and re-drilling. The toggle release now runs unconditionally after the `pcall`.

**State (issue #879):** a whole-stand kill reset the establishing window but
never cleared `field.sownCrop`, so SF kept reporting a planted/growing crop the
vanilla PDA correctly showed as absent (state 0 on the fruit plane). A new
`_recordFullKill` clears the display bridge at every full-kill site (frost and
waterlogged, whole-field and all-cells-killed). Partial kills keep the crop while
surviving cells establish.

New `sf18_kill_state_spec_test.lua` at 10 assertions (toggle always released on a
throwing write, full-kill bridge cleared, partial-kill crop survives). Suite 3060
/ 0 across 78 files, syntax and lint clean. Deployed. The fruit-plane/planter
mechanism behind the unplantable cells still needs the copy-save in-game check to
fully close #878; this lands the two safe, root-adjacent fixes first.


sasha-rf (approved):
Reviewed the full head diff.

One item: #878/#879. Two commits, the fix and its ROADMAP/TODO entries.

The fix is exactly as described. Two defects in the SF-18 establishment kill:

1. killRegion armed setIgnoreDensityChanges(true) and released (false) inside the same pcall, so a mid-write engine error left the growth system ignoring density changes for the rest of the session - freezing crop visibility and re-drilling. The release now runs unconditionally after the pcall; the write error is still swallowed so the daily pass survives.

2. A whole-stand kill reset the establishing window but never cleared field.sownCrop, so SF kept reporting a planted/growing crop the vanilla PDA correctly showed as absent (state 0). _recordFullKill clears the display bridge at all four full-kill sites; the partial-kill path keeps the crop while surviving cells establish.

10 assertions in sf18_kill_state_spec_test.lua cover the throwing-write toggle release, the full-kill bridge clear, and the partial-kill crop survival. Suite 3060/0 across 78 files, syntax and lint clean. UTF-16 scan at head: clean. Base is development, one item. The remaining verification is the copy-save in-game check for the fruit-plane/planter mechanism behind #878, which is Tyson's to run.

Approving.

PR #894, opened 2026-08-27
Add Soil HUD hide/show toggle to the Control Center

TheCodingDad-TisonK:
MasterHUD owns the physical suite HUD keys, so SF_TOGGLE_HUD's native binding stays gated. This adds an SF_TOGGLE_HUD delegate to the Realistic Farming Control Center (calling SoilHUD:toggleVisibility directly, live Hide/Show caption), so a player can hide just the Soil HUD. Composes with MasterHUD's suite-wide # hide.


sasha-rf (approved):

PR #904, opened 2026-08-30
fix(maps): clamp addPaintStrip saturation band (SF-30)

TheCodingDad-TisonK:
**Extracts:** the SF-30 saturation-band clamp for `SoilValueMaps:addPaintStrip` from the contained #899 bundle.

**What it fixes:** `addPaintStrip` filtered out pixels that would overflow the raw range (protecting the no-data sentinel) but returned the full semantic delta as if the whole dose had landed. A large dose silently did nothing to already-stocked ground while the caller believed it applied. Now the pixels the add had to skip are clamped to the ceiling (positive delta) or the layer floor (negative delta), so the return value matches what actually landed.

**Leaves behind:** all other #899 work. No prescription/rate-control behavior, no Precision Pack, no keybindings, no console command, no playtest docs. This is a single, self-contained correction to the existing strip painter.

**Verification:** Lua 5.1 syntax gate passes; pre-commit lint clean; diff is one file, 29 insertions.

### DIFF
--- a/src/maps/SoilValueMaps.lua
+++ b/src/maps/SoilValueMaps.lua
@@ -567,6 +567,35 @@ function SoilValueMaps:addPaintStrip(key, sx, sz, wx, wz, hx, hz, delta)
         self.hasExecuteAdd = false
         return 0
     end
+
+    -- [SF-30] Saturation band: clamp the pixels the add had to skip.
+    --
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):

PR #906, opened 2026-08-30
fix(SF-905): exempt wheel-flattened foliage from amendment burn

### LINKED ISSUE #905: Tire-track foliage state still takes the amendment burn, closing the only lime window on year-round perennials
### Mod Version

2.5.0.0

1.21.1.0

### Game Mode

Singleplayer

### What happened?

Follow-up to #842. That fix closed the *cut state* half of the hole; the **tire-track state** is still open, and for year-round perennials it closes the only application window they have.

Nearly every foliage definition declares a wheel-flattened state right after the cut state, and it is explicitly **not** a cut state:

```xml
<!-- FS25_CrestonNE/maps/foliage/mint/mint.xml -->
<foliageState name="cut"        ... isCut="true"  isDestructibleByWheel="true" isMulched="true"/>
<foliageState name="tiretracks" ... isDestructedByWheel="true" isDestructibleByDisaster="true"/>

<!-- alfalfa.xml -->
<foliageState name="5 cut"    isCut="true"  .../>
<foliageState name="6 TRACKS" isCut="false" isDestructedByWheel="true" .../>
```

`isCut="false"` (or simply absent) means `fruitDesc.cutStates[gs]` is `nil` there, so the guard added in #842 does not cover it:

```lua
if fruitDesc.cutStates and fruitDesc.cutStates[gs] then return false end
if fruitDesc.witheredState and gs == fruitDesc.witheredState then return false end
```

After mowing, a field is a **mix** of cut cells and track cells — the tracks are exactly where the mower and the baler drove. `applyFertilizer` samples one point (`FieldState:update`) to decide `hasCrop` / `burnExempt`, so whether you take an 80% lime burn on a freshly-cut field is down to whether that single sample lands in a wheel rut. In practice on a mowed field it very often does.

Worked example, mint on CrestonNE (`minHarvestingGrowthState = 5`, so `establishedState = max(2, ceil(5 * 0.33)) = 2`):

| gs | state | `cutStates[gs]` | `isAmendmentBurnRisk` |
|---|---|---|---|
| 5 | `harvestReady` | – | true (correct) |
| 6 | `cut` | true | **false** (correct, #842) |
| 7 | `tiretracks` | nil | **true** — burns |

Miscanthus (`7 cut` / `8 TRACKS`) behaves identically. Alfalfa escapes only by luck: its `6 TRACKS` sits above `maxHarvestingGrowthState = 4`, so the perennial branch happens to return false — the annual branch has no such backstop.

Why this bites harder than the stubble case in #842: mint, miscanthus and alfalfa stand on the field year-round. The cut window is the *only* time there is no canopy, so it is the only time you can lime or top-dress at all. Losing the track cells inside that window means these crops effectively cannot be limed for the life of the stand. The same helper drives the HUD, so `Amend. burn risk: Yes` also shows while parked in the tracks of a field that was cut an hour ago.

### Steps to Reproduce

1. Sow mint or miscanthus (any map with a `isDestructedByWheel` track state — CrestonNE, or vanilla barley/canola/maize all have one)
2. Mow it, then drive the field so wheel tracks are laid down
3. Park in a track and check the Soil Monitor: `Amend. burn risk: Yes`, on a field with no canopy
4. Apply LIME / LIQUIDLIME → `amendBurnPenalty` ramps toward `LIME_MAX` (-80%)

### Fix I'm running locally

The engine already resolves this state for us. `FruitTypeDesc.loadFromFoliageXMLFile` parses `isDestructedByWheel="true"` into a single scalar (`fruits/FruitTypeDesc.lua:295-299`, initialised nil at `:147`), so no name matching or per-map table is needed:

```lua
if fruitDesc.cutStates and fruitDesc.cutStates[gs] then return false end
if fruitDesc.witheredState and gs == fruitDesc.witheredState then return false end
-- Wheel-flattened foliage ("tire tracks") is not a canopy either.
if fruitDesc.wheelDestructionState and gs == fruitDesc.wheelDestructionState then return false end
```

Only one state per fruit type can carry the flag (the engine logs an `xmlWarning` and overwrites otherwise), and I checked every foliage XML shipped by the base game data and by CrestonNE — it always lands on a track / stubble / harvested / destroyed state, never on a productive one. Grass and meadow set it on the `cut` state itself, so for those it is a no-op. So the guard cannot accidentally exempt a standing crop.

Two related states, flagged but not changed, since I'd rather you make the call:

- **`fruitDesc.disasterDestructionState`** (same parse block, `:307-311`, but initialised to `0` not `nil`, so it needs a `> 0` test). A hail-flattened crop is arguably as much "no canopy left" as a wheel rut.
- **The single-point sample.** #842 already noted `hasCrop` reads `fsField.posX/posZ` rather than `_lastSprayX/_lastSprayZ`. Even sampling the right spot, one point deciding a whole-field penalty is what makes the cut/track mix a coin flip. A cut-state or track-state hit anywhere in the sprayed area arguably ought to exempt the pass.

TheCodingDad-TisonK:
Closes #905 (follow-up to #842).

## Problem
`isAmendmentBurnRisk` only exempted `cutStates` and `witheredState`, so a **tire-track** cell (`isCut="false"`, `cutStates[gs]` nil) fell straight through the #842 guard. After mowing, a field is a mix of `cut` and `tiretracks` cells, and `applyFertilizer` decides the whole-field burn from a single sample point, so liming a canopy-free field took the -80% `LIME_MAX` burn whenever the sample landed in a wheel rut. This hits year-round perennials (mint, miscanthus) hardest: the cut window is their only application window, so they effectively could not be limed for the life of the stand. The same helper drives the HUD, so `Amend. burn risk: Yes` also showed in the tracks of a freshly-cut field.

## Fix
Guard on `fruitDesc.wheelDestructionState`. The engine parses `isDestructedByWheel` into a single scalar (`FruitTypeDesc.lua:149` nil-init, `:294-298` parse), can only set it on **one state per fruit** (it logs an overwrite warning otherwise), and it always lands on a track/stubble/harvested state, never a productive one. On grass/meadow it sits on the `cut` state, so it is a no-op there. It therefore cannot exempt a standing crop.

```lua
if fruitDesc.wheelDestructionState and gs == fruitDesc.wheelDestructionState then return false end
```

One file, +5 lines (guard + comment). Lua 5.1 syntax gate and lint pass.

## Not in this PR (routed to design)
Two related items the reporter flagged are design decisions, routed to Arissani via the ledger, and intentionally **not** touched here:
- `disasterDestructionState` — should a hail-flattened crop be lime-safe? (needs a `> 0` test, inits to 0 not nil)
- The single-point sample deciding a whole-field penalty (making any cut/track cell in the sprayed area exempt the pass).

### DIFF
--- a/src/SoilFertilitySystem.lua
+++ b/src/SoilFertilitySystem.lua
@@ -5343,6 +5343,11 @@ function SoilFertilitySystem:isAmendmentBurnRisk(fruitTypeIndex, growthState)
     local cutStates = fruitDesc.cutStates
     if cutStates and cutStates[gs] then return false end
     if fruitDesc.witheredState and gs == fruitDesc.witheredState then return false end
+    -- Wheel-flattened foliage ("tire tracks") is not a canopy either, so a rut on a freshly-cut
+    -- field must not take the amendment burn (#905, follow-up to #842). The engine parses
+    -- isDestructedByWheel into a single wheelDestructionState (nil until set, one state per fruit),
+    -- so this can never exempt a standing crop; on grass/meadow it lands on the cut state (no-op).
+    if fruitDesc.wheelDestructionState and gs == fruitDesc.wheelDestructionState then return false end
     if fruitName and perennialSet and perennialSet[fruitName] then
         local minH = fruitDesc.minHarvestingGrowthState
         local maxH = fruitDesc.maxHarvestingGrowthState


sasha-rf (approved):

PR #908, opened 2026-08-31
Keep sprayer cab panels on hide-all, and show live key hints

WizardlyPayload:
## Summary
- Sprayer, spreader, and harvester cab panels stay useful when suite hide-all is on. Soil Monitor, trails, and minimap still hide.
- On-screen Soil key hints (HUD drag, handful changelog, field guide, settings copy) now show the player's live binds instead of shipped defaults. German Umschalt is not eaten.

## Test plan
- In a sprayer, Right Shift+G: rate / VRA / sprayer info stay; monitor hides.
- Rebind HUD drag and handful; HUD hint and handful changelog should follow the new keys.
- Open the Soil field guide and independent-panels setting; chords should match the live binds.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK</author>
-    <version>2.5.0.85</version>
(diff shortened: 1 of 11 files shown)


sasha-rf (approved):

PR #909, opened 2026-09-01
fix: Esc pivot Fit chips and sweep/rain-key captions

WizardlyPayload:
## Summary
- Fit / Remove and Trip chips on the Esc pivot card, plus Sweep start, Sweep end, and Rain key captions.
- Needed when this companion creates the Esc door. Crop Stress guest still paints the invert and fills the captions.

## Test plan
- [ ] Full start. If this mod wins Esc, Fit works and the three captions show on an owned Reinke.


sasha-rf (approved):

PR #910, opened 2026-09-02
Move Soil HUD drag and auto-rate keys off common mod clashes

WizardlyPayload:
## Summary
- Keep Right Shift as the Soil modifier.
- Move **HUD drag** from Right Shift + P to Right Shift + Up. Right Shift + P is the production dialog in widely used production overlay mods, and both are used in the cab.
- Ship **auto rate** as Right Shift + Insert instead of Right Shift + L. Right Shift + L is Real GPS setup, another widely used cab mod.
- Other Soil actions stay unbound (player assigns those in Controls). HUD toggle stays Right Shift + O.

## Test plan
- [ ] Fresh profile / Controls: HUD drag default is Right Shift + Up, auto rate is Right Shift + Insert.
- [ ] With a production overlay mod loaded, Right Shift + P still opens that production dialog and does not enter Soil HUD drag.
- [ ] With Real GPS loaded, Right Shift + L still opens GPS setup and does not toggle Soil auto rate.
- [ ] Existing saves that already rebound these keys are unchanged (this only changes the shipped default).

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK</author>
-    <version>2.5.0.87</version>
+    <version>2.5.0.88</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):

PR #913, opened 2026-09-02
Show field fungicide resistance on the Soil Esc page

WizardlyPayload:
## Summary
- On the Soil Esc page, an owned field now shows which fungicide modes still offer good, reduced, or no control, plus tank pairs. Unscouted fields read not yet known. Read-only: nothing ranks, buys, mixes, or applies.
- The fields table is one row shorter and a little smaller so Resistance has room. Treatment, Products, and Rotation stay above it with even gaps.
- Scout history is saved. Clients wait for the server picture before showing control words. Experimental / release gating is unchanged.

## Test plan
- Esc, Realistic Farming, Soil. Select an owned field: Resistance card under Treatment, 10px gaps, not sitting on Help/Back.
- Unscouted field: not yet known. After scouting: Good / Reduced / No control per mode. Tank pairs opens the partner list.
- Dedicated server + client: client stays unknown until the first delivery, then matches the server words.
- Map cell click still shows blocked / ahead growth. Experimental Systems off: growth block hidden; Resistance still follows its existing gate.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK</author>
-    <version>2.5.0.91</version>
+    <version>2.5.0.92</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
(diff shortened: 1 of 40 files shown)


sasha-rf (approved):

PR #915, opened 2026-09-03
Bind leftover Soil keys and stop F1 help naming Shift+L

WizardlyPayload:
## Summary
- Playtested Right Shift defaults for leftover Soil actions that were still unbound on development (rate up/down, map layer, settings, variable rate, minimap zoom, scout, treatment, handful).
- Auto-rate stays Right Shift + Insert. HUD drag stays Right Shift + Up. Those two already moved so Real GPS and production-dialog mods keep their chords.
- F1 help no longer names Shift+L. It names Auto-Rate Control so the sentence cannot go stale when the player rebinds.

## Test plan
- Fresh save: leftover actions fire on the new Right Shift defaults.
- Auto-rate is Insert, not L. HUD drag is Up, not P.
- F1 Soil help rate paragraph names the action, not Shift+L.
- Rebind auto-rate in Controls; on-screen live hints follow the new chord; F1 still names the action.


sasha-rf (approved):

PR #916, opened 2026-09-03
fix(seespray): centre honours the See & Spray no-target decision

TheCodingDad-TisonK:
## Problem
On a sprayer with See & Spray purchased (verified on the Agrio DinoII), herbicide/fungicide/pesticide only sprayed from the centre section across the whole field, independent of where weeds/disease actually were. All sections flashed on turn-on, then collapsed to centre.

## Root cause
The per-nozzle See & Spray decision was computed correctly for every nozzle, centre included (a debug trace showed the centre reading disease 9.1 vs threshold 10, i.e. suppress). But the deposition aggregation in onUpdate deliberately excluded the centre (sentinel sectionIndex 0), so spec.sectionActive held no centre entry and getIsWorkAreaActive fell through to return true for the centre work area. The centre therefore kept depositing product on a no-target field while the outer sections correctly held. Visible result: only the mid section sprayed.

## Fix
- Record the centre VWW section real index at load (spec.centerSectionIndex).
- Track the centre nozzle decision separately as spec.centerActive in the per-section aggregation.
- Gate the centre work area (its real section index, or a no-section work area) on centerActive in getIsWorkAreaActive. Safe default (spray) when there is no centre nozzle.

Fertiliser is unaffected (it always reports active, so the whole boom sprays as before).

## Testing
Verified in-game on the Agrio DinoII: below-threshold field, the whole boom now holds (centre no longer leaks); forcing disease over threshold (SoilSetDisease 50) sprays the full boom including centre.

### DIFF
--- a/src/specializations/SFNozzleEffects.lua
+++ b/src/specializations/SFNozzleEffects.lua
@@ -327,6 +327,15 @@ function SFNozzleEffects:onPostLoad(savegame)
         return
     end
 
+    -- Real index of the centre VWW section. Nozzles remap the centre to sentinel
+    -- sectionIndex 0 (so it skips the section.isActive gate), but deposition gating needs
+    -- the centre's TRUE section index so getIsWorkAreaActive can suppress it too when See &
+    -- Spray finds no target - otherwise the centre leaks product while the boom holds.
+    spec.centerSectionIndex = nil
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):

PR #921, opened 2026-09-05
Move Soil HUD-move off the look keys onto Right Shift + End

WizardlyPayload:
﻿## Summary

- Default Soil HUD-move is Right Shift + End instead of Right Shift + Up, so the arrow keys stay look.
- Version 2.5.0.104.
- Esc door, Buy/Run on any host, and Soil map field numbers are already on development. This PR is the bind change only. It does not include the held merged-field Soil list.

Playtested on this PC in Soil 2.5.0.104.

## Test plan

- [ ] Fresh profile: arrow keys still look around on foot and in a vehicle.
- [ ] Right Shift + End still moves the Soil HUD.
- [ ] Esc Realistic Farming door, Buy/Run, and Soil map field numbers unchanged.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK</author>
-    <version>2.5.0.103</version>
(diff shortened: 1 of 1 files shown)


sasha-rf (approved):

PR #923, opened 2026-09-06
Group owned field lists by helper GPS outline

WizardlyPayload:
## Summary
- Owned field lists group by the helper/GPS terrain-detail outline, so Montana 86 and 87 show as one row (Field 86-87) instead of two farmland numbers.
- Soil walks the outline once, caches the blocks, and publishes them on `g_currentMission.RfPdaSoilMerge` for Crop Stress and Farm Tablet.
- The rotation planner steps those same blocks. Map numbers, the handful HUD, and FieldSentry stay per farmland.

## Test plan
- [ ] Full start (new Lua file).
- [ ] Open Soil owned-field list on a save where two numbers share one GPS outline; they should join after the first-open walks finish.
- [ ] Nearby fields that are not on the same outline stay separate.
- [ ] Rotation planner steps the joined block as one field.
- [ ] FieldSentry / map numbers still list each farmland id.
- [ ] Esc Realistic Farming glance pages still load.

### DIFF
--- a/modDesc.xml
+++ b/modDesc.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8" standalone="no" ?>
 <modDesc descVersion="108">
     <author>TisonK</author>
-    <version>2.5.0.119</version>
+    <version>2.5.0.121</version>
     <modName>FS25_SoilFertilizer</modName>
     <title>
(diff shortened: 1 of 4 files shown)


TheCodingDad-TisonK (approved):
Sasha review: GPS-outline row merge follows George's closed design, mission bridge pattern sound, cost model documented. Approved.

PR #926, opened 2026-09-09
feat(SCS-041): additive compaction-grain return on the soil value getter

TheCodingDad-TisonK:
## SCS-041 SF read-contract: additive compaction-grain return

Build-order step 1 of the SCS-041 Thirsty Ground member (One Ground hydrology). SeasonalCropStress needs the value map's metres-per-pixel grain to size its per-cell irrigation absorption budget honestly, so the soil value getter now hands that grain back as an optional second return.

### Changes
- `src/maps/SoilValueMaps.lua`: new `getGrainMetres()` returning `terrainSize / resolution` (the shared per-layer grain, about 2 m/px), nil when the maps are unavailable rather than a fabricated number.
- `src/SpatialNutrients.lua`: `getSoilValueAtWorld` returns `value, grainMetres`. Returns nil (no grain) when the value itself is nil.
- `src/SoilFertilityManager.lua`: passes the grain through the existing pcall wrapper.
- `tools/test/lua/sf23_spatial_nutrients_test.lua`: mock value map now mirrors the real `getGrainMetres` API and asserts the grain return.

### Why it is safe
Purely additive. One-value callers ignore the extra return, so every existing caller is unchanged. The SCS-041 consumer reads this pcall-guarded and treats a missing manager/getter/grain as neutral, so it can land ahead of or behind this PR.

### Verification
`bash tools/test/run.sh`: sf23 green (14/0). The 2 failures in `om_213_organic_premium_test.lua` are pre-existing on `development` (reproduced with this branch stashed) and unrelated to this change.

Release stays LOCKED. In-game observation is owed at member completion, not here.

### DIFF
--- a/src/SoilFertilityManager.lua
+++ b/src/SoilFertilityManager.lua
@@ -2620,7 +2620,9 @@ end
 ---@return number|nil
 function SoilFertilityManager:getSoilValueAtWorld(key, x, z)
     if SpatialNutrients == nil or SpatialNutrients.getSoilValueAtWorld == nil then return nil end
-    local ok, v = pcall(SpatialNutrients.getSoilValueAtWorld, SpatialNutrients, self.soilSystem, key, x, z)
+    -- [SCS-041] Pass through the optional second return (metres-per-pixel grain)
+    -- without breaking one-value callers, which ignore it.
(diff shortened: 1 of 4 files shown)


sasha-rf (approved):

PR #935, opened 2026-09-12
fix(soil): SF-934 tillage nutrient change no longer flips with travel direction

### LINKED ISSUE #934: NPK change flips with travel direction on cultivator (Solek, development)
### Mod Version

NPK change flips with travel direction on cultivator (Solek, development)

development (first seen on 2.5.0.0)

### Game Mode

Singleplayer

### What happened?

Discord report from Fippes, 2026-09-12. No log, no save, no vehicle names.

On Solek, N/P/K values change depending on travel direction. After harvest, a cultivator pass reduced N in one direction and added N in the opposite direction. Reporter said the same thing happens with other machinery.

Retest on the development branch:
- Cultivator problem still present.
- Two passes in different directions reduced N and K.
- A third pass added N and K.

Expected: a cultivator pass should not reverse the N/K (and P) change based on which way the machine is driving.

### Steps to Reproduce

Not a controlled staff repro. Player steps as reported:

1. Load a save on Solek.
2. After harvest, work the field with a cultivator.
3. Drive one direction and note N and K (and P if shown).
4. Drive the opposite direction on the next pass and note N and K again.
5. Make a third pass and note N and K again.

Observed on development: first two directions reduced N and K; third pass added N and K.
Reporter also said this happens with other machinery. No vehicle names given.

### Log Snippet

```text
Not provided. Discord text only. No log.txt and no [SoilFertilizer] lines.
```

### Other Mods Loaded

Solek map. No other mod list provided.

### Anything Else?

Original 2.5.0.0 report also said sprayers, manure spreaders, and slurry spreaders applied only on a small strip, not the full work width.

On development, the sprayer now applies on the full width. Manure and slurry were not reconfirmed.

This ticket is for the remaining NPK-by-direction bug, not the width issue.

Source: Realistic Farming Discord, 2026-09-12. TysonK asked for a development retest and said he would look into the cultivator/NPK report.

TheCodingDad-TisonK:
Fixes #934.

## The bug

`vmLocalBump` (`src/SoilFertilitySystem.lua`) called `SoilValueMaps:addValueAtWorld`, which despite the name is a read-modify-**SET**. It sampled ONE pixel under the implement root, added the delta, then `executeSet` that single result flat across a `CELL_SIZE` (10 m) square, replacing every other pixel in the square with whatever the sampled pixel held.

The square is wider than most implements, so neighbouring passes overlap and each pass copied its own sampled pixel over the overlap. Driving back down the same track sampled a different neighbour, so an unchanged **positive** delta could stamp the square **down**: N and K fell on one pass and rose on the next. The nutrient arithmetic never reversed; the stamp copied a different pixel each time.

The position came from `getWorldTranslation(cultivatorSelf.rootNode)` in `HookManager` - a single point, carrying neither working width nor heading.

All six soil-write paths shared that call (sowing, oxidation helper, plowing, cultivation, strip-till, mulching), which is why the reporter saw it with other machinery too. Spraying was unaffected because it had already moved to `addPaintStrip` in RSF-762 / RSF-836; tillage never made that move.

## The fix

Tillage now makes the same move: paint the swept quad between the last worked line and the current one through `addPaintStrip`, a real per-pixel `executeAdd` with a saturation band. Consecutive ticks share no overlap, so a dose never stacks and no pixel is overwritten with a neighbour's value.

- **`HookManager:_recordTillagePoint`** records the worked **line** (the work-area `start`/`width`/`height` node span, so it carries width AND heading at any angle) plus the field id, not just the root point. All six hook sites route through it.
- **Anchors are keyed per field**, so a pass on one field never links to a pass on another.
- **Mass is conserved.** Callers express a delta as the tick's whole amount concentrated in one zone cell, so the painted per-pixel delta is `delta * CELL_AREA_HA / stripAreaHa` - the same arithmetic `paintBoomStrip` uses.
- **One tick, one footprint.** Cultivation's residue bump and the oxidation bump that follows it in the same tick land on the same ground, instead of the second one reseeding onto a thin strip and concentrating its delta.
- **The fallback is additive too.** First tick, direction reversal, teleport, standing still, field change, and a missing or degenerate work line all seed a thin strip *through `addPaintStrip`*. A point-write fallback would have kept the flattening stamp alive on exactly the ticks this bug appears on: the first tick of every pass and every headland turn.

`addValueAtWorld` keeps a DO-NOT-USE note explaining the set-over-area semantics. It has no callers left.

## Verification

New bench `tools/test/lua/sf934_tillage_strip_direction_test.lua`, **38 assertions**:

- additive in both directions, never a stamp
- the delta keeps the caller's sign driving up AND driving back (a gain stays a gain, a loss stays a loss)
- swept-quad geometry: base is the previous worked line, travel edge is the current one
- mass conservation: twice the swept area carries half the per-pixel delta
- per-field anchors, teleport seeding, stationary seeding
- every fallback path, including no work line and a degenerate span
- a zero delta writes nothing

**Control run:** the bench was executed against the pre-fix code and fails there, 3 passed / 35 failed, including `934: the first tick NEVER stamps a sampled pixel :: got 1 want 0`. It is bound to the new behaviour, not merely passing alongside it.

Gates: `syntax-check` OK (94 files), `lint` clean (94 files), pre-commit staged gate OK. Suite **3759 passed, 2 failed**; both failures are `om_213_organic_premium_test.lua` and are present unchanged on clean `development`.

## Still owed in game

Offline benches cannot witness the real density maps. What remains for the in-game check on Solek: cultivate one direction, note N and K, cultivate the opposite direction, note them again, then a third pass. The values should move the same way every pass. Also worth a look: headland turns (the seeding path) and a pass that crosses a field boundary.

### DIFF
--- a/src/SoilFertilitySystem.lua
+++ b/src/SoilFertilitySystem.lua
@@ -4162,11 +4162,152 @@ end
 
 --- Local read-modify-write bump around a work position (residue/amendment
 --- incorporation at the tillage tool). deltas = {nitrogen=dN, ...}.
-function SoilFertilitySystem:vmLocalBump(worldX, worldZ, deltas, radius)
-    if not worldX or not worldZ or not self:vmAvailable() then return end
+-- [SF-934] Teleport guard for the tillage sweep, in multiples of the worked line
+-- length: a jump farther than this never spans a quad, it seeds a fresh strip.
+SoilFertilitySystem.TILLAGE_TELEPORT_FACTOR = 3.0
+
+--- [SF-934] The parallelogram vmLocalBump paints this tick, plus the anchor
+--- bookkeeping behind it. Split out so every write inside ONE tick shares ONE
+--- footprint: cultivation bumps residue and then oxidation bumps OM in the same
+--- tick, and a second call that reseeded would concentrate its delta onto a thin
+--- strip while the first spread over the full swept quad. The per-tick cache keeps
+--- both on the same ground, which is also what physically happened.
+---@param fieldId number|nil  anchors are keyed per field so a pass never links fields
+---@return table|nil  { sx, sz, wx, wz, hx, hz, areaM2 }, or nil when nothing to paint
+function SoilFertilitySystem:_tillageQuad(fieldId, ax, az, bx, bz, lineLen)
+    local nowMs = (g_currentMission and g_currentMission.time) or 0
+    local cached = self._vmTillQuad
+    if cached ~= nil and cached.time == nowMs and cached.fieldId == fieldId
+        and cached.ax == ax and cached.az == az and cached.bx == bx and cached.bz == bz then
+        return cached.quad
+    end
+
+    local anchors = self._vmTillAnchors
+    if anchors == nil then anchors = {}; self._vmTillAnchors = anchors end
+    local anchor = (fieldId ~= nil) and anchors[fieldId] or nil
+
+    local sx, sz, wx, wz, hx, hz, areaM2
+
+    if anchor ~= nil then
+        -- Tip-swap guard: the work-area node pair can arrive in either order from one
+        -- tick to the next, so pair each tip with whichever anchor tip it is nearer.
+        -- Without it a swapped pair folds the quad into a bow tie and the paint dies.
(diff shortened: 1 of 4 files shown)


sasha-rf (approved):

PR #937, opened 2026-09-13
test(om-213): drive the MarketEngine bench as the host after MD-15

TheCodingDad-TisonK:
## What
Test-only fix for the two red assertions in `tools/test/lua/om_213_organic_premium_test.lua` on development.

MD-15 (MarketDynamics 7521897, merged via PR #148 on 2026-09-08) added a pure-client guard at the top of `MarketEngine:_recalculate`: `if g_server == nil and entry.current ~= nil then return entry.current end` (MarketDynamics origin/development `src/MarketEngine.lua:346`). Neither the SF prelude nor this bench ever set `g_server`, so the engine acted as a pure client and the seeded `current = 10` never moved. Harness drift, not a regression: the composition at `:350-374` is unchanged, and in single player or on a listen host `g_server` is set.

Change:
- `g_server = {}` right before `MarketEngine.new()`, mirroring MD's own MD-15 bench (`MD-15-calendar_quotes_spec_test.lua:41`, `:51`).
- A pure-client witness before the registration lifecycle block: seed a received quote of 12.5, set `g_server = nil`, call `_recalculate`, assert the quote is kept. `g_server` is left nil afterwards.

No MarketDynamics code change. No prelude change on purpose: whether the engine under test is the host or a client is a per-bench decision, and a prelude-wide `g_server` would silently move every other bench onto the server path.

## Evidence
- Before: `om_213_organic_premium_test.lua` 18 passed, 2 failed ("recalculate moves the price for an organic seller :: got 10 want ~11.3333", "clamp B clips the composition band at 3.0 :: got 10 want ~30"). Full suite 3759 passed, 2 failed across 82 files.
- After: 21 passed, 0 failed. Full suite **3762 passed, 0 failed across 82 files**.
- The local sibling MarketDynamics working copy is dirty (branch `fix/MD-15-locked-session-latch`), so the bench was also run in isolation against `git show origin/development:src/MarketEngine.lua` and `OrganicPremiumBridge.lua` (142b627): 21 passed, 0 failed.
- Syntax gate: `check-staged.mjs` passes the test file; a deliberately broken control file blocks with exit 1. Pre-commit hook ran on commit.

## Owed in game
Nothing new from this change. The OM-213 LOCKED acceptance item (the organic premium visible at a sell point) remains owed as before.

Follow-up outside this PR: ROADMAP.md:151 and TODO.md:151 still explain earlier om_213 reds as "sibling MarketDynamics repo absent in the temp clone"; the later "pre-existing om_213 pair red" rows were this MD-15 drift. Left untouched here to keep the PR test-only.


sasha-rf (approved):

PR #942, opened 2026-09-14
feat(handful): test kit shows band word and figure for N, P, K and pH (RSF-F217)

TheCodingDad-TisonK:
## RSF-F217: the soil-test-kit entitlement should show the promised N, P, K and pH readings

Brief: ecosystem-dev-tracking, Office Tyson/RSF-F217-test-kit-quantitative-readout-IMPLEMENTATION-BRIEF.md (v1.0 plus amendment v1.2), certified in RSF-SOIL-WAVE-CD-HANDOFF-CERTIFICATE-2026-09-13.

### What changed
- `src/ui/SoilHandfulDialog.lua`
  - `_populate`: one `local kit = (p.testKitActive == true)` drives the four tiles and the footer. False, absent, `1`, `"true"` or a table keep today's band output.
  - New `kitNutrient(clause, key)`: requires the record shape, a finite `.value`, a Poor/Fair/Good status (case-insensitive), `SoilConstants.PPM_DISPLAY` as a table and a finite factor strictly greater than zero. Converts with `math.floor(value * factor + 0.5)`, rejects a non-finite result, then composes the band word and `sf_handful_ppm` through `sf_handful_kit_tile`. Returns `(text, severity)` from the band helper, or nil so the tile helper draws the dash in neutral grey.
  - New `kitPh(ph)`: finite guard, `string.format("%.1f", ph)`, band word and severity from `bandPh`.
  - New `patternHas` and `kitFormat`: a translated pattern must carry exactly the expected conversions (`%%` is a literal); otherwise the exact Lua fallback is used, and formatting runs under pcall so `_populate` never throws.
  - Header contract and stale comments updated.
- `translations/translation_*.xml`, all 27 files: `sf_handful_ppm` = `%d ppm` and `sf_handful_kit_tile` = `%s %s`, hand-added after `sf_handful_kit_no`. Non-English rows carry the `eh` hash the way lang_sync formats them; `lang_sync.py sync` was not used because it writes `[EN]` placeholders. `sf_hud_unit_ppm` is not reused.
- Not touched: `HandfulRead.lua`, thresholds, the tile helper, layout, OM and every other clause. F219 remains uncommissioned.

### Verification
- `node tools/test/run-tests.mjs`: 3851 passed, 0 failed across 84 files.
- `handful_panel_sf39_test.lua` gains 57 assertions: N50/P45/K40/pH6.7 gives "Good 150 ppm", "Fair 27 ppm", "Poor 160 ppm", "Good 6.7" with band colours; N40 reads "Fair 120 ppm" on wheat and "Good 120 ppm" on soybean; false/nil/1/"true"/table keep the bands and the no-kit footer; per-tile dash in neutral grey for missing, NaN, infinite, string, Unknown status and missing value; zero, missing, negative factor and missing table dash that tile only; a valid 0 renders "Poor 0 ppm"; 27.45 rounds to 27, 0.5 rounds to 1, pH 6.25 shows 6.3; malformed patterns (extra, missing, wrong-letter, empty) fall back without throwing, an escaped `%%` is literal, a valid reorder is honoured; the payload is not mutated.
- `handful_read_sf38_test.lua` gains 4: the reader reports true only when the manager grants the farm, false otherwise, false for a non-boolean grant.
- Locale check by grep: both keys present in 27 files, each `%d ppm` once and `%s %s` once; `lang_sync.py check` shows 0 missing for every locale.
- `py build.py --deploy` built and deployed 2.5.0.128. Native check owed at release: kneel with and without the ProStaff kit, in English and one other language.

### Note for Sasha
The two new values are format patterns, not sentences, so no speaker review is needed for them.

### Bob's cold review, carried
APPROVE. Three minors, none a change request:
- The two new English rows carry no `eh`; the next `lang_sync.py sync` will add them and dirty translation_en.xml. Stale detection recomputes from the English value, so the 26 non-English rows read as translated, not stale.
- `sf_handful_ppm` will be counted untranslated in the 26 non-English files because the token "ppm" is not a pure format string. Report noise only; check does not fail on untranslated.
- `HandfulRead.lua:246` still says "neutral-false until the test-kit member ships". The brief forbids touching the reader, so it stays; it is live doc drift for a follow-up.
- A digit-only tail with no conversion letter (for example `%s %s %2`) passes the pattern check and fails inside string.format; the pcall plus fallback retry catches it, so the two layers together are what make it safe.


sasha-rf (approved):

UTF-16 scan: 30 file(s) read at the PR head, all clean.

PR #943, opened 2026-09-14
fix(tip): ground-tip wrapper honours the engine gates and is switched per mission (RSF-F227)

TheCodingDad-TisonK:
## RSF-F227: Ground-tip permission and wrapper lifetime repair

Brief: ecosystem-dev-tracking, Office Tyson/RSF-F227-ground-tip-override-IMPLEMENTATION-BRIEF.md (v0.4), certified in RSF-F227-ground-tip-override-HANDOFF-CERTIFICATE-2026-09-13.

### What changed
- NEW `src/hooks/GroundTipGate.lua`: the wrapper, the once-per-process install flag, the runtime flag and the captured predecessor. `install()` wraps once and returns true only on the call that installed; `enable()` turns the gates on (no-op unless installed); `disable()` turns them off and leaves the live callable alone. Wrapper body, matching the engine at DensityMapHeightUtil.lua:26-37: flag off, return the predecessor's answer; `mgr:getIsValid()` false, return false (colon call, no nil-manager guard, as the engine does); height type present, `ht.canBeTipped and true or false`; otherwise the predecessor.
- `src/main.lua`: sources the module after HookManager; the file-level `_canTipHookInstalled` and the inline closure are replaced by `GroundTipGate.install()` plus `GroundTipGate.enable()` inside the existing `registered > 0` branch of the `if tmpl` fallback block (brief item 4, the enable never leaves that block); `unload()` calls `GroundTipGate.disable()` as its first statement, outside the `if sfm` block (item 3). The misleading install comment is corrected: the hazard is chaining, the manager is resolved live, the flag is what makes the guard correct.
- The optional identity-guarded removal was NOT taken. Bob found `reloadDlcsAndMods` re-sources main.lua on a normal game path, so the live callable could never match and removal could restore the wrong thing.

### Deviation from the brief, stated plainly
The brief asks for a file-level local in main.lua. main.lua cannot load under the fengari bench, so the required spec test could not exercise a wrapper living there. The module keeps every semantic the brief fixes and adds the accessor SG6 joined mode needs to drive the same flag from another file, which a file-local could never allow. Bob's cold review is judging this; Tyson has the veto.

### Invariants
No material gains or loses an engine permission: injected types carry `canBeTipped = true` at main.lua:445. No change to weights, costs, soil effects, the `registered > 0` install condition, settings, save or network.

### Verification
- `node tools/test/run-tests.mjs`: 3830 passed, 0 failed across 85 files.
- `RSF-F227-ground_tip_lifetime_spec_test.lua`, 39 assertions on the real wrapper with stubs: flag on gives invalid manager false, permitted true, forbidden false, missing field false not nil, no height type delegates, invalid manager never reaches the predecessor; flag off returns exactly the predecessor for every input; lifetime false at declaration, enable before install a no-op, false after an unload whose teardown throws, true again next mission with no reinstall and the same function object, inert when injection did not run; a foreign hook stacked on top survives our unload and our gate answers beneath it once re-enabled; install refuses when the util is absent.
- `py build.py --deploy` built and deployed 2.5.0.128; the module is in the zip.
- Native checks owed at release, from the brief: tip a forbidden engine material, a permitted one and a soil-injected one; return to menu, load a second save, repeat, once career and once non-career; with a second mod hooking the same function loaded after Soil, return to menu and confirm its hook still runs. The engine restart lifetime remains not determinable from Lua.

### Bob's cold review, carried
APPROVE, deviation accepted as a build-shape choice inside RESTORES. Two notes to keep:
- After `reloadDlcsAndMods` re-sources the mod (reached from the main menu whenever mods changed), the fresh chunk's install flag is false, so the next mission installs a second wrapper on top of the stale one. Not a regression: the old file-local had the same property, and unload has already switched the stale wrapper to pure delegation, so it is an inert frame. This is the one footnote to "never chains".
- `GroundTipGate._resetForTests()` ships in production source as a test-only entry point, the only one under src/. Inert; named here so it is a decision, not an accident.
- The checklist item "enable sits inside the if-tmpl block and nowhere else" is source-proven at main.lua:494-498, not test-proven, since main.lua cannot load under the bench.


sasha-rf (approved):

PR #945, opened 2026-09-14
fix(burn): amendment burn accrues again and is metered on covered ground (RSF-F905, F347)

TheCodingDad-TisonK:
## RSF-F905 / F347: the amendment burn is switched back on, it burns the ground the boom actually covered, and every player is told the truth about it

Brief of record: `Office Tyson/RSF-F905-amendment-burn-covered-ground-IMPLEMENTATION-BRIEF.md` v0.5, Stage 6A VERIFIED, handoff certificate `F905-HANDOFF-V05-20260914` CERTIFIED. Player report #905 (antler22), register finding F347 (the dead gate). Lane: FAST TRACK (verified implementation of ruled intent, Arissani 2026-09-14).

### Tyson's ruling on the covered share (U1, accepted from Bob's reading)

A MEASURED covered-crop share DECIDES and replaces the single-point hasCrop / burnExempt verdict. The slice runs even at a share of 0 so the pass clock advances. The toast fires only when the tick actually charged something (share nil or above 0). A missing or failed coverage read keeps today's single-point verdict at the last spray position and is never treated as "nothing burnable".

### What changed

**Part A, the dead gate (F347).** `not field._amendBurnNotified` is out of the entry condition in `applyFertilizer`. The flag still throttles the toast to once per field per crop cycle and still clears at harvest and in `clearAmendmentBurn`. The metered slice now runs on every tick of a pass, which is what it was written for.

**Part B, covered ground decides (#905).** New `SoilFertilitySystem:computeCoveredBurnShare(fieldId, field, boomPoints)`: this tick's boom points are floored and deduplicated to cells exactly as `markBoomCells` does, kept only when the cell centre lies inside any parcel of the metered farmland's complete polygon union (`_getFarmlandPolygons`, the SF-52 helper, not the first-match `_getFieldPolyVerts`, which would reject every cell on the second parcel of a shared farmland and switch the burn off there for good), and each kept cell is read with the engine's own `FSDensityMapUtil.getFruitTypeIndexAtWorldPos` at that same centre (as GrowthBlock, GrowthCredit and ZoneYield already do). A cell is burnable when the unchanged `isAmendmentBurnRisk` says so. The share is burnable over sampled. `applyAmendmentBurnSlice` gains an optional `share` argument (nil = today's full increment, so existing two-arg callers are unchanged); a zero share advances the clock and never touches the record.

Missing versus empty: nil points, a polygon cached as unavailable, no engine per-cell get, or a failed read all return nil (fall back to the point verdict). A set that was obtained and holds no burnable cell, including one the polygon rejected entirely, is a real zero.

`HookManager` captures the boom points once, before the application, at both apply sites (primary `applySingle` and multi-tank `applyMulti`), only for a lime or organic profile, and the later `markBoomCells` sweeps reuse that capture instead of re-walking the nodes. The points travel as an argument into `onFertilizerApplied` and `applyFertilizer`; the share is cached per field keyed on the points table identity, so sibling sections on one field share one read and a section metering a different field uses that field's own polygon.

**Part C, the wire and the five surfaces.** `amendBurnPenalty` now rides every field delivery as a value plus an owner-written known flag:
- `NetworkEvents.lua`: all six stream functions (full sync read/write, batch sync write/read, field update read/write), each pair inserted immediately after that function's field-level `compaction` line. The four per-cell zone `compaction` sites are untouched.
- `SoilNetworkSyncBridge.lua`: the pair is written and read explicitly after the `SCALARS` loop and before the crop names. No `SCALARS` entry (the list stays at twenty). `deserializeFields` now returns `(out, ok)` and `_onReadState` refuses the whole apply when the walk does not land on `#arr + 1`.
- The writer always sends known = true with `field.amendBurnPenalty or 0`, so a cleared or never-limed field is an authoritative KNOWN zero. Not-known is a client-only state with one cause: the element was not in the payload.
- `getFieldInfo` publishes `amendBurnKnown = (g_server ~= nil) or (field.amendBurnKnown == true)`. On a host, single player included, that is always true, so the pre-emptive risk row behaves exactly as today.
- `SoilHUD` row counter, percentage row, risk row and `showBurnRisk` for the native field-info line; `RfPdaSoilMerge.aggregateInfo` (group known only when every member is known, penalty = worst known member, not area-weighted); `SoilTreatmentRates` "wait before amending" tip: all gate on known. Known non-zero shows the percentage, known zero shows the risk row when at risk, not-known shows neither. No string added or changed.

### Release note (carry into the changelog)

- The amendment burn is re-enabled. Liming or spreading organics over an established crop costs yield again, up to 80% for lime and the existing organic caps, in proportion to the standing crop the boom actually drove over. This is designed behaviour being restored, ruled 2026-09-14.
- **Server and all clients must update together.** Six native field streams grow by two values per field and the NetworkSync whole-map payload grows by two values per field, with no version tag. An old client on a new server misreads every field; a new client on an old server now refuses every whole-map apply (it keeps its last good copy and loses soil sync) instead of silently applying shifted values. Mixed versions were already unsafe for pH and compaction; this repair does not fix that and does not claim to (registered as owed in the brief).
- No save change. `amendBurnPenalty` is saved and loaded exactly as before; the known flag is never persisted.
- On a multi-parcel farmland every parcel counts (Bob's cold-review blocker, fixed before commit). A boom straddling two farmlands charges each its own within-farmland share for the full tick, which is what brief item 2 specifies.
- The toast fires only when a tick actually charged something (Tyson's U1 ruling above). The brief's item 1 kept the throttle "exactly as today"; this is the one player-visible rule added on top and it is named here for the 6B record.

### Verification

- `tools/test`: syntax OK, lint clean, 4148 assertions, 0 failed across 87 files. `burn_test.lua` untouched and green.
- New bench `RSF-F905-covered_share_and_payload_spec_test.lua` (165 assertions): Group A payload invariants on real source, Groups B to F the supplied reference contracts, Groups G to L on the real bodies (scaled slice, `computeCoveredBurnShare` missing/empty/mixed/dedupe/centre/cache/cross-field/failed-read, the `applyFertilizer` gate across ticks with the F347 trap, the bridge pair and whole refusal through `_onReadState`, the publisher under `g_server` nil and set, and one table of (known, penalty, risk) triples through the row counter, the field-info line, the merge and the treatment tip).
- `network_events_roundtrip_test.lua` sample field carries a penalty and asserts the pair arrives known on all three events (the typed FIFO proves each write/read pair drains exactly).
- FS25 APIs verified by Bob against the decompiled scripts: `FSDensityMapUtil.getFruitTypeIndexAtWorldPos` (utils/FSDensityMapUtil.lua:2849), `streamWriteBool` / `streamReadBool` (engine natives, already used in this file), `getWorldTranslation` (already under pcall in both existing helpers).
- Bob cold review of the diff: FAIL on one blocker (first-match polygon on a shared farmland), fixed with a new bench case H19/H20 that fails against the first-match contract. His should-fixes are in: the refusal warning is throttled to once per mismatch (the bridge path runs at 1 Hz), H15 now discriminates the two farmlands, the bench header claims four of five surfaces on real source, and both hook captures carry the same guard. Everything else verified clean: six sites as three matched pairs, four zone decoys untouched, no guessed API, no other caller of `deserializeFields` receives the second return in a position that matters, no scope drift.
- Residual MP notes from Bob, stated rather than fixed: `applyFertilizer` is not server-gated (pre-existing), so a guest driving the sprayer accrues a local penalty the next payload overwrites; and because SF's join full-sync is not delegated to NetworkSync, a refused bridge apply leaves a mismatched client on its join snapshot and loses only ongoing deltas.
- In-game acceptance (release lane, owed): lime an established crop across a part-harvested field and confirm the loss matches the standing share driven over; confirm the risk warning goes quiet once a burn is recorded on the HUD, the field-info panel and the PDA tip; on a dedicated server confirm a guest's screen shows the same burn the host recorded and no confirmed no-burn while a payload is unavailable.

### Unchanged, by design

Caps and clocks, `isAmendmentBurnRisk` verdicts, the harvest consume and the three clears (still nil), the save and ledger paths, `applyScorchEffect`, `markBoomCells` (still count-everything for the hectare meter), NetworkSync's API and registration, every player-facing string.


sasha-rf (approved):
﻿Reviewed the diff directly against the description: amendment burn increment is now scaled by covered-crop share per tick, nil share still accrues the full increment (today's behavior preserved), a measured zero keeps the pass clock running but adds nothing. Six native field streams each grow by two values, matching the release note that server and all clients must update together, no version tag, no save change. Bob's engineering pass already came back PASS; this checks out on my read too.

UTF-16 scan: 10 file(s) read at the PR head, all clean.

PR #946, opened 2026-09-14
fix(l10n): fungicide names stop apologising where the shown name is final, and the HUD status words and map shortfall go through the locale files (RSF-F345 wave 1, F346)

TheCodingDad-TisonK:
## RSF-F345 wave 1 / F346: fungicide names stop apologising where a real name is shown, and the HUD status words and map shortfall go through the locale files

Brief of record: `Office Tyson/RSF-F345-WAVE1-soil-translation-IMPLEMENTATION-BRIEF.md` v0.4 (Stage 6A VERIFIED, handoff certificate `F345-W1-HANDOFF-CERTIFICATE` CERTIFIED). Findings F345 and F346. Lane: FAST TRACK (Arissani's ruling 2026-09-14). SoilFertilizer half only; the FarmTablet half is its own PR (RSF-F345 wave 1 in FS25_FarmTablet). No runtime contract joins the two.

### Part A, the 23 fungicide display names

The governing rule from the brief: the `[EN] ` marker comes off a value only where that value is the form a reader of that locale would be shown as the finished name.

- `en`, `de`, `fr`: untouched.
- `ru`, `uk`: the brief's Cyrillic table, all 23 keys, marker off.
- 17 Latin-script files (`br cz da ea es fi hu id it nl no pl pt ro sv tr vi`): marker off on the 21 ISO common names, kept on `SULFUR` and `COPPER_HYDROXIDE`.
- `cs ct fc jp kr`: all 23 keep the marker.

Every `eh` attribute on every edited row is byte-identical to before (script-verified). 403 rows edited, `v=` only.

### Part B, five words at four render sites

- `SoilHUD.lua`: new `SoilHUD.statusText(token)` with `SoilHUD.STATUS_TEXT_KEYS`, called at the three render sites only (FIELD INFO Soil Grade value, nutrient row status, pH row status). The token stays the internal value everywhere: `statusColor`, `overallStatus`, the grade fold and every comparison still receive bare `Good` / `Fair` / `Poor` / `Unknown`. The helper uses `hasText` before `getText`, both under `pcall`, and rejects an empty result, a non-string and the `$l10n_<key>` placeholder, so a missing key renders exactly today's English token and never a key name. The existing `sf_fieldinfo_grade` label lookup on the same line is left as it was, per the brief.
- `SoilMapOverlay.lua`: the shortfall row formats through `tr("sf_map_gap_needed", "%d %s needed")`. The format runs under `pcall` with the English pattern as fallback, so a malformed translation cannot throw out of the tooltip draw. Sibling rows unchanged.
- Five new keys in all 27 locale files, unmarked English, inserted after `sf_dog_barn_warning`: `sf_hud_status_good`, `sf_hud_status_fair`, `sf_hud_status_poor`, `sf_hud_status_unknown`, `sf_map_gap_needed`. Hashes computed with the tool's own `get_hash`, as the F192 rows were.

### Locale tooling

`lang_sync.py check`, `report` and `sync` were run over the patched files in scratch. The five new keys report as untranslated in every non-English file by the exact-match branch (German 381 to 386). `sync` on a throwaway copy neither re-marks any edited `sf_chem_` row nor strips any new key. The pre-existing stale rows and the 41 missing Czech keys the tool also reports are not this wave. Bob's note for translators: because the five keys ship without the `[EN] ` prefix, any tooling that greps for the marker alone will not flag them; only the exact-match branch of `lang_sync.py` does.

### Verification

- `tools/test`: syntax OK, lint clean, 4034 assertions, 0 failed across 87 files. New bench `RSF-F345-wave1_status_text_test.lua` (57 assertions) drives the real `SoilHUD.statusText`: the four-key map, no i18n, translated values, unmarked English identical to pre-wave text, each key deleted in turn, the placeholder-from-getText shape that defeats an `or`, empty and non-string results, a non-boolean `hasText`, throwing lookups, an i18n without `hasText`, foreign tokens passing through, and `statusColor` still keying on the bare token.
- Brief checklist run as a script over the 27 files: en/de/fr chem rows byte-identical, ru/uk Cyrillic and unmarked, Latin 21 off and 2 on, non-Latin 23 on, five keys unmarked in all 27, every `eh` unchanged. No failures.
- The map tooltip shortfall site is inspection-verified rather than benched: `tr()` is file-local and the row sits inside the tooltip builder. The bench covers the three SoilHUD sites.
- Bob cold review: FarmTablet PASS; SoilFertilizer delta clean, one staging blocker (the scratch copy had been taken on the F905 branch) fixed by cutting the branch from `development` and applying only the F345 files, confirmed on the real diff. His two should-fixes are in: the bench's `statusColor` block asserts callability unconditionally instead of silencing itself, and the helper comment no longer claims the engine returns a `$l10n_` placeholder (the engine's `getText` returns a "Missing ... in l10n" string on a miss, `I18N.lua:175-210`; `hasText` is what closes the hole, the placeholder guard mirrors the file's `tr()` helpers). Bob verified every `sf_chem_` row, every `eh`, the five new rows and the sync behaviour independently, found no missed render site in either repo, and confirmed nothing here is multiplayer state.
- No save, network or internal value changes. No string in `en`, `de` or `fr` changes.

### Release note

- Fungicide names in the Soil PDA and treatment lists no longer carry the `[EN]` marker in 19 languages where the international name (or, for Russian and Ukrainian, a Cyrillic name) is what a farmer would be shown. Sulfur and copper hydroxide keep the marker outside English, German and French because they have real national names still owed. Chinese, Japanese and Korean keep the marker on all 23.
- The Soil HUD status words and the map tooltip's shortfall line can now be translated. They ship as unmarked English in every language until a translator picks them up, so no screen reads worse than today.

### Owed, carried from the brief

Native check on the Cyrillic table (two rows named: `METALAXYL_M` keeps a Latin M, `FOSETYL_AL` uses a space and a genitive). Native chemical names for the five non-Latin files. The map tooltip's `sf_map_target` and `sf_map_gap` labels are still marked in most locales and absent from Czech. Real translations of the five new keys. The `translation_fc.xml` Chinese-bodied file is F348. In-game acceptance on the release lane: a German HUD shows unmarked Good / Fair / Poor and never `[EN] Good`; a Polish fungicide list shows unmarked technical names beside a still-marked Sulfur.


sasha-rf (approved):
﻿Reviewed the diff against the description: the fungicide names drop the [EN] marker where the localized name is final, verified against the translation file diffs for the affected locales, Sulfur and Copper Hydroxide keep the marker, cs/ct/fc/jp/kr keep all names marked as described. The new HUD status keys and map shortfall line route through getText with a hasText gate and a "$l10n_" placeholder rejection, so a missing key falls back to the plain English token instead of leaking a key name. Bob's pass came back PASS; this checks out on my read too.

UTF-16 scan: 30 file(s) read at the PR head, all clean.

PR #947, opened 2026-09-14
fix(dog): the barn warning fires through a SoilFertilizer-owned reader that decides per disease record (RSF-F190)

TheCodingDad-TisonK:
## RSF-F190: the dog's barn warning finally fires, through a SoilFertilizer-owned reader that decides per disease record

Instruction of record: `Office Tyson/RSF-F190-V08-CERTIFIED-HANDOFF.md` (effective v0.8: Iris v0.1 baseline plus ClaudeA amendment), Stage 6A `RSF-F190-V08-STAGE6A-VERIFICATION.md` VERIFIED, fresh Hazel handoff certificate `RSF-F190-V08-HANDOFF-CERTIFICATE-2026-09-14.md` CERTIFIED, delivered by Iris 2026-09-14. Lane: repair (RESTORES, nothing adds intent). Build only; in-game, multiplayer, dedicated and save observations are owed to release per the certificate.

### The farm sentence

Owning a doghouse earns a farmer a passive signal that an owned field or barn is wrong, nothing more. The crop half has always worked. The barn half has never fired once: it called `g_diseaseManager:getDiseaseTitles()` and read `d.active`, neither of which exists on any version of the livestock provider. The barn sentence has been shipped text since CD-13. Nothing here adds a string; it makes an existing one reachable.

### What changed

- **New `src/LivestockWarningReader.lua`** (handoff items 1, 2). One SoilFertilizer-owned reader, sited in the source root beside the other outside-facing bridge, not a method of `SoilFertilitySystem` and not inside `NpcSoilBridge`. It publishes nothing on the mission. `isBarnActivelySick(placeable)` returns `true` when any animal in the barn holds a record that is active sickness, `nil` for everything else. Never a disease name or a count.
- **Gate, then verdict** (items 4 to 7). The provider's own `animal:getHasAnyDisease()` is called under `pcall` and must return boolean `true`, which is how this mod learns provider presence and the diseases-enabled state without reaching across the sandbox for `g_diseaseManager`. That gate never decides: on the installed 1.2.6.0 it is true for any attached record, cured animals and carriers included. The verdict walks `animal.diseases` and counts a record only when `cured` and `isCarrier` are both real booleans reading `false`. A carrier record beside an active one still warns. A malformed record never counts (the opposite of DairyCore's herd-score bridge, on purpose: the dog's law is no false positives). Any throw, missing list, healthy herd or refused gate is `nil`.
- **Containment** (item 8). Each animal is read under its own `pcall` in the reader, and the dog wraps each barn in its own `pcall`, so one bad animal cannot silence its barn and one bad barn cannot silence the farm. DairyCore's session-wide `safeRead` degrade is deliberately not copied.
- **`DogEarlyWarning.lua` barn branch** (item 3). Replaced the dead manager check, the mod-loaded check and the title scan with: read the placeable list, then per barn, same-farm filter, ask the reader, test presence. Same shape as the crop line. Unchanged: crop branch, doghouse gate, `"livestock"` dedupe type and key, dedupe clearing, one-minute cadence, HUD delivery, the F192 locale keys and fallbacks, the `"barn"` id fallback.
- **`main.lua`**: one `source` line for the reader immediately before `DogEarlyWarning.lua`.

### Verification

- `tools/test`: syntax OK, lint clean, full suite green. New bench `RSF-F190-livestock_warning_reader_test.lua` (89 assertions) drives the real reader and the real `DogEarlyWarning:scan`: every 6A section 10 discriminator (active, cured, carrier, mixed, healthy, provider absent, diseases disabled, malformed, thrown) against both provider gate shapes (1.2.6.0 any-record and 1.3.2.1 strict), the gate refused or non-boolean, per-animal and per-barn containment, the accepted limit where the only sick animal throws, same-farm filter, doghouse gate, dedupe once while sick and clearing and re-warn, the crop half still firing beside a healthy barn, reader absent leaving the crop half alone, and every string and the cadence unchanged.
- Provider facts re-read from the installed `FS25_RealisticLivestockRM` 1.2.6.0 bytes: `getHasAnyDisease` at `RealisticLivestock_Animal.lua:1591-1593`, `self.diseases` at `:306`, `Disease.new` flags at `Disease.lua:10,15`, dense `getAnimals` at `RealisticLivestock_AnimalClusterSystem.lua:43-47`. Engine: `spec.clusterSystem` at `PlaceableHusbandryAnimals.lua:161`, `Placeable:getUniqueId` at `Placeable.lua:814`.
- No save, network or player-facing string change. No provider write.

### Accepted limits, from the handoff

A throw on the only sick animal in a barn leaves that barn quiet for that scan. A barn whose list cannot be walked is quiet. A transient unavailable read then recovery blinks the same warning twice, at a sixty-second cadence. The dog gains no failure bark (Arissani's ruling 2026-09-14). The dog's delivery fence (a local HUD call, no server-only guard, cannot reach remote players) is existing behaviour outside this repair.

### Owed to release

Watched initial join and rejoin ordering against the first scan; active, cured, carrier, mixed, healthy, absent, disabled, malformed and thrown results in game against both provider lanes; native local and remote HUD delivery; dedicated server; save round trip. No performance figure is claimed.


sasha-rf (approved):
﻿Reviewed the diff against the description: the barn warning now goes through a new SoilFertilizer-owned LivestockWarningReader instead of reaching into the provider's disease titles directly. Each barn and each animal is checked under its own pcall, so one bad record can't silence the whole farm. No new player-facing string, no save or network change, matching the release note. Bob's pass came back PASS; this checks out on my read too.

Issue #677, opened 2026-06-21
admin settings panel hot key issue


Arissani:
### Mod Version

newest

### Farming Simulator 25 Version

1.8

### Game Mode

Singleplayer

### What happened?

I am encountering a persistent issue where the custom hotkey assigned to open the admin settings panel fails to save across game sessions. Regardless of successfully mapping the key in the in-game "Controls" menu, the binding resets or reverts upon restarting Farming Simulator 25. This requires manual remapping at the start of every play session to access the admin functionality. it saves the key as the control but i need to remap it to use it every new session even though in controls it always states the right  key i set  for example i use right alt + M everytime i load a save in contros it still says right alt + M but it wont work until i remap it to right alt + m again 

EDIT - this just needs further investigation as now it seems like at times it saves but other times it doesn't 

### Steps to Reproduce

Navigate to the game Settings > Controls > Mods menu.

Locate the Admin Settings Panel action for the mod.

Assign a custom key combination (e.g., Left Ctrl + Shift + O).

Apply and Save changes.

Close the game completely and relaunch.

Observe that the keybinding has reverted to the default or is unassigned.

### Log Snippet

```text
n/a
```

### Other Mods Loaded

n/a

TheCodingDad-TisonK:
Fixed

PR #949, opened 2026-09-15
fix(input): RSF-F201 context-qualified input registration

TheCodingDad-TisonK:
## RSF-F201: keep cab and on-foot controls valid without repeatedly rebuilding them

**Cohort PR, one of eight.** IncomeMod, TaxMod, SettingsHub, MasterHUD, FuelCosts, WorkerCosts, RandomWorldEvents, SoilFertilizer. F201 item 8 forbids mixed old and new live-code injection across the suite, so the eight merge and deploy together; release is held until all eight pass in TESTING.md.

### Root cause (engine, bound by symbol)
The engine keys an action event by action name, target object and trigger shape only (`InputEvent:makeId`, InputEvent.lua:56-61). Registering the same action with the same target in PLAYER and VEHICLE therefore produced one identifier and one global slot; the VEHICLE context is recreated on foot-to-cab entry (`createContext` calls `deleteContext`, InputBinding.lua:2636-2655), and that wipe took the PLAYER handle with it. The key still fired, but every identifier-keyed setter (:883-938) went dark. On top of that, 17 result captures across four mods were shadowed inside the MasterHUD key-ownership gate, so handles that did register were dropped.

### What this PR does in FS25_SoilFertilizer
All PLAYER and VEHICLE Soil actions register through private per-context forwarding targets (new src/utils/SoilContextInput.lua) with the hook record on the SoilFertilityManager class table. The vehicle hook's PLAYER purge and PLAYER rebuild are gone (item 13). registerPlayerContextInputEvents is now the complete PLAYER reconciliation behind the unchanged two-second one-shot (#677). Admission reset is the first act of update(), before that one-shot. delete() retires the owner without restoring the predecessors; the existing targeted removeActionEvent cleanup stays. Cab F1-strip presentation (visibility + GS_PRIO_HIGH) unchanged.

### Shape (same in every primary participant)
- Private forwarding target per context; the target forwards to the live owner's handler with the engine's argument list untouched.
- Membership asked of the wrap's own context (`binding.contexts[name].actionEvents[action]`), matching action, owned target, callback and the three trigger flags. A stored id is never treated as presence; a lost id is recovered from the live event.
- One batch per participant and context; no delta means no begin/end.
- Captured predecessor always called; the in-flight flag covers only this participant's own injection; close attempted on throw, error propagated.
- Session-lived wrapper record on a persistent home table; mission teardown retires the owner and never restores the captured method.
- Per-update-interval attempt memo instead of a timer.

### Verification
- Lua 5.1 parse of every changed file; pre-commit hooks green.
- SettingsHub: the certified F201 bar (221 assertions) re-pointed at production, PASS. Composition smoke (two stacked real helper copies on the model binding: balanced brackets, no double close, foreign events untouched, retire restores nothing, throw propagates with protection released), PASS.
- Zip built, not deployed. Deploy as one cohort.

### Manual test
Row in ecosystem-dev-tracking/TESTING.md. Native observations owed: PLAYER and VEHICLE entry/exit, vehicle-to-vehicle, animal-riding driver and passenger, controls-menu remap and gamepad replug, repeated mission load, hot reload, MasterHUD present and absent, disconnect/rejoin, dedicated server, help-strip visibility, steady-state frame cost.

### Design
ecosystem-dev-tracking: Office Tyson/RSF-F201-V14-EFFECTIVE-DESIGN-2026-09-14.md (certified v1.4, Hazel Stage 6B). Intake: Bob, 2026-09-15, FAST TRACK.


sasha-rf (approved):
﻿Sasha review: RSF-F201 cohort (8/8), FS25_SoilFertilizer.

Private context-input module (SoilContextInput.lua) and F201 test present, pointed at production. CHANGELOG updated, and also removes the old purge-and-rebuild-on-cab-close workaround. TESTING.md row in place in ecosystem-dev-tracking. Bob's engineering review already passed with MAJORs addressed (including the immediate catch-up MAJOR).

PASS. This PR merges and deploys together with the other seven F201 PRs only; do not land it alone.

UTF-16 scan: 6 file(s) read at the PR head, all clean.

PR #950, opened 2026-09-15
fix(RSF-F202): bounded circular debug history and a truthful export line

TheCodingDad-TisonK:
## RSF-F202: bounded circular debug history and a truthful export line

Base: `development` at 51fe3c5e. One item, one PR. Built to the certified effective handoff (original v1.0 plus amendment v0.7), Bob intake FAST TRACK 2026-09-15.

### What changed
- `src/utils/Logger.lua` only, plus the changelog lines the amendment names.
- The debug history is a bounded circular store: two file-local values (`historyOldest`, `historyCount`) initialized beside the fresh buffer at module execution (re-source resets both). An accepted message writes one slot and overwrites the oldest once full. No `table.remove`, no shift, copy or scan on append. Capacity stays 500.
- `flushDebugLog` walks the ring oldest to newest and writes the same zero-based keys and count. Early returns, propagation and the clear-only-after-normal-return boundary are unchanged (a nonthrowing false save still clears, as before).
- Amendment item 10: one info line, `Debug log written: <base>/Debug/debug.xml (<n> entries)`, printed after the reset only when `xml:save()` returned boolean true. False, nil or anything else announces nothing. Info level, so the SoilDebug off-edge prints it with debug already off; it does not enter the history.
- Item 12: the flush comment now names the two real callers (SoilDebug off-edge, session teardown) and says the other debug-off routes save without exporting.
- Item 11: CHANGELOG's debug output header and `debug.xml` bullet replaced with the handoff's exact text; the `field_dump.xml` and `soil_export.xml` lines are untouched.
- Item 9 rebinds honoured: no flush call site moved (`SoilFertilityManager:delete` and the SoilDebug command are untouched); no new trigger, setting, command or retention change.

### Verification
- `tools/test/run-tests.mjs`: RSF-F202 186/0. The bar is re-pointed to the production ring as the handoff requires: GROUP A asserts the repaired overflow (zero head removals, one observed storage write per accepted message through a proxy table, zero history reads on append, exact newest-500 export), the real-source groups seed history through the public `SoilLogger.debug` path and drain through the real flush boundary, an actual re-source of the chunk shows only fresh records, and GROUP J covers the item 10 announcement (true only, after reset, ungated, once, count equals XML count; false/nil/number/string silent with the original clear boundary; early returns silent; delete throw after a true save silent with history retained). GROUPS E-I keep the delivered reference ring untouched. Full suite green.
- `check-staged.mjs` green. No em dashes in added lines.
- The harness needed `npm install` in `tools/test` (node_modules was empty locally); nothing committed from that.

### Boundaries carried from the handoff
Native file bytes, durability, partial or stale files, ordinary versus abrupt teardown, per-process client/host/dedicated profile behaviour and frame cost are Wizard and release observations, listed in the TESTING.md row.

### Governing docs
`Office Tyson/RSF-F202-bounded-debug-history-IMPLEMENTATION-BRIEF.md` (v0.7-effective), `RSF-F202-V07-HANDOFF-CERTIFICATE-2026-09-14.md`, bar `Office Tyson/mods/FS25_SoilFertilizer/RSF-F202-debug_history_spec_test.lua`.


sasha-rf (approved):

Reviewed cold: debug history ring buffer genuinely bounded (500 entries), append/flush math hand-verified correct including wraparound, on-disk format unchanged. Reset only happens after xml:save/delete both succeed, so a mid-flush exception preserves history. New export line strictly gated on xml:save()==true and reflects the actual flushed count. CHANGELOG matches code. No em dashes, clean scope, 753-line test file.

TheCodingDad-TisonK:
**Engineering review (Bob): PASS** at 788e7546

Cold review against the RSF-F202 effective handoff (original v1.0 plus amendment v0.7).

- **Ring:** before the store fills, a new message takes slot count+1; once full, it overwrites the oldest slot and advances. No `table.remove`, no scan on append.
- **Flush:** walks oldest to newest with the same zero-based keys, then resets storage and both counters after save and delete.
- **Export line:** printed only when the save returns boolean true. `XMLFile:save` returns true on success (engine `xml/XMLFile.lua:85-90`), so the gate is reachable.
- **Callers:** exactly the SoilDebug off-edge (`src/settings/SoilSettingsGUI.lua:1318`) and teardown (`src/SoilFertilityManager.lua:2014`).
- **CHANGELOG:** matches item 11's exact text.
- **Bench:** 4490/0 on the head (+186 F202 over base). Syntax green, numstat equals `--ignore-cr-at-eol`, no em dashes.
- **TESTING.md row:** present with concrete checks.

Ready for PR review. Merge is Tyson's.

PR #951, opened 2026-09-15
feat(SG-6): Soil capacity participation and joined ground-tip lifetime

TheCodingDad-TisonK:
## SG-6 (Soil half): capacity participation and joined ground-tip lifetime

Base: `development` at 51fe3c5e (F227 merged in #943, preserved). One item, one PR. Built to `Office Tyson/StockGuard-First-Family-2026-09-15/implementation/SG-6-SOIL-CAPACITY-INTEGRATION-BRIEF.md` (v1.2) on Bob's step 1 intake (ledger 0fbb8a8). Paired with the StockGuard half (feat/SG-6-native-capacity, separate PR); the two develop independently against this fixed contract and are joined for preflight, ground initialization and failure acceptance. Merge gate C1 is Operator's. Built in a fresh worktree off origin/development per Bob's C4.

### What changed
- New `src/hooks/SoilCapacityIntegration.lua`: module-level table with `protocolVersion = 2` and the three dot-call functions. `beginCapacityLoad` binds the mission and calls `GroundTipGate.disable()`; `endCapacityLoad` clears the matching binding and leaves the override disabled; `isJoined(mission)` is what main.lua's fence asks. `prepareGroundTypes` validates all twelve names, templates and slots first (reasons `MISSING_NATIVE_FILL`, `MISSING_TEMPLATE`, `GROUND_CAPACITY`, `INVALID_ARGS`, each with the offending name), then appends the missing rows in the fixed order UREA, AN, AMS, MAP, DAP, POTASH, POLIFOSKA, GYPSUM, COMPOST, BIOSOLIDS, CHICKEN_MANURE, PELLETIZED_MANURE with the same shallow-copied template rows and field values the legacy fallback uses (`allowsSmoothing=false`, `canBeTipped=true`, own index and name). No sort of the combined roster.
- `src/main.lua`: sources the module after `GroundTipGate.lua`; the legacy ground block in `loadedMission` is now `if SoilCapacityIntegration.isJoined(mission) then <skip and log> else <unchanged block> end`; `unload` calls `endCapacityLoad(g_currentMission)` right after `GroundTipGate.disable()`.
- `CHANGELOG.md` entry. Soil's standalone path, the F227 wrapper (captured predecessor once, current manager, runtime flag, never chained, never removed) and every other subsystem are unchanged.

### Engine seams (D:\FS25_Decoded\dataS\scripts_decompiled)
Mod environment published as `_G[modName]` with `getfenv` redirect (mods.lua:429-448); `DensityMapHeightManager` tables `heightTypes`, `fillTypeNameToHeightType`, `fillTypeIndexToHeightType`, `heightTypeIndexToFillTypeIndex`, `numHeightTypes` (initDataStructures :41-49) and the per-row fields set by `loadDensityMapHeightTypeFromXML` (:197-222) with `sortHeightTypes` (:187-195); `DensityMapHeightUtil.getCanTipToGround` (densityMaps/DensityMapHeightUtil.lua:26-37); `FillTypeManager:constructTerrainFillLayers` (fillTypes/FillTypeManager.lua:232).

### Bounded readings
- "A different still-active mission is rejected": a mission counts as still active until `endCapacityLoad` (or Soil's unload) clears it, so the reject is on an uncleared binding; a nil `endCapacityLoad(nil)` clears any binding for teardown paths that no longer hold the mission.
- The fence keys on the mission table passed to `loadedMission`, the same object StockGuard's preflight passes to `beginCapacityLoad`.

### Verification
- `tools/test/run-tests.mjs`: SG-6 Soil 75/0 (protocol shape including the dot-versus-colon behaviour the family bar checks, joined-mode fence with the real GroundTipGate, begin/end lifetime and idempotence, preparation order with existing rows preserved, both templates, map consistency, idempotent repeat, missing fill / exhausted limit / missing templates with no partial insertion, single-template fallback, invalid arguments, and the bar's R3 discriminators: joined before prepare, failed prepare keeps the fence, standalone-then-joined never returns a blanket true, invalid manager follows the predecessor). RSF-F227 39/0 unchanged. Full SoilFertilizer suite green. Syntax gate green. No em dashes.
- Not proven here: native initialization with the prepared rows, actual tipping and appearance, terrain updater association, saved ground identity across the append, hosted/dedicated join, first failure and second load. Those are the TESTING.md observations and F227's owed native checks ride on this row.


TheCodingDad-TisonK:
**Engineering review (Bob): PASS** at f30391d9 (re-check after fixes)

The first review at 3946c7e7 was PASS with MAJORs. All of them are fixed in f30391d9:

- **Colon calls:** `beginCapacityLoad` and `endCapacityLoad` now refuse the API table as the mission.
- **Templates:** both FERTILIZER and MANURE are required before any write. `MISSING_TEMPLATE` names the absent one, and there is no cross-template fallback.
- **TESTING.md row:** now expects one additional inert `getCanTipToGround wrapped` line after a main-menu mods-set change (the engine re-sources mods, `mods.lua:1109-1122`).
- **End binding:** `endCapacityLoad` clears only the matching binding. Unload passes the mission being deleted (prepend at `src/main.lua:900`).
- **Guard:** all four height maps are checked. A drifted count whose next slot is occupied is refused.

**Contract and bench checks:**
- Every missing name, template and slot is validated before any write.
- Rows are appended at `numHeightTypes+1` in the fixed twelve-name order, with no sort.
- The legacy ground block is fenced in every joined phase.
- F227 `GroundTipGate` is untouched.
- **Bench:** SG-6 Soil 100/0, F227 39/0, suite 4425/0. Syntax green, no em dashes.

Ready for PR review. Merge is Tyson's. Should merge before the Soil ground-condition work (F208).

sasha-rf (approved):

Re-reviewed cold at fix head f30391d9 after Bob's majors. All three majors and both minors verified as genuine structural fixes, not symptom suppression: beginCapacityLoad/endCapacityLoad now refuse a colon-call binding the API table as a mission; both fertilizer and manure templates are checked independently before any write with the correct offending name returned on MISSING_TEMPLATE; endCapacityLoad(nil) no longer matches an active binding so a nil unload can't clear someone else's join; main.lua's unload hook now takes the mission from the FSBaseMission.delete prepend argument instead of assuming g_currentMission; the INVALID_ARGS guard covers the two maps that were missing and refuses on drift into an already-occupied height-type slot. 294 new test lines target each fix by name including a two-saves-in-one-process scenario. No em dashes. Scope stays to the SG-6 Soil half only.

PR #940, opened 2026-09-14
fix(l10n): German soil unit label says ppm, not kg/ha (RSF-F218)

TheCodingDad-TisonK:
## RSF-F218: the German soil display labels a ppm reading as kg/ha

Brief: ecosystem-dev-tracking, Office Tyson/RSF-F218-german-ppm-unit-IMPLEMENTATION-BRIEF.md (v0.2), certified in RSF-SOIL-WAVE-CD-HANDOFF-CERTIFICATE-2026-09-13.

### What changed
- `translations/translation_de.xml` line 563: `sf_hud_unit_ppm` goes from `kg/ha` to `(ppm)`. German agronomy uses "ppm" for parts per million, so the form matches the English label.

### Unit check across all 27 locales (brief item 2)
| Value | Locales |
|---|---|
| `(ppm)` | 24 (English plus 23 unchanged) |
| `(млн⁻¹)` | Russian, correct, untouched |
| `(百万分比)` | Chinese Simplified, correct, untouched |
| `kg/ha` | German, fixed here |

### Not changed
The key name, the English value, `PPM_DISPLAY`, both readers (SoilHUD.lua:1678, SoilMapOverlay.lua:1332), and anything in FarmTablet.

### Verification
- `node tools/test/run-tests.mjs`: 3773 passed, 0 failed across 83 files.
- `py lang_sync.py check`: German shows 0 missing before and after. Its "untranslated" count rises by one because the tool counts a value identical to English as untranslated; that is the intended value here. The pre-existing missing-key report for other locales is unrelated to this change.
- `py build.py --deploy` built and deployed 2.5.0.128. In-game German HUD read owed at release.

PR #952, opened 2026-09-15
fix(RSF-F219): honest pH grain on the handful, map tooltip and sprayer cab

TheCodingDad-TisonK:
## RSF-F219: honest pH grain on the handful, map tooltip and sprayer cab

Base: `development` at 51fe3c5e (F217 kit #942 and F218 ppm #940 already in). One item, one PR. Built to `Office Tyson/RSF-F219-handful-ph-grain-IMPLEMENTATION-BRIEF.md` v1.3 (Hazel HANDOFF certificate HAZEL-F219-V13-C1, Stage 6A v1.0), Bob's FAST TRACK intake. Brief version on tracking origin/main re-checked at build time: v1.3, unchanged.

### What the farmer gets
The handful's pH word says whether the actual sample answered or only a recorded field value remains. Kit formatting is unchanged (band without the kit, band plus one decimal with it). A missing current pH never uses a field report or default as the spot. The map click avoids the nil formatter. The cab pH row shows the host dash, dim, over an empty bar instead of 0.0, poor, or a stale current measurement. All 27 locales carry the three new words.

### The seven items
1. `src/SoilFertilitySystem.lua` getFieldInfo: positional miss assigned outside the `_ensurePHReport` guard (nil pH, UNAVAILABLE, nil grain); coordinate 0 present; non-positional path and its number predicate unchanged; `_phReportRead` never on the positional path. The missing-pixel comment now says where the request goes.
2. On that miss a CURRENT report with a numeric value refreshes `pHLastKnown` only.
3. `src/HandfulRead.lua`: `pHStatus`, `pHGrainMetres`, `pHLastKnown` ride the frozen payload beside `pH`; `fromZoneCell` kept for N/P/K/OM; no new network payload.
4. `src/ui/SoilHandfulDialog.lua` + `xml/gui/SoilHandfulDialog.xml`: new `hfPhGrain` Text under the pH tile (profile `sfHandful_tileGrain`, same placement as `hfDisGrain`), registered in ELEMENT_IDS. `phTile(p, kit)` decides value, severity and word per status; kit decision unchanged (strict `testKitActive == true`). The shared header word now describes N/P/K/OM and its comment says so.
5. `src/ui/SoilMapOverlay.lua` drawCellTooltip pH branch: numeric row guarded at the format call; No data condition row kept; nothing else in the overlay touched (non-positional getLayerColor paths unchanged).
6. `src/ui/SoilSprayerInfoPanel.lua`: update loop clears the smoothed pH slot when the current `info.pH` is not numeric; draw loop special-cases the pH row (no `or 0` coercion, fill 0, C_DIM colour, "--"), other rows unchanged.
7. `translations/translation_*.xml` (27): `sf_handful_ph_unavailable`, `sf_handful_ph_last_known`, `sf_handful_ph_approx`, translated per locale (cs Simplified Chinese and cz Czech separately). Wording chosen by Implementation within the brief's meaning; speaker pass welcome.

### Bounded readings
- LOCAL word is "spot (2.0 m)" style: the existing `sf_handful_grain_spot` word plus the carrier grain with a plain metre symbol; no new unit key (F218 owns `sf_hud_unit` keys, untouched).
- A recorded-field band keeps its band severity colour, under its own label, as the brief allows.
- Any status the tile does not know (STALE, unknown, absent) dashes with an empty word.
- The cab's native progress-bar path (SoilHUD renderer) receives fill 0 and the dim colour for the unknown row.

### Verification
- `tools/test/run-tests.mjs`: new `RSF-F219-handful_ph_grain_spec_test.lua` 118/0 driving the real producer, HandfulRead.assemble, dialog `_populate`, `onMapClick` + `drawCellTooltip`, and the cab `update`/`draw` loops against engine stubs. Cases: helper present/absent, both/one/zero coordinates, coordinate 0, LOCAL hit, APPROXIMATE zone cell, report CURRENT/EMPTY, needsFertilization known/unknown, forwarding, kit off/on for LOCAL/APPROXIMATE/recorded, unavailable dash, stray current value under UNAVAILABLE, STALE/unknown/absent, header independence, element id binding, direct pH click, other-layer click then settings-panel switch, setLayer dismissal, known pH numeric row, cab known/unknown/known recovery, other nutrients keep smoothing, table pH, no field.
- Full suite 4422 passed, 0 failed (the worktree layout cannot load `../FS25_MarketDynamics` for om_213, same on base).
- Lint clean (97 files), syntax gate green, no em dashes in the diff (the six in translation_cs.xml predate this PR).
- Built and deployed with `py build.py --deploy`.
- Not proven here: native dialog fit of the word under the pH tile in every locale, map/cab rendering, listen host/dedicated/client join and resync, save/load. Those are the TESTING.md observations.


sasha-rf (approved):

Reviewed cold: phGrain/phLastKnown are pre-declared locals earlier in getFieldInfo, so the new positional-miss branch's reassignments are safe. The UNAVAILABLE-on-positional-miss path correctly leaves ph nil without touching the existing needsFertilization logic. Every consuming UI surface is guarded at the point of use rather than coercing to a fake number: handful tile dashes, map tooltip skips the numeric row, sprayer cab clears the smoothed value and draws a dim dash over an empty bar. 27 locales carry the three new keys. No em dashes. 118 new test cases, full suite 4422/0.

UTF-16 scan: 36 file(s) read at the PR head, all clean.

TheCodingDad-TisonK:
**Engineering review (Bob): PASS** at d3e6ffbf

Cold review against RSF-F219 brief v1.3, items 1 to 7. No findings.

1. **Positional miss:** `getFieldInfo` tests both coordinates before the report helper, then sets `ph=nil`, `UNAVAILABLE` and `phGrain=nil`. All pH variables are locals, and `_phReportRead` is not called on the positional path.
2. **Last-known value:** a CURRENT report refreshes `pHLastKnown` only.
3. **Payload:** `HandfulRead` forwards `pHStatus`, `pHGrainMetres` and `pHLastKnown`; `fromZoneCell` is kept.
4. **Handful tile:** new registered `hfPhGrain` element with per-status wording. Unknown statuses dash, and the kit rule is unchanged.
5. **Map tooltip:** the numeric pH row is guarded at the format call, including after a settings-panel layer switch.
6. **Sprayer cab:** the pH smoothing slot is cleared; the row draws a dim `--` over an empty bar with no `or 0` coercion. N/P/K/OM are unchanged.
7. **Locales:** all 27 carry the three keys in translated text.

**Bench:** 4443/0 on the head with the MarketDynamics sibling present. Syntax green, numstat equals `--ignore-cr-at-eol`, no em dashes. The TESTING.md row has concrete in-game checks.

Ready for PR review. Merge is Tyson's.

PR #953, opened 2026-09-15
chore(CD-15): native cell probe, read-only logging for the F35/F210 cost call

TheCodingDad-TisonK:
## chore(CD-15): native cell probe, read-only logging for the F35/F210 cost call

Base: `development` at 56e102c6. One item, one PR. Assigned by Tyson through Operator (Bob's prep ledger c662b2d, assignment e6f2db5). Chore: no gameplay effect, no saved state, no change to disease, dog or treatment behaviour; removable without trace (one source line in `src/main.lua` and the file `src/probe/CD15NativeCellProbe.lua`).

### What it measures
Behind the console command `sfCd15Probe` (off until asked):
- `planes`: every fruit's density plane id, map size, pixel size, state channels, cut and withered states, and whether a plane is shared by more than one fruit; Soil's cell size (terrainSize / resolution, the 1024..4096 clamp at `src/maps/SoilValueMaps.lua:255`).
- `here`, `at x z`, `cell gx gz`, `area x1 z1 x2 z2`: for each Soil cell, a DensityMapMultiModifier count over the cell parallelogram (one EQUAL filter per raw state plus GREATER 0, FieldGetInfoTask's shape) exact and with a 0.1-pixel inset, against a full per-pixel enumeration of the cell through `getDensityTypeIndexAtWorldPos` / `getDensityStatesAtWorldPos` (their raw values are logged; no decompiled script calls them), classified living / cut / withered / empty / other, with a type-index histogram for shared-plane isolation, and the microseconds per execute.
- `hooks on|off`: wraps `Cutter.processCutterArea`, `Mower.processMowerArea` and `SowingMachine.processSowingMachineArea` on all four surfaces (class, registered vehicle types, live instances, live work-area stored pointers, since WorkArea.lua:266 copies the pointer at load and :182-183 calls only the copy). While on, each call logs the touched cells before the native call, delegates once with unchanged arguments, and reads again in the same frame; the per-frame probe cost is tracked. `hooks off` restores every slot.
- `bench n`: n executes on the origin cell for every fruit, mean and max in microseconds, tagged singleplayer/listen or dedicated.
- `report`, `reset`, `inset f`.

### Measured results
None yet. This PR ships the instrument; the numbers are collected by running it on a 2 km map and on a 16x map, in singleplayer and on a dedicated server, and are pasted here by whoever runs it (the TESTING row lists the runs). Bob asked for the numbers in this body; they will be added as a comment and copied into the body when the runs are done.

### Engine seams (D:\FS25_Decoded\dataS\scripts_decompiled)
HarvestMission:createModifier missions/field/HarvestMission.lua:187-199, execute :288-290; FieldGetInfoTask field/FieldGetInfoTask.lua:20-31, :72; DebugDensityMap debug/elements/DebugDensityMap.lua:52; FSDensityMapUtil utils/FSDensityMapUtil.lua:3-8 (INCLUSIVE rounding on every modifier) and :232; FruitTypeDesc fruits/FruitTypeDesc.lua:127-128, :142-143, :233, :240, :601; WorkArea vehicles/specializations/WorkArea.lua:266 and :182-183; Cutter.lua:584, Mower.lua:328, SowingMachine.lua:362; getTimeSec (field/course/ai/AIFieldCourse.lua:184); Soil's own four-surface wrap pattern src/hooks/HookManager.lua:2726-2800.

### Bounded readings
- Shared-plane isolation is not claimed through a filter: the probe records whether a plane is shared and logs the per-pixel type-index histogram so the isolation question is answered by observation.
- "Living" is every non-zero state that is neither a cut state nor the withered state; a state above the plane's channel range is "other" (a modded plane).
- The point enumeration steps one fruit pixel, sampling pixel centres inside the cell bounds; a cell that is not a whole number of pixels wide shows that in the counts.

### Verification
- `tools/test/run-tests.mjs`: new `CD-15-native_cell_probe_spec_test.lua` 61/0 (cell geometry with the reference bar's half-open rule, inset, classification, comparison rows, timing summary, cells under a quad, the engine parts through recording stubs, the four-surface hooks: pure delegate while off, single delegation with unchanged arguments while on, every slot restored by off). Suite green apart from the worktree's known om_213 sibling-load.
- Lint clean, syntax gate green, no em dashes.
- Built and deployed with `py build.py --deploy` so the runs can start.

### Review fixes
Bob's cold review (3 MAJORs, 3 MINORs) arrived after the merge; the fixes are PR #954 (commit 7d7d65a8). Run the measurements on #954 or later.


sasha-rf (approved):

Reviewed cold: genuinely inert until invoked via the console command. makeWrapper is a pure passthrough when P.state.enabled is false (doesn't even build a snapshot in that branch), and when enabled it delegates exactly once with unchanged arguments and passes all return values through untouched. All snapshot/logging work is wrapped in pcall so a probe-internal failure can never propagate into gameplay code. hooksOff restores every wrapped slot (class, registered vehicle types, live instances, live work-area pointers) to the exact original function reference, verified by test F11-F14. Every engine call is a read (DensityMapModifier/Filter/MultiModifier execute, getDensityTypeIndexAtWorldPos/getDensityStatesAtWorldPos) - no write path anywhere, no save state touched. Cell geometry correctly uses the half-open rule (negative edge admitted, positive edge refused, off-map refuses rather than clamps). No em dashes. Thorough test coverage including the full four-surface hook lifecycle (off passthrough, on single-delegation, restore).

TheCodingDad-TisonK:
**Engineering review (Bob): PASS with MAJORs** at 418e01ae

Cold review of the CD-15 native cell probe. It has no gameplay risk: nothing runs until `sfCd15Probe` is typed, and every wrapper is a pure delegate while the probe is off. The MAJORs below would make the measurements wrong, so fix them before the 2 km, 16x and dedicated runs.

**MAJOR**
1. **Ground truth reads the raw state value** (`countCellPoints`). The probe classifies the raw `getDensityStatesAtWorldPos` value. The engine decodes that value with `desc:getGrowthStateByDensityState(states)` (`utils/FSDensityMapUtil.lua:2854`), a shift by `startStateChannel` and a mask to `numStateChannels` bits (`fruits/FruitTypeDesc.lua`). For a fruit with a channel offset, the raw value is not the growth state. Fix: classify the decoded state and log the raw value beside it.
2. **Shared planes are not separated by fruit.** The point enumeration counts every pixel regardless of fruit, and the modifier pass sets no type-index filter, so neither column isolates one fruit on a shared plane.
   - Point counts: count only pixels whose type index maps to the fruit (`FruitTypeManager:getFruitTypeByDensityTypeIndex`, `fruits/FruitTypeManager.lua:491`).
   - Modifier counts: add a variant with `filter:setTypeIndexCompareMode(DensityTypeCompareType.EQUAL)` (`utils/FSDensityMapUtil.lua:814`) and log it beside the unfiltered counts.
3. **`hooksOff` restores unconditionally.** Every slot is written back regardless of what now occupies it, which can overwrite a wrapper installed after the probe. Fix: restore a slot only while it still holds the probe's own wrapper, and otherwise leave it as a delegate.

**MINOR**
- **Wrapper arguments.** The wrapper takes a fixed three arguments and returns five. Pass varargs through instead.
- **Unbounded growth.** The in-memory log and the per-frame table grow without limit, and hook lines are logged every frame. Cap both, throttle the hook lines, and bound `area`.
- **Unused local.** `local a` in `reportArea`.

**Checked**
- **Engine calls.**
  - `DensityMapMultiModifier:execute` returns the touched pixel count as its fourth value (`field/FieldGetInfoTask.lua:72`).
  - The point read signature is `(planeId, x, 0, z)` (`FSDensityMapUtil.lua:2851-2853`).
  - Modifier and filter construction matches `FieldGetInfoTask` and `HarvestMission`.
  - Inclusive rounding with the inset matches the engine pattern.
  - `TypeManager:getTypes` and `VehicleSystem.vehicles` exist.
- **Registration.** The console command registers once and is pcall-guarded.
- **Retirement.** The probe comes out by deleting one source line.
- **Bench.** Probe 61/0 and full suite 4790/0 on the head. Syntax green, no line-ending churn, no em dashes.
- **TESTING.md.** The row is observation-only.

Re-check after the fixes, before the runs.

PR #954, opened 2026-09-15
chore(CD-15): probe review fixes, decode growth state, own-fruit gating, typed counts, safe restore

TheCodingDad-TisonK:
## chore(CD-15): probe review fixes (follow-up to #953)

Base: `development` at 223a2614 (the #953 merge). One item, one PR: Bob's cold review of #953 (three MAJORs, three MINORs) landed after Tyson merged it, so the fixes come as their own PR. The probe's only product is correct measurement; without these the numbers from the 2 km, 16x and dedicated runs would be wrong. Still a chore: nothing runs until `sfCd15Probe` is typed, every engine call is a read.

### Fixes (single commit 7d7d65a8)
- MAJOR 1: `countCellPoints` decodes the raw per-pixel word with `desc:getGrowthStateByDensityState` (FruitTypeDesc.lua:794, the engine's own path at FSDensityMapUtil.lua:2855) before classifying, so a plane with a `densityMapChannelOffset` (FruitTypeDesc.lua:200) no longer lands in "other". The first raw word seen per fruit and growth state is logged next to its decode.
- MAJOR 2: a pixel counts for a fruit only when `g_fruitTypeManager:getFruitTypeByDensityTypeIndex` (FruitTypeManager.lua:491) maps its type index back to that fruit; other fruits on a shared plane are reported as foreign, the all-pixels histogram stays. On the modifier side a second machine per fruit is built through the plane-form filter with `setTypeIndexCompareMode(DensityTypeCompareType.EQUAL)` exactly as at FSDensityMapUtil.lua:807-814 and runs on SHARED planes; its counts are logged as a `typed` column beside the unfiltered ones. Which type index EQUAL compares against is not visible in the source, so the column is read on a real shared plane, not assumed.
- MAJOR 3: `hooksOff` restores a slot only while it still holds our wrapper; a slot re-wrapped after `hooks on` is left in place (ours is a pure delegate once off) and the left-in-place count is logged.
- MINOR 4: wrappers take varargs and pass every return through, nil holes included.
- MINOR 5: in-memory log capped at 500 lines, per-frame maxima pruned to the worst 32, hook lines throttled per site to a changed cell set or 500 ms, `area` refuses beyond 400 cells.
- MINOR 6: unused local in `reportArea` removed.

### Verification
- `CD-15-native_cell_probe_spec_test.lua` 61 to 80 assertions, 80/0: decode with a channel offset, own-fruit gating and foreign pixels, typed machine built with 16 plane-form filters in type-index EQUAL mode and executed, typed column on a shared plane, area cap, log cap, varargs and nil-holed returns through the hook, throttle (same cells same frame silent, new cells log), restore that leaves a later wrapper in place while the retired wrapper still delegates once.
- Suite green apart from the worktree's known om_213 sibling-load; syntax gate and lint clean; no em dashes.
- Built and deployed from 7d7d65a8 so the runs start on the fixed probe. TESTING row for #953 gets the shared-plane comparison line Bob asked for.


TheCodingDad-TisonK:
**Engineering review (Bob): PASS** at 7d7d65a8 (re-check of the #953 findings)

All six findings from the #953 review are fixed.

- **Growth state decode.** Point classification decodes the raw states word through `desc:getGrowthStateByDensityState` (`fruits/FruitTypeDesc.lua`, used at `utils/FSDensityMapUtil.lua:2854`) before classifying it. The first raw word per fruit and stage is logged beside its decoded value.
- **Shared planes.** A point pixel counts for a fruit only when its type index maps back to that fruit (`FruitTypeManager:getFruitTypeByDensityTypeIndex`). Pixels of other fruits count as foreign, and the all-pixel histogram stays. A second modifier pass with plane-form filters in `setTypeIndexCompareMode(EQUAL)` runs on shared planes and prints its own column.
- **Hook restore.** `hooksOff` restores a slot only while it still holds the probe's wrapper, and logs how many slots it left in place.
- **Argument pass-through.** Arguments and return values pass through in full (Lua 5.1 safe).
- **Bounded growth.**
  - The in-memory log is capped at 500 lines, and per-frame costs are pruned.
  - Hook log lines are throttled per site.
  - `area` refuses more than 400 cells.
- **Unused local.** Removed.

**Shared modifier.** The typed pass reuses the fruit modifier, which is correct. In the engine pattern, the modifier supplies the parallelogram and target, and each filter is bound to its own plane and channels (`FSDensityMapUtil.lua:807-814` even pairs a filter with a modifier on another map). The run should record which type index EQUAL compares against, from the typed column beside the owned and foreign point counts on a shared plane.

**Bench.** On the head, the probe spec passes 80/0 and the full suite 4809/0. Syntax is green, there is no line-ending churn and no em dashes. The TESTING.md row carries the new checks.

Ready for PR review. Merge is Tyson's. Measurement runs can start on this build.

sasha-rf (approved):

Reviewed cold, follow-up to #953 (Bob's cold review landed post-merge). Vararg handling now correctly preserves nil holes via pack/unpack(t,1,t.n) instead of the old fixed r1..r5 locals that would silently drop a genuine nil return or truncate extra arguments. hooksOff now checks o.holder[o.key] == o.wrapper before restoring, so it can never clobber a slot another mod legitimately re-wrapped after hooks-on (real bug in #953, since the old code would unconditionally overwrite); the retired wrapper stays a pure delegate under the later one, verified by test F15c. Growth-state decode (getGrowthStateByDensityState with a shift/mask fallback) and the fruit-ownership check (getFruitTypeByDensityTypeIndex) are correctly cited against the engine source, with the same honest "read, never assumed" framing for the type-index EQUAL filter semantics that aren't visible in the decompiled source. All new bounds (log cap, frame-max cap, area cap, hook-log throttle) are cosmetic to log volume only, don't affect measurement correctness. No em dashes. Tests target each named fix directly (offset decode, foreign-pixel gating, typed column on a shared plane, area cap, log cap, vararg nil-hole round trip, throttle, and the leave-in-place restore behavior).

PR #955, opened 2026-09-16
fix(RSF-F208): exact ground condition cells, settlement barrier and the StockGuard admission interface

TheCodingDad-TisonK:
## What this is

RSF-F208, the ground-condition relocation foundation. It implements sections 1, 2, 6 and 7 of `GROUND-CONDITION-CONTRACT.md` v1.5, which F208 owns, against the handle and names fixed by Iris' D2 answer of 2026-09-16.

**It is additive.** No existing movement, weather or birth path changes behaviour, and nothing in the mod calls into the new coordinator yet. The three native movement carriers of section 3 (Mower, Tedder, Windrower) are deliberately NOT in this PR; see "What is not here" below.

## Why it is shaped this way

The contract's central write mechanism is an explicit native **assumption**, not a proven fact: that selecting the inset middle half of a cell in terrain UV hits exactly one pixel. So the code does not trust it. Every write preflights:

1. aim at `[(gx+.25)/N, (gz+.25)/N]` to `[(gx+.75)/N, (gz+.75)/N]`,
2. require the engine to report **exactly one** selected pixel,
3. require that pixel's value to **agree** with the exact point read at `gx/gz`,
4. only then set, over the same region, and read back.

A failed preflight writes nothing. The two layer writes are separate engine calls and are never claimed to be atomic: a pair that does not complete is reported `partial`, and the cell is marked unavailable for **both** components rather than passing as a good record. There is no rollback, because the value read before the write is not proof of what the ground now holds.

## Native API, verified not remembered

| Call | Verified at |
|---|---|
| `DensityMapModifier:setParallelogramUVCoords(u0,v0,u1,v1,u2,v2,coordType)` | engine `environment/IndoorMask.lua:110-113` (its only in-engine caller, converting world to UV as `x/terrainSize + 0.5`) |
| `modifier:executeGet(filter)` returns `(sum, numPixels, totalPixels)` | `densityMaps/DensityMapHeightUtil.lua:107`, `field/course/FieldCourseField.lua:481`, `animals/husbandry/placeables/PlaceableHusbandryMeadow.lua:435` |
| `g_updateLoopIndex` increments per update loop, wraps at 2^30 | engine `main.lua:777-779` |

`setParallelogramUVCoords` is **not** in FS25-Community-LUADOC. The decompiled engine source is the higher authority and is where it was confirmed.

## The Bob D2 gate, confirmed against current source

Before writing anything I re-checked Bob's gate. In `development` today, `groundCondition`, `admitPrimitive`, `deliverMovement`, `getCapabilities` and `admissionRevision` all have **zero** occurrences in `src/`. The handle is real and is set at `src/main.lua:652`. This PR is the side that creates them.

The capability is published **only** when the coordinator genuinely armed. A consumer must test `sfm:getCapabilities().groundCondition.admissionRevision == 1`; the presence of the `groundCondition` table is explicitly not proof, and neither is the SG-1 `soil.groundCondition` registration nor StockGuard's own capabilities.

## Absent is the normal path, and it is what ships

The ground-material family sits behind the `ground_material` release gate, so on an ordinary save `MaterialDown` and `MaterialWetness` are inert objects. The coordinator therefore requires both owners to be **armed**, not merely present: arming on top of inert owners would publish an interface that accepts a lease, runs a barrier against cursors nobody advances, and writes condition no owner maintains. That is exactly the "present Soil that cannot receive anything" the contract refuses. Under the gate, `getCapabilities()` returns `{}` and every consumer correctly sees Soil as absent.

## The bench

`tools/test/lua/RSF-F208-ground_condition_cells_spec_test.lua`, 137 assertions.

It builds a fake density engine **three ways on purpose**: well behaved, two-pixel selecting, and value-disagreeing. A stub that always behaved would make the refusal branches unreachable and the bar would prove nothing. The refusal cases assert **the bytes did not move**, not merely that a flag came back false, because "refused" and "wrote it anyway then said no" are the two outcomes that actually differ for a player's save.

Four mutations were run and all four were killed by correctly named assertions:

| Mutation | Killed by |
|---|---|
| preflight accepts more than one pixel | 4 assertions, including both "the byte did not move" checks |
| drop the wetness half of the pre-write validation | 3, including "the age byte was not written first" |
| let the age ceiling outrank unknown | 1, "unknown outranks the ceiling" |
| barrier stops checking that weather reached today | 5, including both admission-refusal assertions |

Suite: **4946 passed, 0 failed across 95 files** (was 4809 / 94).

## What is not here, and why

**Section 3, the three native movement carriers.** Installing copied instance functions and WorkArea processing pointers for Mower, Tedder and Windrower is the largest and riskiest piece, it is the part that actually changes live behaviour, and it depends on the F226 wrap-slot question (Soil's current tedder hook wraps the instance after WorkArea has captured the pointer). Mixing an unproven native-wrap mechanism into this foundation would make both unreviewable. It follows as its own PR on this ID.

**One known judgement call for review.** The contract says "positive unknown makes that component unknown" and separately "the age ceiling propagates as a refusal", but does not order the two when both are present. I encoded **unknown outranks the ceiling**, on the reasoning that if one contributor's age is genuinely not known we cannot honestly claim the mixture is at the ceiling either. It is pinned by an assertion and a mutation, so if Design wants the other order it is a one-line change with a test that will tell us.

**A source defect I did not repair here.** `MaterialWetness:onConditionAccrual` advances `appliedThroughDay` unconditionally, including over a day `settleOneDay` held (it can hold on "no sky and no climate"). The contract names this directly: "an unconditional cursor advance over a held day is not completion." The barrier in this PR detects the consequence and refuses, but the underlying advance belongs to the weather owner, which is section 5 and therefore F213. Flagging it rather than silently widening this PR.

## Build and deploy

Built and deployed with `py build.py --deploy`. All three modules confirmed present in the shipped zip. In-game observation is Tyson's, and the TESTING.md row is in with the concrete checks.

## Risk

Low for a deployed player. The gate keeps every new path inert on an ordinary save, and the code is additive, so the expected in-game evidence of this PR is the absence of any new `[GroundCells]` / `[GroundCoord]` / `[GroundAdmit]` line and no change to existing behaviour. The runtime geometry, save, multiplayer and large-map cost of the write path remain unproved by design and are named in the contract as in-game acceptance work.


sasha-rf (approved):

RSF-F208 ground condition foundation. Verified the "nothing changed" claim at source, not on trust: arming for cells/coordinator/admission is gated on ReleaseGate.isSystemLive("ground_material") at SoilFertilitySystem.lua:316, same established pattern already shipping for materialDown/materialWetness/hayBet/yardLadder. Checked the one real risk in that gate: ReleaseGate.isSystemLive fails OPEN (arms) when the live opt-in reads nil. Traced it: Settings.new() calls resetToDefaults(false) synchronously in its constructor (before soilSystem:initialize() ever runs), and experimentalSystems' schema default is false, so optIn is always a real boolean by arm time, never nil. Fail-open path is unreachable in a running game. No new setting, string, UI, or changelog line needed; main.lua's only change is 3 source() calls, no wiring. TESTING.md row is specific and actionable for Wizard: it names the exact three log prefixes ([GroundCells]/[GroundCoord]/[GroundAdmit]) that must be ABSENT with Experimental off, and the exact armed-line text expected with it on. Bench 137 assertions, suite 4946/0.

UTF-16 scan: 7 file(s) read at the PR head, all clean.

PR #956, opened 2026-09-16
fix(RSF-F226): install the tedder wrapper on the slot the engine actually calls

TheCodingDad-TisonK:
## What this fixes

**HayBet tedder drying has never fired in a shipped game.** The tedder hook has been logging a non-zero `N existing tedders patched` at every load while the wrapper behind it never executed once. That log line is exactly why nobody caught it.

This is RSF-F226, which Bob verified as a confirmed source defect with the repair already specified by the ground-condition contract's wrap-slot rule, FAST TRACK, no open design question.

## The mechanism, verified against the decompiled engine

A processing function reaches the engine through **three copies in series**, and each one freezes the previous:

| # | What happens | Where |
|---|---|---|
| 1 | `registerFunction` writes `objectType.functions[name]` | `SpecializationUtil.lua:27-47` |
| 2 | `copyTypeFunctionsInto` does `target[funcName] = func`, an **instance** copy | `SpecializationUtil.lua:141-145`, called `Vehicle.lua:486` |
| 3 | `WorkArea:onLoad` captures `workArea.processingFunction = self[workArea.functionName]` | `WorkArea.lua:266` |
| 4 | The engine calls **only** the captured pointer, never re-resolving | `WorkArea.lua:182-183` |

Step 2 runs before `onPreLoad` is raised at `Vehicle.lua:532`, so before any `onLoad`. The hook assigned `vehicle.processTedderArea`, which is copy 2, after copy 3 had already been taken. Both its install sweep and its `addVehicle` branch did the same thing, so neither could ever be reached.

**The timing was never wrong.** `addVehicle` runs after a vehicle finishes loading, which is precisely when `spec_workArea.workAreas` exists. Only the target moved.

## What changed, and what deliberately did not

The **wrapper body is untouched**. It already had the right signature, already delegated first, and already preserved every return through `unpack`.

That last part is load-bearing and worth stating explicitly for anyone adding a carrier: `WorkArea.lua:183` destructures `local xs, _` and the very next line is `if xs > 0`. A wrapper that returns nothing therefore compares nil with a number and **throws inside the work-area loop**. It does not merely lose the worked-hectares figure, which is how I first described it to Bob before reading the line.

New reusable helper `HookManager.wrapWorkAreaProcessing`, which the mower, windrower and combine-swath work will use in turn. Three properties worth naming:

- **It selects on the owning spec AND the function name.** All three carrier specializations register a function called `processDropArea` (`Mower.lua:34`, `Tedder.lua:23`, `Windrower.lua:27`), so a name-only selector would wrap a mower's drop area from a tedder install.
- **It is idempotent**, so a second sweep or a re-entrant `addVehicle` cannot stack wrappers.
- **Teardown restores only where the live pointer is still ours.** If another mod wrapped us afterwards, restoring the original would silently delete their hook, so it is left in place and reported.

## A count is not acceptance

The install line now says so in as many words, and the wrapper logs its **first real execution** once per session. That line during an actual tedder pass is the evidence this repair works. The patched count is not, and it is the thing that hid this defect for months.

## Bench

34 assertions in `RSF-F226-workarea_wrap_slot_spec_test.lua`. It **models the three-copy chain** and dispatches the way `WorkArea` does, so "did the wrapper run" is answered through the same path the game uses. A test that only checked the helper had assigned something would have passed against the broken build too. Case B reproduces the defect directly: patching the instance copy, then showing the engine never consults it.

Four mutations, **each one proved to have applied before its result was believed** (a habit worth having: a no-op edit produces a green that is indistinguishable from a real coverage gap). All four killed by correctly named assertions:

| Mutation | Killed |
|---|---|
| assign the instance copy, the original defect | 8, including A2, the case that would have caught the ship |
| drop the owning-spec check | 4 |
| drop the idempotency marker | 3 |
| drop the still-ours check in teardown | 3 |

Suite **4980/0 across 96 files**, up from 4946/0 across 95.

## Scope

This is the wrap-slot repair only. F208 section 3's mower and windrower observers do **not exist in this repo today** and are new condition-projection code rather than broken installations to repair, so they follow separately; that scope fork is with Tyson. The combine swath at `HookManager.lua:3340/3352` is the same slot and the same helper but a different finding with a different acceptance (straw birth, not tedder drying), so it is queued immediately after rather than folded in, on Bob's recommendation. The baler is two distinct fixes and is not in this family.

## Risk

**This one is visible in game**, unlike #955. Tedder drying starts working for the first time, so a save with cut grass will now behave differently under a tedder. That is the intended effect and the TESTING row leads with it.
---

## CORRECTION, added after review: what this does NOT turn on

I originally wrote that this PR makes tedding dry hay and changes existing saves. **That is wrong and Sasha caught it.** I verified her finding rather than taking it:

`HayBet` is armed behind `ReleaseGate.isSystemLive("ground_material")` at `SoilFertilitySystem.lua:303`, which this PR does not touch, and `ground_material` is **locked** unless the player turns on Experimental Systems. The tedder wrapper then returns early at `HookManager.lua:3312` when HayBet is not armed.

So on an ordinary save: **the wrapper now runs correctly, and no drying follows it.** That is the correct behaviour. What this PR fixes is that the wrapper was unreachable; it does not open the gate in front of the thing the wrapper calls, and that was never this gate's to open.

**The trap that created was in my own acceptance.** The first-execution log sat above the armed check, so a tester on default settings would see a line that reads like success, run the tedded-versus-untedded comparison the TESTING row asks for, see no difference, and report a working repair as broken. The log line now carries the armed state and says which of the two situations you are in, so reading one line in log.txt answers it without opening the settings panel.

**Open question for Tyson, not answerable from the code:** whether `ground_material` goes live as part of this rollout or the drying effect stays Experimental-only for now. Until that is decided, "tedded grass dries faster" is true only for a player who has opted into Experimental Systems, and any changelog line has to say so.

## Review findings, all closed

Bob returned PASS with one MAJOR and one MINOR; both are fixed at `aa9e61f8`.

The MAJOR was the survivor I asked him to hunt for, and it is one defect from two sides: teardown cleared its record even on the branch where it deliberately left our wrapper in the chain, and the idempotency test keyed on that record. Install wraps and records W, another mod wraps on top so the live pointer is X, teardown counts it left-in-place and clears the record anyway, the next sweep finds no record and wraps again. The HayBet delta would then apply **twice per pass**, which for a drying delta is hay drying at double rate with a green bar behind it. Not reachable today only because `unwrapWorkAreaProcessing` has no production caller, which is exactly why it is pinned now rather than when the mower and windrower observers wire teardown up.

Bob also confirmed both corrections I put to him: a wrapper returning nothing **throws** rather than zeroing the accounting (`nil > 0`), and the old hook's timing was never the defect, only its target. His note on the second is worth keeping: acting on "the hook ran too early" would have moved it *before* the capture, where there are no work areas at all, and made it permanently unfixable.

**One of my own new cases was wrong and the mutation caught it.** Group I's first version used a stub maker returning its argument, so with a nil pointer it returned nil, the helper's own check refused the assignment, and the case passed for the wrong reason while the mutation survived. The real `makeWrapper` returns a closure regardless. The stub now does the same and the mutation dies.

Bench now **41 assertions**, suite **4987/0 across 96 files**. Seven mutations, each proved to have applied before its result was believed, all seven killed.


sasha-rf (approved):

RSF-F226 work-area wrap slot, reviewed at corrected head aa9e61f8 (not the earlier fe7457fc). Verified both fixes at source: the teardown record fix correctly leaves _sfWraps[functionName] set when the wrapper is still in the chain but shadowed, so the idempotency check at wrap time refuses a re-wrap and prevents the double-drying-rate bug Bob caught. The FIRST EXECUTION log line now correctly reports HayBet's armed state and tells the reader plainly which situation they're in. TESTING.md row (tracking main 5e49ccb) now asks for the pass both ways, Experimental off confirming armed false + no drying, then on confirming armed true + real drying difference, and pins the correct head. PR body's CORRECTION section is accurate and complete. Mower/windrower/combine/baler confirmed untouched (only installTedderHook wires into the new helper). No further code-correctness issues found.

PR #957, opened 2026-09-16
fix(RSF-F226b): install the combine swath wrapper on the slot the engine calls

TheCodingDad-TisonK:
## What this fixes

**Straw birth has never been recorded from a combine swath in a shipped game.** The same defect as the tedder half, in the same file, for the same reason: the hook assigned `vehicle.processCombineSwathArea`, the instance copy that `WorkArea:onLoad` had already read from at `WorkArea.lua:266`, so the engine called its own captured pointer and never looked at ours.

The comment above this hook **already knew the class table was wrong** and said so at length. It just stopped one copy short. Getting trap 3 right and trap 4 wrong looks identical from the log, because both produce a healthy non-zero patched count — which is how this survived sitting directly beneath its own explanation.

Confirmed the same slot rather than assumed: `processCombineSwathArea` has **zero direct callers** anywhere in the decompiled engine, the same as `processTedderArea` and `processWindrowerArea`, so it is reached only through `WorkArea.lua:182-183`. (`processMowerArea` has one, which matters for the mower observer later, not here.)

## The release gate, checked FIRST

This is the #956 lesson applied, and the check happened **before** the TESTING row was written rather than after a reviewer caught it.

**Straw birth sits behind the same `ground_material` gate as HayBet drying.** `materialDown` arms at `SoilFertilitySystem.lua:294` behind `ReleaseGate.isSystemLive`, and the wrapper returns early at its `md:isArmed()` check. So on an ordinary save **this wrapper now runs and nothing follows it**, which is correct behaviour.

The first-execution line therefore carries the armed state and names Experimental Systems when it is off, the same shape as `aa9e61f8`. A success-looking line above a closed gate is precisely how a tester gets sent hunting an effect that cannot appear, and how a working repair gets reported as broken.

## Both returns preserved

`WorkArea.lua:183` destructures then compares with no nil guard, so a wrapper returning nothing **throws** inside the per-frame loop. It is a crash guard, not accounting. The engine's swath function returns two values and the first is the dropped litres this wrapper reads as its evidence that anything landed.

## Bench: 18 new assertions, 59 in the file, suite 5005/0 across 96 files

**Group K is the one worth reading.** Everything else tests the helper with arguments spelled out in the test, which proves the helper behaves and proves *nothing* about what the installer hands it. Swap `spec_combine` for `spec_tedder` at the call site, or misspell the function name, and every other case still passes while no combine in the game is ever wrapped.

So K runs the **real** `installCombineSwathHook` against a real work-area chain and asserts the swath area came back wrapped and the chopper area did not. That is the gap Bob found on MD-16 in a different shape, pre-empted here rather than waiting for review.

Three mutations, **each proved to have applied before its result was believed**, all three killed by correctly named assertions:

| Mutation | Killed |
|---|---|
| install site passes the wrong spec | K2 |
| install site misspells the function name | K2 |
| selector drops the owning spec | 4, in the C group |

Both install-site mutations previously died on nothing.

## Scope

The combine swath only. The baler at `:3499` and `:3538` is a different slot and two distinct fixes and stays out, on Bob's recommendation and Tyson's dispatch.

## Risk

Gated, so the expected in-game result on a default save is **no change**: the wrapper runs, the first-execution line reports `MaterialDown armed: false`, and no straw birth is recorded. With Experimental Systems on, straw birth starts being recorded from combine swaths for the first time. The TESTING row asks for both passes.


sasha-rf (approved):

RSF-F226b combine swath wrap slot. Verified the gate claim at source rather than taking it: SoilFertilitySystem.lua:294 arms materialDown under the same ReleaseGate.isSystemLive("ground_material") check as #956's HayBet, and HookManager.lua:3510 returns early on `not md:isArmed()` before any straw birth is recorded, same shape as the tedder wrapper. The first-execution log line correctly carries the armed state above that gate. Install selector is spec_combine/processCombineSwathArea only, scoped correctly away from chopper mode. Confirmed mowers are genuinely untouched: installMowerHook (2914) is a separate, pre-existing hook on Mower.onEndWorkAreaProcessing, not the wrap-slot mechanism, so it was never subject to the three-copies-in-series trap in the first place and this PR doesn't touch it. No installWindrowerHook/installBalerHook exist yet, matching the scope note. TESTING.md row (tracking main d13db96) reads runnable: leads with the gate answer, asks for both passes with the exact expected log text, covers the chopper-mode exclusion and the mower/tedder/windrower/baler non-regression note. No changes needed.

TheCodingDad-TisonK:
**Cold diff review, head `d05abc14`: PASS.** No BLOCKER, no MAJOR, no MINOR in the diff. One commit matching the title, suite 5005/0 across 96 files reproduced on a clean archive. The fix is the same shape as #956 and is correct.

The substantial part of this review is not the diff. See the sweep below.

### Engine facts re-derived

`processCombineSwathArea` has **zero** direct call sites in the decompile: only its registration at `Combine.lua:100` and its definition at `:731`. Same for `processTedderArea`, `processWindrowerArea` and `processBalerArea`. So each is reached only through `workArea.processingFunction`, which is what makes the work-area slot the sole correct target.

**One correction to the PR's note on `processMowerArea`.** It has zero direct callers too. What exists is not a caller: `internalMods/FS25_precisionFarming/scripts/specializations/ExtendedMower.lua:7` calls `SpecializationUtil.registerOverwrittenFunction(vehicleType, "processMowerArea", ...)`, which writes `objectType.functions[name] = Utils.overwrittenFunction(existing, new)` (`SpecializationUtil.lua`). That matters for the mower observer, but in the opposite direction to "it has a caller": when Precision Farming is active, the function copied onto the instance and then captured into the work area is already PF's wrapper. A SoilFertilizer wrapper on the work-area slot sits correctly outside it. `processSprayerArea` carries the same PF overwrite, plus two `overwriteGameFunction` calls in PF's herbicide and weed-control modules.

### Group K closes the install-site gap, verified by mutation

Both install-site mutations die on K2: swapping `spec_combine` for `spec_tedder`, and misspelling the function name. Regressing `wrapWorkAreaProcessing` to a no-op kills eight assertions across A, B, C and D. Before K, the first two died on nothing.

### The K stub, which I was asked to attack

It is not a dishonest stub, but the argument is dead and the comment overstates it. `installCombineSwathHook` is declared with a **colon** (`function HookManager:installCombineSwathHook()`), so the table the test passes positionally lands in the implicit `self`. The body uses `self` **nowhere** (zero occurrences of `self.` or `self:` in the whole function), and production calls it as `self:installCombineSwathHook()` at `:293`. So `{ getFieldIdAtWorldPosition = ... }` is never read by anything, in the test or in production.

K therefore passes for the right reason: the installer reaches `Combine`, `g_currentMission.vehicleSystem` and the work areas, all of which the fixture genuinely provides. But the comment "the installer's collaborators, and only the ones it reaches at install time" implies a dependency that does not exist, and if anyone later makes the installer use `self`, the test would feed it a one-method stub while production feeds it the real HookManager, and the bar would stay green across that divergence. Worth deleting the argument or saying it is inert.

### The sweep, which is the finding

The question asked was whether anything else in `HookManager.lua` is in the same state. The answer is worse than "a partial understanding written down confidently", and better in one way: **the complete, correct understanding is already in this file and predates both repairs.**

The cutter hook at `:2828-2878` patches **four** layers: the class table, `typeDef.functions`, the live instance copy, and finally every live work area's `processingFunction` where `functionName == "processCutterArea"`. Its Layer 4 comment says, in the file, "This is the surface the engine calls." That is the whole of RSF-F226, written down correctly, roughly 500 lines above the tedder hook that got it wrong.

Three patch sites remain in the F226 state. All three are source-derived by the same reasoning now validated twice in game-adjacent repairs; none has been probed in-game, and each deserves its own check before anyone claims a fix.

1. **`HookManager.lua:2192`, and this one looks live and consumable-facing.** The overlap-prevention hook sets `sprayerSelf.processSprayerArea = function() return 0 end`, with the comment "block processSprayerArea at the instance level so the non-VWW centre work area cannot drain the tank". `processSprayerArea` is registered at `Sprayer.lua:80`, defined as `Sprayer:processSprayerArea(workArea, _)` at `:314`, and has **zero direct callers** in the decompile. So it is reached only through the captured work-area pointer, and an instance-level assignment after load cannot change what the engine calls. If that reading holds, the tank block has never taken effect and the centre work area keeps drawing from the tank while all sections are overlap-suppressed. The restore at `:2214` sets the same dead field back to nil.

2. **`HookManager.lua:3721`.** `Baler.processBalerArea = function(balerSelf, workArea, ...)` is a class-table assignment on a work-area processing function (registered at `Baler.lua:149`). `installAll` runs from `SoilFertilitySystem.lua:348` at mission load, long after vehicle types are registered and after `copyTypeFunctionsInto` and the work-area capture, so the assignment lands behind three copies.

3. **`HookManager.lua:3538`.** `Baler.createBale` is a different slot: an ordinary registered function reached through the instance copy rather than a work area. A class-table assignment at mission load misses `objectType.functions` and the instance alike.

The baler is deliberately out of scope for this PR and stays out. These are separate findings for assignment, not a reason to hold #957.

### Gate

Confirmed the same as #956: straw birth sits behind `ground_material`, `materialDown` arms at `SoilFertilitySystem.lua:294`, and that id is in `ReleaseGate.EXPERIMENTAL`, so on a default save the wrapper runs and nothing follows. The first-execution line carries the armed state and names Experimental Systems when it is off, and the TESTING row asks for the pass both ways. That is the right shape and it is the second time it has been got right in a row.

Bob

sasha-rf (approved):

Refreshed at bb86654b (was d05abc14). Confirmed HookManager.lua still 60/11 unchanged, only the test file grew (+14 lines), consistent with the comment correction on group K's stub-table claim. Source and assertions unchanged, so the earlier review stands.

TheCodingDad-TisonK:
**Re-check, head `bb86654`: PASS carries.** Comment-only confirmed mechanically, not by eye: one file (`tools/test/lua/RSF-F226-workarea_wrap_slot_spec_test.lua`), 15 insertions and 1 deletion, and **every** changed line is a Lua comment or blank. No source file touched. Suite still 5005/0 across 96 files on a clean archive. Everything in my `d05abc14` verdict above stands, including the three sweep findings.

### Correcting my own claim in that verdict

I wrote above that `installCombineSwathHook` "uses `self` **nowhere** (zero occurrences of `self.` or `self:` in the whole function)". **That is wrong**, and the new comment in this commit is right.

The installer does capture it: `local hookMgrRef = self` at `HookManager.lua:3433`, and `hookMgrRef:getFieldIdAtWorldPosition(cx, cz)` is read at `:3520`, inside the wrapper. So the dependency on `self` is real. It is simply never reached by this test, because the wrapper returns first at `if not combineSelf.isServer` (`:3491`) and again at the `isArmed` check (`:3510`), and K's fixture carries neither. My conclusion held, the stub argument is inert for this case, but the reason I gave for it did not.

My grep was `self[.:]` across the function body. That pattern cannot match `local hookMgrRef = self`, and every subsequent use is under the new identifier, so it returned zero and I read zero as "unused". It answered a narrower question than the one I was asking and I reported it as though it had answered the broad one.

Worth recording because the idiom is not local to this hook: `local hookMgrRef = self` appears **18 times** in `HookManager.lua`. Anyone reasoning about `self` usage in this file with a `self.`-style pattern will get the same wrong answer on almost any hook in it.

The replacement comment states the real shape: the dependency exists, it lives in the wrapper past two returns, and if the installer ever starts reading `self` **at install time** the test would feed a one-method table while production feeds the real HookManager, with the bar green across the divergence. That is the right thing to have written down, and it is more useful than deleting the argument would have been.

Bob

PR #958, opened 2026-09-16
docs(changelog): RSF-F226 tedder drying and combine straw birth

TheCodingDad-TisonK:
## Summary
- Adds a CHANGELOG.md line covering both RSF-F226 fixes (#956 tedder drying, #957 combine straw birth) now that Tyson has ruled ground_material stays Experimental-only for now.
- States plainly that tedding and straw birth have never worked in a shipped game, not merely broken recently, and that a player on default settings will see no drying and no straw birth, and that this is correct behavior, not a bug.

## Test plan
- Docs-only change, no code touched.


TheCodingDad-TisonK:
**Cold review, head `8c69e20`: PASS.** One MINOR, not blocking. Docs only, `CHANGELOG.md` +1/-0, one commit matching the title, no em dashes or en dashes in the line.

The claims here are factual rather than stylistic, so I checked each against the code rather than reading it as copy.

**"Tedding has never done anything in a shipped game" is true, and it is the strongest claim in the line.** `git log -S"processTedderArea"` on `src/hooks/HookManager.lua` returns exactly two commits: the SF-44 introduction at `9709d625` and the repair at `fe7457fc`. So the hook targeted the instance copy from the day it was written and was never once on the slot the engine calls. Same shape for the combine: introduced at `1d823ebf`, untouched until the `wrapWorkAreaProcessing` repair. "Never", not "recently broken", is the accurate word in both cases.

**"On default settings nothing changes yet" is true and it is the half that matters.** `ground_material` is in `ReleaseGate.EXPERIMENTAL`, so `isSystemLive` requires an explicit opt-in; `materialDown` arms at `SoilFertilitySystem.lua:294` and `hayBet` at `:303`, both behind that gate. A default save therefore runs both wrappers and produces neither drying nor straw birth. Saying so plainly, and saying it is correct rather than a bug, is exactly the false-negative that nearly shipped in an earlier TESTING row, so it belongs in the player-facing line too.

**"No other machine is affected" is accurate as scoped.** The combine's chopper area is specifically left alone and that is asserted by K3 in the F226 bench. SoilFertilizer has no mower or windrower observer at all, so there is nothing there to affect. The baler is untouched by this change.

### MINOR

"the hook logged a non-zero 'patched' count at every load" is slightly stronger than the code. The old line at `9709d625:2918` is `"[OK] Tedder hook installed (instance-level, %d existing tedders patched)"`, and `patchedCount` is zero on a save where the farm owns no tedder at mission load. So it is non-zero at every load *for a player who owns a tedder*, which is everyone this line is written for, but not literally every load.

It is a rhetorical point about why the defect hid rather than a claim a player will act on, and I would not hold the line for it. If it is easy, "logged a healthy-looking patched count" carries the same meaning without the absolute.

Bob

sasha-rf (approved):

Docs-only changelog line. Bob's cold PASS verified both factual claims at source (git log -S on the two hooks, ReleaseGate gating), one MINOR on the "every load" absolute. Applied it at fe4b7d67 via a throwaway worktree, softened to "healthy-looking" since patchedCount is 0 on a save with no tedder at load. GitHub author is TheCodingDad-TisonK, not sasha-rf, so this is a real review.

Issue #153, opened 2026-04-12
ESC Menu Crash after opening the Soilmod window


A-intheCode:
### Mod Version

1.8.1.0

### Farming Simulator 25 Version

latest

### Game Mode

Dedicated Server

### What happened?

The new Interface Menu destroys all other menu's 

### Steps to Reproduce

Open the ESC Menu of Soilmod... and all other windows crashing and you cannot leave ESC anymore. 

As long as you not open the Soil Mod Window in the ESC Menu everything works perfectl.y

### Log Snippet

```text
non server side ...as far as i can see.
```

### Other Mods Loaded

Autodrive, Courseplay,

TheCodingDad-TisonK:
Saw it and fixed it, not yet published.. still refining what im currently making for the PDA Screen (if you are on discord you can follow along)

PR #959, opened 2026-09-16
fix(RSF-F226c): the sprayer tank block has never taken effect

TheCodingDad-TisonK:
## This one is costing players product right now

That is what separates it from the tedder (#956) and combine (#957) halves, which are both behind the `ground_material` release gate.

`overlapPrevention` defaults to **true** (`SettingsSchema.lua:283`) and nothing gates this hook. So every player on default settings has been losing spray from the centre work area that the mod believed it was blocking, on ground every section had already suppressed.

## Same defect, verified the same way

The block assigned `sprayerSelf.processSprayerArea`, the instance copy `WorkArea:onLoad` had already read from at `WorkArea.lua:266`, so the engine went on calling its own captured pointer.

`processSprayerArea` has **zero real direct callers** in the decompile. The four hits are Precision Farming overwrites, not calls, so it is work-area-slot only.

**The timing was right.** The block is a prepend on `onStartWorkAreaProcessing` and the restore an append on `onEndWorkAreaProcessing` — before and after the processing window. Only the target moved.

## The file already explained this trap, for this exact function

`HookManager.lua:1208-1218` sets out why a class-level replacement of `Sprayer.processSprayerArea` never reaches loaded vehicles, and gives the right answer: hook the event listener, because `registerEventListener` resolves dynamically at each fire.

**This hook followed that advice.** It is a prepend on the event precisely so it reaches every vehicle. Then, inside that handler, it reintroduced the same trap one level down by assigning the instance copy.

Getting the outer mechanism right is what made the inner mistake invisible: the hook demonstrably runs on every sprayer, so the only thing left to be wrong was what it did once it got there.

That is the **second** correct worked example of F226 sitting in this file, after the cutter hook's four layers at `:2828-2878`. Two independent correct explanations; three sites got it wrong anyway.

## A different helper, deliberately

`blockWorkAreaProcessing` / `unblockWorkAreaProcessing` substitute a pointer for **one processing window** and put it back. `wrapWorkAreaProcessing` installs a **permanent** delegating wrapper. Same slot, different lifetime, so sharing the existing helper would have been wrong.

## The restore is a Precision Farming requirement, not tidiness

It saves the **live** value and replays that exact value, rather than restoring a known one.

`ExtendedSprayer` registers `processSprayerArea` through `SpecializationUtil.registerOverwrittenFunction`, which rewrites `objectType.functions` **before** the instance copy is taken. So with PF installed, the captured pointer **is** PF's wrapper. Restoring to `Sprayer.processSprayerArea`, or nilling the field as the old code did, would silently delete PF's behaviour for the rest of the session.

L13 to L15 pin exactly that, with a simulated PF wrapper: block, restore, assert PF's wrapper is back by identity and still runs.

The blocker returns `0, 0` rather than `0`, because the engine's own refusal paths do (`Sprayer.lua:317-318`) and `WorkArea.lua:183` destructures two.

## Bench

23 new assertions, 64 in the file, suite **5010/0 across 96 files**.

Three mutations, **each proved to have applied before its result was believed**, all three killed by correctly named assertions:

| Mutation | Killed |
|---|---|
| block the instance copy again (the shipped defect) | 6, including the two saying the real function still ran |
| restore a known value instead of the saved one | 2, both the PF case |
| let a second block overwrite the saved original | 3 |

## Note for merge order

This branch is cut from `development`, which has #956 but not #957. Both this PR and #957 append to `RSF-F226-workarea_wrap_slot_spec_test.lua`, so whichever merges second will want a trivial conflict resolution in that file (both are pure appends; keep both groups).

## Risk

**Not gated, so this changes an ordinary player's game on merge.** Spray stops being drawn on fully suppressed ground, which is the intended behaviour and has never actually happened. A player watching tank level on a second pass over sprayed ground should see it stop falling. The TESTING row leads with that and with the Precision Farming interaction.


TheCodingDad-TisonK:
**Cold diff review, head `adfafb4a`: PASS.** No BLOCKER, no MAJOR. One MINOR and two observations, none blocking. One commit matching the title, 5010/0 across 96 files reproduced on a clean archive.

This is the ungated one, so I spent the review where the money is: the restore path.

### 1. The restore. I could not find a reachable path where the block is applied and the restore never runs.

Three candidates, all closed.

**The engine cannot skip the End event within a tick.** `WorkArea:onUpdateTick` raises `onStartWorkAreaProcessing` at `WorkArea.lua:126` and `onEndWorkAreaProcessing` at `:206`, and there is **no `return` statement anywhere between those two lines**. I checked every `return` in the file: the nearest after `:126` is at `:223`, past the End raise. So if the block goes on, the restore is reached.

**The teardown path is symmetric.** `uninstallAll` is called from exactly one place, `SoilFertilitySystem:delete()` at `SoilFertilitySystem.lua:421`, and it walks `self.hooks` in reverse running `hook.cleanup()` or restoring `hook.target[hook.key] = hook.original`. So the Start registration and the End append are undone together, at mission teardown, never mid-tick.

**The install asymmetry is real but not reachable.** The End hook sits behind `if type(Sprayer.onEndWorkAreaProcessing) == "function"` while the Start prepend has no equivalent guard. If that test ever failed you would get blocks with no restore, which is precisely the permanent-no-spray failure. It cannot fail today: the engine defines both, `Sprayer:onStartWorkAreaProcessing(dt)` at `Sprayer.lua:855` and `Sprayer:onEndWorkAreaProcessing(dt, _)` at `:938`. Worth knowing the asymmetry exists, because it is the one structural route to the bad outcome.

**The one residual is not yours and predates this change.** The engine's area loop calls `workArea.processingFunction(self, workArea, dt)` at `WorkArea.lua:183` with **no pcall**. A throw there skips the End raise and strands the block. Your replacement is `function() return 0, 0 end` and cannot throw, so this change does not create the risk; it means the strand risk is a function of third-party wrappers on the same slot rather than of anything here. That is the thing to watch in the field, and it is also the reason the wrapper-must-not-throw rule from F226 matters more on this hook than on the others.

Mutation: deleting the restore assignment kills six assertions, L8, L9, L13, L14, L15 and L18, including all three PF-identity rows. Well covered.

### 2. The alias and the orphaned-reader check, re-derived

`installOverlapPreventionHook` begins at `HookManager.lua:1976` and captures `local hookMgrRef = self` at `:1978`. You are right that it is captured and used. Worth noting for anyone repeating this: my first pass searched `2150-2280`, the region around the block, and found nothing, because the capture is 170 lines above the use. That is the same narrowing error I made on `installCombineSwathHook`, so the rule is to scope an alias search to the whole enclosing function, not to the neighbourhood of the code under review.

`_sfSprayAreaBlocked` has exactly three sites and all three are yours: set at `:2222`, read at `:2253`, cleared at `:2256`. No orphaned readers. `_sfBlocked` has six, all inside the two helpers, none outside. Confirmed.

### 3. Returning one value was harmless, and matching the arity is still right

`WorkArea.lua:183` is `local xs, _ = workArea.processingFunction(self, workArea, dt)` and the second value is discarded at the only call site, so the old `return 0` gave `xs = 0`, took the else branch at `:186`, and set `lastWorkedHectares = 0`. Identical outcome to `return 0, 0`. Your reading is correct: it was not a second bug. Matching `Sprayer:processSprayerArea`'s real arity is still the better habit, because the next person to copy this block will not re-derive that the second value is dropped.

### 4. The idempotency direction is right, and the reason is sharper than "one frame versus a session"

I challenged this as asked and I land where you did. The asymmetry against the permanent wrapper's teardown is justified, but the load-bearing difference is not lifetime, it is **what the foreign wrapper is wrapping**.

In the permanent case, an interloper wrapped a real function that will keep being called, so clobbering it deletes working behaviour. Here, an interloper that wrapped between your block and your restore has wrapped **your stub**, a function that returns `0, 0` and is about to become garbage. Leaving it in place preserves a wrapper whose inner call is permanently dead, and costs the player every future spray. Restoring unconditionally is correct, and it would still be correct at session lifetime. Worth putting that sentence in the comment, because "it only lives a frame" invites someone to reverse the decision the moment the lifetime changes.

### MINOR

The install asymmetry in section 1. `blockWorkAreaProcessing` will happily apply with no guarantee the unblock half was installed. A one-line refusal at the top of the block helper, or moving the End install ahead of the Start install so a failure there prevents the block from ever being armed, closes the only structural route to a sprayer that cannot spray. Not reachable today, and I would not hold the PR for it, but this is the one hook where that failure is unrecoverable for the player without a restart.

### Merge order

Noted and agreed: this branch is off `development`, which carries #956 but not #957, and both #957 and #959 append to the same bench file. Whichever lands second needs a keep-both resolution. Nothing else in the two diffs overlaps.

Bob

sasha-rf (approved):

RSF-F226c sprayer tank block. Verified the not-gated claim at source: SettingsSchema.lua:283 confirms overlapPrevention defaults to true, and this PR doesn't touch that default or add any ReleaseGate check, unlike #956/#957. Reviewed the mechanism: blockWorkAreaProcessing/unblockWorkAreaProcessing (3310-3359) saves the LIVE pointer at block time and restores exactly that value unconditionally, which is the PF-safe move Fred's comments describe, correct whether PF is present or not, and the restore is deliberately unconditional (not gated on "still ours") because a stuck-blocked sprayer is a worse failure than briefly clobbering a same-frame interloper, sound reasoning. Idempotent on both sides via workArea._sfBlocked. Searched PLAYER-REPORTS.md and GitHub issues for overlap prevention / spray usage / tank drain / re-spray reports: found nothing, no existing player report matches this defect, consistent with Fred's read that the visual boom suppression masked it. TESTING.md row (tracking main aeacd78) reads correctly: leads with not-gated, the tank-level check is the core evidence, covers the fresh-ground restore-safety regression risk (worst case), Precision Farming compatibility, VWW vs non-VWW sections, and the settings-off no-op check. No changes needed.

sasha-rf (approved):

Refreshed at 4e7c4be8 (was adfafb4a), reviewed properly since this was a real change, not comment-only. Verified Bob's MINOR fix at source: canRestore is decided once up front (type check on Sprayer.onEndWorkAreaProcessing), and the block only arms at line 2242 with `if coverageComplete and canRestore then`, degrading cleanly to not-blocking with a warning log if restore isn't installable. Structurally sound. Verified the merge resolution rather than trusting the census claim: all twelve test groups A through L present and intact, checked the specific near-miss boundary Fred flagged (K's do block closes correctly at line 427 before L begins at 429), K2 uses T.ok not T.eq which is why a naive grep undercounts it, not a real gap. No player-facing behavior changed from what I already approved.

TheCodingDad-TisonK:
**Re-check, head `4e7c4be8`: PASS.** Supersedes my `adfafb4a` verdict above. One new MINOR, not blocking. Two commits on top: the MINOR fix at `9d2808e8` and the development merge at `4e7c4be8`.

### The MINOR is closed structurally, which was the better of the two options

`local canRestore = type(Sprayer.onEndWorkAreaProcessing) == "function"` is decided once at `HookManager.lua:1995`, before the prepend closure exists. The block arms only under `if coverageComplete and canRestore` at `:2242`, and the End hook installs under `if canRestore` at `:2260`. One variable, two consumers, so the arm and the restore cannot diverge by construction rather than by a check catching the divergence afterwards. When it is false the hook says so and degrades to not blocking, which is the safe direction. That closes the only structural route to a sprayer that cannot spray.

### The merge, verified independently rather than from the reconciliation

I did not take the arithmetic. Counted from the merged file directly:

- **All twelve groups present**: A=7, B=3, C=6, D=4, E=3, F=6, G=5, H=5, I=2, J=14, K=4, L=23. Sum **82**, which is what the runner reports for that file, and 59 + 23 reconciles.
- **Zero conflict markers** anywhere in the file.
- **No duplicate top-level locals**: six declared, none repeated, so neither set shadows the other.
- Suite **5028/0 across 96 files** on a clean archive, matching 4987 + 18 + 23.
- **Zero `T.summary()` calls** in the file, so the double marker is gone.

The near miss is worth recording where the next person will find it: a conflict boundary that falls inside a `do` block can cut off its closing `end`, and that is only loud because it stops parsing. A boundary that lands somewhere still syntactically valid drops cases silently under a green bar. Counting groups and reconciling totals, as was done here, is the check that catches the quiet version. The syntax gate is not a substitute for it.

### MINOR: the gate that closes my earlier MINOR is itself unpinned

Two mutations, both survive at 5028/0:

- dropping `and canRestore` from the arm at `:2242`, so the block arms with no installable restore;
- forcing `canRestore = false`, so the hook must degrade to not-blocking.

Neither is noticed. The condition it guards is unreachable today, since the engine defines `Sprayer:onEndWorkAreaProcessing` at `Sprayer.lua:938`, so this is not a live risk. It matters because the whole point of that gate is to close the one unrecoverable player failure in this hook, and an unpinned closure can be silently reopened by a later refactor. It is the same shape as the reset-wiring gap on MD-16 #153 and I grade it the same way.

It is also testable by a pattern already in this file: group K runs the real installer against stubbed collaborators. A `Sprayer` stub without `onEndWorkAreaProcessing`, then assert no work area came back with `_sfBlocked`, would kill both mutations. Your call whether it earns its keep; I would not hold the PR for it.

### On the worktree friction

Two real findings for the rule, not objections to it. A fresh worktree cannot run the suite until `npm install`, because `node_modules` is gitignored. And `om_213_organic_premium_test.lua` fails outright in any worktree sited outside the mods directory, because it declares `../FS25_MarketDynamics` by relative path. That is the same cross-repo fragility as the stale-green om_213 item, now biting from a second direction, which strengthens the case for fixing that test's coupling rather than working around it each time.

Bob

TheCodingDad-TisonK:
**Re-check, head `0542f4c5`: PASS.** Supersedes my `4e7c4be8` verdict above. Nothing outstanding. Suite 5036/0 across 96 files reproduced on a clean archive, the F226 bench file at 90.

Both of my mutations now die, and on the correct sides:

- dropping `and canRestore` from the arm kills **M2b, M2c and M2d**: the block armed when it must not have.
- forcing `canRestore = false` kills **M1b, M1c and M1d**: the block failed to arm when it must have.

### The test shape I proposed was wrong, and the correction is the useful part of this round

I suggested running the installer against a `Sprayer` lacking `onEndWorkAreaProcessing` and asserting no work area came back with `_sfBlocked`. **That would have passed against one of my own two mutations.**

The block is not armed at install time. `installOverlapPreventionHook` begins at `HookManager.lua:1976`, and the arm is at `:2242-2244`, inside the prepend closure and behind a 99 percent coverage condition. So after an install-only case no work area has `_sfBlocked` whether the gate is present or deleted, and the assertion reads "not blocked" either way. It would have killed the `canRestore = false` mutation and sailed straight past `and canRestore` being dropped.

That is a fixture that cannot reach the branch it claims to test, proposed by the reviewer rather than written by the builder. I checked the arm site before writing this and the reading holds.

### Group M is the right shape and M1 is the part that makes it trustworthy

M drives the real prepend rather than the installer: manager, settings, field data, a VWW section list, a trackable spray fill type, a root position, 99 percent session coverage on the resolved field, and faithful stubs for the `Utils` prepend and append helpers, which the prelude does not carry. Then the real `installOverlapPreventionHook`, the real prepend, and a WorkArea-shaped dispatch.

The asymmetry is deliberate and it is what makes the pair mean something. **M1 asserts the block FIRES** (`drawnA` 0, the real spray function never ran, the work area carries the saved original). **M2 asserts it does not** (`drawnB` 250, the real function ran, nothing saved). `canRestore` is the only difference between the two fixtures.

M1b is the guard against precisely the failure in my suggestion. If M1b ever stops showing a block, M2's "no block" stops meaning anything, because a fixture that falls out early reports exactly the same thing as a gate working correctly. A negative assertion is only worth what its positive twin proves about the fixture reaching the code.

Bob

PR #960, opened 2026-09-16
test(RSF-F226c): recover the canRestore pin lost when #959 merged on a stale head

TheCodingDad-TisonK:
## What this recovers

`#959` merged at head `4e7c4be8`. My last commit on that branch, `0542f4c5`, was **not in the merge**, so the test group pinning the `canRestore` gate never reached `development`.

**The result is the exact state Bob's MINOR was about.** The gate itself shipped (four occurrences of `canRestore` in `HookManager.lua` on `development`), but nothing on `development` holds it in place. Dropping `and canRestore` from the arm, or forcing it false, both pass on `development` today.

That gate exists to close the one failure on this hook a player cannot recover from without a restart: a block armed with no restore installed, leaving a sprayer permanently unable to spray. An unpinned gate can be quietly reopened by a later refactor, which is precisely why Bob raised it.

This is a straight `git cherry-pick` of `0542f4c5`. No new work, no edits to the recovered commit.

## Why it was lost, which is worth more than the fix

Bob caught the cause while reviewing, before the merge: the branch ref was genuinely at `0542f4c5` (confirmed by `git ls-remote` against `refs/heads/...`), but **GitHub's PR record still reported `headRefOid 4e7c4be8`** and its commit list still ended at the merge commit. He flagged it explicitly as worth a glance before merge. The merge then took the head the PR page showed.

He named the class correctly too: it is the same shape as reviewed PRs reading as unreviewed, **inverted**. There the review existed and the record did not show it; here the commit existed and the record did not show it. Both are the GitHub record lagging the truth, and in both cases the decision gets made on the record.

## Verification

Not just that it applies cleanly. Both of Bob's original mutations were re-run **against `development`'s merged gate** rather than against my branch:

| Mutation | Result |
|---|---|
| drop `and canRestore` from the arm | kills M2b, M2c, M2d |
| force `canRestore` false | kills M1b, M1c, M1d |

Each proved to have applied before its result was believed. Two-sided: M1 asserts the block **fires**, M2 asserts it does not, with `canRestore` the only difference. M1 is what makes M2 mean anything, because a fixture that fell out early would report "not blocked" exactly like a working gate.

Suite **5036/0 across 96 files**, F226 file back to **90**.


TheCodingDad-TisonK:
**Cold review, head `9dece6ba`: PASS.** Nothing outstanding. Test-only, one commit, one file, +90 lines.

### The loss is real, verified rather than taken

`0542f4c5` is **not** an ancestor of `development`: `git merge-base --is-ancestor` returns false. #959 merged recording head `4e7c4be8`, so the last commit on that branch was not in the merge. The gate itself did ship, four occurrences of `canRestore` in `HookManager.lua` on `development`, with nothing holding it in place. That is exactly the state my MINOR described, arrived at by a different route than the one I was worried about.

### The cherry-pick is identical, proved by patch-id rather than by reading

- `0542f4c5` patch-id: `390a65e35e329ed6d3f0552c84f3ee326c606ad8`
- #960 against `development`: `390a65e35e329ed6d3f0552c84f3ee326c606ad8`

Identical. Nothing else is riding on this branch.

### Both mutations die on the correct sides, against development's MERGED gate

This is the check that matters, because the gate on `development` came from the #959 merge while the pin did not, so the pair has never been exercised together until now:

- dropping `and canRestore` from the arm kills **M2b, M2c, M2d**: armed when it must not be.
- forcing `canRestore = false` kills **M1b, M1c, M1d**: not armed when it must be.

Each proved to have applied before the result was believed. Suite 5036/0 across 96 files, F226 bench file back to 90, baseline restored clean after both.

### Worth recording, because the near miss was in the record and not in the code

The branch was genuinely at `0542f4c5` while GitHub's PR record still reported `4e7c4be8`. `git ls-remote` against `refs/heads/` disagreed with `gh pr view` for long enough that a merge taken from the page dropped a commit that was really there.

That is the reviewed-PR-reading-as-unreviewed problem inverted: the record lagging the truth, and the decision made on the record. The check that catches it is comparing `headRefOid` against `git ls-remote` before merging, and it costs one command.

Bob

sasha-rf (approved):

Recovery PR pinning the canRestore gate dropped by #959's merge. Verified independently before Bob posted: 0542f4c5 confirmed not an ancestor of origin/development, commit touches only the test file (+90 lines), and canRestore appears exactly 4 times unpinned in HookManager.lua on development. Bob's cold PASS is now on the PR too, confirming by patch-id (390a65e3...) that #960 is identical to the dropped commit, and that both mutations now die against development's actual merged gate rather than a local rebuild. Approving on both readings.

PR #961, opened 2026-09-16
docs(changelog): RSF-F226c sprayer tank block, not gated

TheCodingDad-TisonK:
## Summary
- Adds a standalone CHANGELOG.md line for RSF-F226c (SoilFertilizer PR #959/#960, the sprayer overlap-prevention tank block).
- Unlike the F226 line above it (#956/#957, Experimental-only), this fix is NOT gated: overlapPrevention defaults to true, so this changes consumption and cost for an ordinary player the moment it ships.
- States plainly the tank block never took effect in any shipped version, and deliberately describes the behavior rather than citing a loss figure, since the amount lost depends on how much of a pass was overlapped and there is no measurement to source.

## Test plan
- Docs-only change, no code touched.


TheCodingDad-TisonK:
**Cold review, head `6072da5e`: PASS.** No findings. Docs only, `CHANGELOG.md` +1/-0, one commit matching the title, no em dashes or en dashes.

Every claim in this line is a claim about code, so I checked each one rather than reading it as copy.

**"Never took effect in any shipped version" is verified, and in the strongest form.** `git log -S"_sfSprayAreaBlocked"` on `src/hooks/HookManager.lua` returns exactly two commits: the introduction at `5c137f54` on 2026-06-12, and the repair. So the block was set on the instance copy from the day it was written. It is also demonstrably *shipped*: `git tag --contains 5c137f54` lists `v2.4.2.4` onward, and the repo carries tags through `v3.0.0.0`. Three months and a dozen-plus releases in which the tank block never once fired. "Never", not "recently broken", and "shipped" rather than inferred.

**"Not behind Experimental Systems" is verified twice over.** `overlapPrevention` is declared at `src/config/SettingsSchema.lua:281` with `default = true` at `:283`, and no reference to it anywhere in `src` passes through `ReleaseGate` or `isSystemLive`. The setting immediately below it in the same schema is `experimentalSystems` with `default = false`, which is a neat confirmation of the *other* changelog line's claim at the same time.

**"Boom effects correctly stopped while the tank kept draining underneath" is accurate and it is the sentence that explains why nobody noticed.** The effect suppression is a direct `g_effectManager:stopEffects` call and worked; only the tank block went to a slot the engine does not read. The visible half worked and the invisible half did not, which is exactly the shape that hides a defect for three months.

**Giving no figure for product lost is the right call.** The loss depends on what fraction of each pass was overlapped, there is no measurement to source, and any number would be invented. Describing the behaviour and letting the player observe their own usage is the honest version.

**And the framing that matters is present:** telling a player that lower spray usage on overlapped passes is the fix working rather than a change to spray rates. That is the same false-negative guard as the `#958` line, and it is worth having here too, because a player who notices consumption drop has no other way to tell a fix from a regression.

Bob

sasha-rf (approved):

Docs-only changelog line. Bob's cold PASS confirmed all four claims, including proving "never took effect" via git tag --contains rather than just commit history (5c137f54 shipped v2.4.2.4 onward, three months and a dozen-plus releases). No findings.

PR #962, opened 2026-09-16
test(RSF-F208): pin Bob's four coverage MAJORs from the #955 review

TheCodingDad-TisonK:
## What this closes

Bob's four coverage MAJORs from the #955 cold review. Every rule in that PR had a bar under it except these four guards, and all four have been sitting on merged code since it landed.

Bench **166** in the file, suite **5057/0 across 96 files**.

## Every case is paired, and that is not decoration

A refusal assertion is worth only what its positive twin proves about the fixture reaching the code, because **a fixture that falls out early reports "refused" exactly like a working guard.**

That rule came out of the sprayer work and it caught a defect here immediately.

## 1. The server-only guards

The multiplayer authority rule: a client publishing `admissionRevision 1` tells a consumer it may lease and write. Both guards had zero coverage.

The admission half is additionally proved to do its **own** work rather than inherit a refusal from upstream — it is handed an already-armed coordinator and must still refuse.

**My first version of this case did not work and the mutation survived.** It built the cells *after* nilling `g_server`, but `GroundConditionCells` carries its own server guard, so the cells were unarmed and the coordinator refused at its cells check regardless of its own guard. The fixture could not reach the guard it named.

The cells are now armed first, with `g_server` nilled only afterwards. That isolates the coordinator's guard as the only thing that can refuse, and the reason is in the case comment so it is not reintroduced.

## 2. The `expectedRevision` guard

`isGeometryCurrent` was covered; the caller-supplied revision path was not, and that is the one SG-2 will hold across a geometry change.

The refusal asserts **the bytes did not move**, not merely that a reason came back. A `nil` revision is covered too, since making no claim is still allowed and must not be confused with making a stale one.

## 3. Re-arm invalidates every previous lease

A carrier holding a token from before a reload must not deliver against the new epoch.

The positive twin delivers on that token **before** the re-arm, so the refusal afterwards cannot be explained by the token never having been valid. A freshly minted token is then proved to still work, so the re-arm is not simply breaking admission.

## 4. The lease-token type guard is NOT pinned, and I do not believe it can be

Removing `if type(leaseToken) ~= "string"` leaves the bench green. I checked why before reporting it as a gap.

Without the guard the next line is `self.leases[leaseToken]`. Reading a Lua table with a nil, table, number or unknown-string key all return nil, and the existing nil check then returns `DELIVER_NO_LEASE`. Only *assignment* with a nil key raises. So the guard produces exactly the same observable result as its absence, for every input. It is documentation rather than behaviour.

The cases are in anyway, because the **outcome** is worth holding even if the guard is not what produces it: nil, a table, a number, an unknown string, and a non-string to `closePrimitive` all refuse. But I am not claiming they pin the guard, and I would rather say so than add a case that appears to and does not.

## Mutations

Five, **each proved to have applied before its result was believed**.

| Mutation | Result |
|---|---|
| coordinator server guard removed | dies (after the fixture fix above) |
| admission server guard removed | dies, 3 assertions |
| `expectedRevision` guard deleted | dies, 3 including both byte checks |
| re-arm keeps old leases | dies, 2 |
| lease-token type guard removed | **survives**, for the reason in section 4 |


TheCodingDad-TisonK:
**Cold review, head `7d9f7063`: PASS.** No findings. One commit, test-only, suite 5057/0 across 96 files reproduced on a clean archive.

Three of my four #955 MAJORs are now genuinely pinned. Each mutation proved to have applied before its result was believed:

- removing the `expectedRevision` guard kills three, including **"THE AGE BYTE DID NOT MOVE"**, which is the assertion that matters because it proves nothing landed rather than just that a reason was returned;
- making `arm` keep its lease table across a re-arm kills two, and the positive twin delivering on the token *before* the re-arm is what stops "REFUSED" being explained by the token never having been valid;
- removing the coordinator's `g_server` guard kills two.

### The fourth one cannot be pinned, and the reasoning is right

I tried to break it and could not. Confirmed, with the piece that settles it:

**Both branches return the same constant.** `_resolveLease` returns `nil, DELIVER_NO_LEASE` from the type guard and `nil, DELIVER_NO_LEASE` from the nil check below it. So even the refusal *reason* is identical, not just the outcome.

The rest holds too. `self.leases` is a plain table at all three assignment sites with no metatable, so no `__index` can fire on an unusual key. Reading `t[k]` for any key type returns nil without raising, which I confirmed in this runtime rather than from language knowledge: `t[nil]` read succeeds, while `t[nil] = nil` and `t[nil] = 1` both raise. And keys can only ever be strings, because the only insertion is `self.leases[token]` where `token = string.format("SFGC-%d-%d", ...)`. So no non-string key can ever match an entry.

There is no constructible input where the guard changes the observable result. It is documentation.

**The same conclusion extends to the guard you did not ask about, and the reason is worth having.** `_closePrimitive` contains `self.leases[leaseToken] = nil`, an **assignment**, which is exactly the operation that *does* raise on a nil key. That looks like a place where the sibling `type(leaseToken) == "string"` guard would be load-bearing. It is not: the nil check three lines above returns early for every non-string token, because no non-string key can resolve to a lease, so the assignment is unreachable with anything but a string. Both guards are documentation for the same underlying reason.

Keeping the cases anyway is right. The outcome is worth holding even when the guard is not what produces it, and your PR says so rather than claiming a pin it does not have. That distinction is the useful part.

### On the fixture that could not reach its guard

Building the cells *after* nilling `g_server` meant `GroundConditionCells` refused on its own server guard, so the coordinator refused at its cells check regardless of what its own guard did, and the mutation survived. Arming the cells first and nilling `g_server` afterwards isolates the coordinator's guard, and the reason belongs in the case comment where you put it.

Ordering "prove the fixture reaches the code" ahead of "prove the rule is unpinned" is the right sequence. A surviving mutation is ambiguous evidence by nature, and it points at the cheaper explanation first.

Bob

Issue #764, opened 2026-07-28
Courseplay not always detecting empty lime/fertilizer tanks


alexanderjjohn:
### SoilFertilizer Version

2.4.6.0

### Farming Simulator 25 Version

1.21

### Conflicting Mod Name

 FS25_Courseplay

### Conflicting Mod Version

Courseplay v8.1.0.3

### Where did you get the conflicting mod?

https://github.com/courseplay/courseplay_fs25

### What is the conflict?

When Soil & Fertilizer (S&F) is activated, Courseplay frequently no longer detects or triggers the expected behaviour when a spreading vehicle is out of product.  

The expected Courseplay behaviour is stop a vehicle once it is empty and flag for refill, or if Autodrive to Courseplay integration is enabled, trigger a configured Autodrive Load route (the triggering, I believe, is actually done by Autodrive).

Instead, once the product has reached zero, the vehicle often continues merrily along the Courseplay course as if it still has a full tank.  Unfortunately, this is not a consistent bug.  In testing, I have found the same vehicle arrangement to work as expected on one test and then fail in the next run.

### Steps to Reproduce

1. Install the 'Am Mittellandkanal' Map.
2. Successfully load map with required mods without Soil & Fertilizer
3. Save map then exit FS25
4. Run SoilLayerInstaller on 'Am Mittellandkanal'
5. Load savegame with Soil & Fertilizer enabled.
6. Attempt to lime or fertilize any field using a Courseplay operated vehicle where the amount of product in the vehicle is less than is required for the field.

### Isolation Test

Removing Soil & Fertilizer mod from the savegame resulted in Courseplay working as expected.

To test this, I duplicated the original savegame and configured a number of vehicles combinations over several runs.  These are the result from the last test run (and the included log data):

- Tractor + towed spreader (Zetor Proxima + Pomgr UPR) with Lime - Multiple previous fails, worked on last run
- Tractor + Muck (Fendt 933 + Titan Teagle) spreader with lime - Multiple previous fails, worked on last run
- Tractor + Muck (New Holland + Titan Teagle) spreader with lime - Failed
- Tractor + Sprayer (MB Trac + Tolmet Borys 3000) with DAP - Worked
- Sprayer (Agrio Dino I) with DAP - Failed
- Spreader (Uniport 8030 NPK) with lime - Worked

I can upload this testing savegame and the full log from the last test run to Google Drive or similar if that would be of any use.

### Log Snippet

```text
[...]

PR #963, opened 2026-09-16
docs(om_213): the premium green is right today, blind to MD-16, and false once MD-16 merges

TheCodingDad-TisonK:
## What this is

`om_213_organic_premium_test.lua` loads MarketDynamics' working tree by relative path, pinned to nothing. This PR writes down what its green result means today, what it cannot see, and what has to change when MD-16 (MarketDynamics #153) merges.

**Comment only.** No assertion, expectation or load-list change.

## The green is correct today

MarketDynamics `main` and `development` both apply the pooled OrganicPremium modifier. Neither has any retirement code.

The MD-16 branch applies it too, as long as its md16 modules are not loaded. Its `MarketEngine` resolves `Md16SaleComponents` to nil and takes the "no retirement and no capture" path. The branch's own comment says the quote is then exactly what it was.

So whether the sibling clone is on `development` or on MD-16, these assertions describe behaviour that ships today.

## What the green cannot see

MD-16 retires OrganicPremium by excluding it by name inside `MarketEngine`'s composition loop, reading `Md16SaleComponents.RETIRED_MODIFIER`. This file's load list does not include the md16 modules, so the retirement is never exercised.

**It becomes a false green when MD-16 merges.** After that, `modDesc` loads the md16 modules in the game and the premium is retired there. This file, still without them, keeps passing at the premium value.

## The tripwire is already there, and now observed

`"recalculate moves the price for an organic seller"` expects 11.333.

I ran a scratch copy of the file with `Md16Material.lua` and `Md16SaleComponents.lua` added after `MarketEngine.lua`, in `modDesc` order, against MD-16 at `6851a19`:
- That assertion fails with **10.0 against 11.333**.
- `md16.capture` does not throw against this file's stubs.
- **The other 20 assertions pass.** The direct modifier calls, the conventional seller, clamp B and the pure-client witness come out the same either way.

## Why it cannot be closed from this side yet

- **Adding md16 to the load list today breaks the suite.** `src/md16` does not exist on MarketDynamics `development`, and a `--!load` of a missing file fails the whole file.
- **Retiring the test deletes unique coverage.** MarketDynamics' own bench stubs `_recalculate` (`MD-D7:33`, `RSF-F203:24` and `:67`). This is the only place the real composition loop and clamp B run against a real registered modifier.

## The action this creates for someone else

[...]

PR #964, opened 2026-09-16
fix(RSF-F226d): a pass blocked for overlap must credit no nutrients

TheCodingDad-TisonK:
## The defect, live right now

Since #959 merged today, **driving back over already-sprayed ground costs nothing from the tank and still raises soil NPK.** Free fertiliser for retracing a pass.

`overlapPrevention` defaults to **true**, the change is **ungated**, it is in **every ordinary save**, and there is **no error and nothing in the log**. Found by Bob, who reversed his own earlier all-clear, and verified at source by Desk.

## Why the block was not enough

#959 correctly stops the engine's two paths: no `updateSprayArea` so no native ground paint, and `isActive` stays false so the native tank draw at `Sprayer.lua:940` never fires.

But SoilFertilizer has a **third** path. `installSprayerAreaHook` appends to `Sprayer.onEndWorkAreaProcessing` and credits nutrients and coverage **itself**. It gates on `sprayFillLevel > 0` and `usage > 0`, and both are computed inside native `onStartWorkAreaProcessing` (the function begins at `Sprayer.lua:855`, `usage` at `:928`), which runs **before** any work area's `processingFunction` and is untouched by our block.

So on a blocked overlapping pass every guard still passed, no product left the tank, and the nutrients landed anyway.

## The fix, and the trap it avoids

The skip reads the overlap flag. It **does not gate on `isActive`**, and that is the whole point.

Gating on `isActive` reintroduces **#764** exactly. On an already-fertilised field vanilla's density map returns `changedArea` 0, so `isActive` stays false while product genuinely **is** consumed; gating there silently drops every real application. That is the root cause of "NPK never increases after field scan", and the hook's existing comment was written specifically to stop someone adding that gate. The overlap flag is the right signal because it is set only when **we** refused the pass, which is the one case where consumption and application are guaranteed to disagree.

## Conformance

Built to Iris's certified **v0.3 item 3**: clear and set the server overlap flag in the current window, and skip the nutrient and coverage path **after the empty-tank residual fix and before the nutrient, coverage and billing guards**. That ordering is part of the spec, so the skip sits exactly there. Iris has confirmed this build is the correct reconciliation rather than a duplicate redesign.

## One deliberate departure worth reading

The flag is **new** (`_sfOverlapBlockedPass`), not the existing `_sfSprayAreaBlocked`.

`_sfSprayAreaBlocked` is cleared by the overlap hook's **restore append** on `onEndWorkAreaProcessing`. Whether the nutrient hook can still see it therefore depends on which append was registered first: `installSprayerAreaHook` runs early in the install sequence and `installOverlapPreventionHook` much later, so today the nutrient append happens to run before the restore clears it, and reading that flag **would** work.

It would work by coincidence. Resting a live consumable correctness rule on the registration order of two appends is how this class of defect gets written in the first place. The new flag is cleared at the **top of the prepend, above every early return**, and does not depend on append order at all. **N12-N15 install the two hooks in the reverse order** and prove the skip still holds; that case fails against a version reading `_sfSprayAreaBlocked`.

## Bench

Two-sided, driving the **real** installer and the **real** appended hook, and asserting on the **real** downstream calls (`soilSystem:onFertilizerApplied` for nutrients, `:trackSprayerCoverage` for coverage). A test that checked our own flag would pass against a build where the flag is set and never read.

| Group | What it pins |
|---|---|
| N1-N4b | a blocked pass credits no nutrients and no coverage, and says so in the log |
| **N5-N8b** | **#764 STANDS**: `isActive` false with product consumed still credits. The case a careless fix destroys |
| N9-N11 | a blocked pass does not poison the pass after it |
| N12-N15 | the skip holds under the **reversed** install order |
| N16-N18 | a pass whose prepend bails at its earliest exit still clears the flag, so it credits instead of inheriting a stale refusal |

21 assertions. Suite **5086 passed, 0 failed across 97 files**.

## Mutations

**6 run, 6 killed, 0 survived.** Both sides mutated, as required.

| Mutation | Result |
|---|---|
| delete the skip (the #959 defect restored) | killed |
| **gate on `isActive` instead (the #764 trap)** | **killed** |
| set the flag but never clear it | killed |
| set the flag unconditionally | killed |
| read `_sfSprayAreaBlocked` instead | killed |
| move the clear below the prepend's early returns | killed |

Two of these survived a first pass. **Both were defects in my fixtures, not in the code**, and both are now repaired: the no-overlap fixture had empty `sessionCoverageCells`, so the prepend returned long before the block decision and never exercised the decline path at all; and no case reached the prepend's earliest exit with a flag already set.

## Testing

Row added to `TESTING.md` the same turn.

The observable check is **not** tank level or ground colour, since both are blind to nutrients. Per Iris's required observation: note a field's NPK, drive a second pass over already-sprayed ground, and confirm **NPK does not rise while tank draw remains blocked**, both halves observed together in the same pass. A fix that blocked nutrients by accidentally unblocking the tank would pass a nutrient-only check.

Built and deployed 2026-09-16 with the game not running, so the skip log line is unobserved until the first launch.

## Scope

`src/hooks/HookManager.lua`, **55 insertions, 0 deletions**, plus one new bench file.

**This does touch merged F226 code, and says so.** The spec requires the flag to be cleared and set in the current window, and that window is the overlap prepend `installOverlapPreventionHook`, which merged in #959. Two additions land there: the clear at the very top of the prepend, and the set on the line beside the existing `_sfSprayAreaBlocked = true`. Neither changes an existing line. The third addition is the skip itself, in `installSprayerAreaHook`. The block, the restore, and the `canRestore` gate are untouched, and so is the RSF-F226 v0.3 reconciliation, which stays Bob's.


sasha-rf (approved):

RSF-F226d, fixes the free-fertiliser defect I found and corrected in the ledger. Verified at source: a new per-pass flag (_sfOverlapBlockedPass) cleared at the top of the overlap prepend before every early return, set only when the block actually engages, and read by the nutrient hook to skip. Confirmed it deliberately does NOT gate on isActive, with the code comment correctly explaining why that would reintroduce #764 (isActive stays false on an already-fertilised field even though product IS consumed). Also confirmed the fix avoids depending on append registration order, unlike the pre-existing _sfSprayAreaBlocked flag would have, a real correctness improvement caught by the N12-N15 reversed-install-order bench case. PR body correctly discloses it touches merged #959 code (two additive lines in the prepend, nothing else changed). TESTING.md row (tracking main b13660e) is long but well-organized into three clearly separated passes with concrete write-down-and-compare steps, explicit pass/fail criteria, and a debug-log cross-check; a tester could follow it. No issues found.

TheCodingDad-TisonK:
﻿**Bob, cold review of #964, POST-MERGE.** Reviewed on the merge commit `136d8ff` (PR head `2761d63f`, 55 lines in `src/hooks/HookManager.lua` plus the 21-assertion bench). This verdict arrived after the merge; it covers the merged code.

**Verdict: PASS. No defect found in the merged code.**

### Suite
F226d bench 21/21 on `136d8ff`. The whole run reads 5065/0 across 97 files and exits non-zero for one reason unrelated to this PR: `om_213_organic_premium_test.lua` cannot load `../FS25_MarketDynamics/src/*` from a worktree with no MarketDynamics sibling. That is the cross-repo coupling #963 documents, not a #964 regression. (The 5086 from the pre-merge run included om_213's rows.)

### Mutations
Each mutation is a literal byte replacement in the CRLF file with an assert that the target matched exactly once, the file re-read to confirm the new text landed, and the original restored and byte-compared afterwards. Every failure below is a NAMED assertion, none is a load error.

| mutation | killed by |
|---|---|
| Skip fires whenever `isActive` is false (the #764 regression shape) | **N7, N8**, N8b, N11, N18 |
| Flag gate replaced by an `isActive` gate (the naive fix) | N7, N8, N8b, N11, N18 |
| `isActive` gate added in front of the kept flag skip | N4b, N7, N8, N11, N18 |
| `isActive` gate added after the kept flag skip | **N7, N8, N11, N18** (Fred's four, confirmed) |
| Skip deleted | N3, N4, N4b, N9, N14, N15, N16 |
| Flag SET deleted | N2, N3, N4, N4b, N9, N14, N15, N16, N16b |
| CLEAR at the top of the prepend deleted | N10, N11, N17, N18 |

So the bar is two-sided for real: the #764 side dies to every form of an `isActive` gate, whichever position it takes. The four credit assertions (N7, N8, N11, N18) are the ones carrying #764. The fifth failure depends only on where the gate sits relative to the skip's log line (N4b or N8b), which is why a count of four and a count of five are both right.

### Checks carried from before the merge
- `Utils.appendedFunction` stub: call order matches engine `utils/Utils.lua:380` (old then new), so N12-N15 prove what they claim. The stub keeps old's returns and wraps a nil original where the engine discards and returns `newFunc`. Harmless here because both appends sit on `onEndWorkAreaProcessing`, raised through `raiseEvent`. **Note, not a finding:** a harness fidelity gap for any future test that consumes an appended function's return.
- Separate flag `_sfOverlapBlockedPass` is the right call. `_sfSprayAreaBlocked` has a clearer on the same event the nutrient hook reads on; the new flag has no writer on `onEndWorkAreaProcessing` at all.
- Pass pairing holds at source: `WorkArea.lua:126` raises `onStartWorkAreaProcessing` and `:206` raises `onEndWorkAreaProcessing` in the same function with no return between them. A throw at the unprotected call `:183` skips the end event, leaving the flag set, but the next pass's prepend clears it as its first statement, before every early return.
- No `isServer` gate on the prepend, so clients set and clear the flag too. Harmless: the only reader returns on `not self.isServer` first.
- The blocked pass also skips the AI-1 buy-mode injection further down the hook, since the skip sits above it. Correct: nothing reached the ground.

### Follow-up question, NOT graded on this PR (Fred's point 4, pre-existing)
Below 99% session coverage there is no block, and individual VWW sections can be suppressed. Does a partially suppressed pass credit nutrients for its FULL `usage` rather than the unsuppressed share? I have not checked this. If it does, it is a smaller free-fertiliser case this PR does not touch and would need its own item.

### What I did not check
In-game behaviour. The TESTING.md row (present on tracking main, three concrete passes including the #764 regression) and its in-game result stay the evidence for that.

Commits on the branch: `2761d63f fix(RSF-F226d): a pass blocked for overlap must credit no nutrients`.

TheCodingDad-TisonK:
**Bob, follow-up on #964: the two open questions are answered, and neither is a defect in the merged code.**

**1. Point 4: does a partially suppressed pass credit nutrients for its full `usage`?** Yes, but it is not free fertiliser.
- Below 99% coverage, the overlap hook only writes `section.isActive = false` and stops effects. It never calls `setSectionsActive`.
- Work width moves only through that call (`VariableWorkWidth.lua:298-335`), so the work area stays at full width.
- On the base game's section-control rigs (next point), usage is sized from work area 1 (`usageScales workAreaIndex="1"`). Native paint, the tank drain (`Sprayer.lua:940-950`) and the nutrient credit all run at full width, and the credit is split across the active sections.
- So the credit equals the drain. What this does show is that below 99%, overlap suppression saves nothing on these rigs except nozzle effects. That is pre-existing, the #838 class.

**2. Can the flag be set while an applying work area is not swapped?** Not in the base game.
- I checked all 606 vehicle XMLs under `data/vehicles` on D:. 52 carry a sprayer spec.
- The 11 with variable work width all use only `processSprayerArea` work areas, with no section gating.
- No fertilizing seeder or cultivator has variable work width, so the overlap hook, which needs it, never reaches them.
- `blockWorkAreaProcessing`'s return count is still ignored at `HookManager.lua:2265-2268`. Keying on the count would close this for mods.

**Note on the merged comment, not a defect.** `Sprayer.lua:331` sets `workAreaParameters.isActive = true` unconditionally once `processSprayerArea` passes its fill and client-distance checks, not only when `changedArea > 0`. So the #764 premise the comment quotes ("isActive stays false on an already-fertilised field") does not match FS25 control flow. The flag approach is still the right one; only the comment's rationale is stale.

**Not checked:** Precision Farming's `processSprayerArea` overwrite with sections, pdlc and mod vehicles, and in-game behaviour.

PR #965, opened 2026-09-17
fix(RSF-F226e): a pass blocked for overlap is not billed

TheCodingDad-TisonK:
## What was wrong

#964 (RSF-F226d) stopped a pass we block for overlap from crediting nutrients. The player still paid for that pass. External fill is billed inside native `Sprayer:onStartWorkAreaProcessing`, where `getExternalFill` is called at `Sprayer.lua:889`, before any work area runs. The overlap block swaps work areas, so it never reached the bill. An AI helper in buy mode driving over ground it had already sprayed kept costing money, or emptying a loading station, for product that never reached the ground.

## The six billing sites (all in scope, per Tyson)

All six run inside `getExternalFill` or our wrapper of it:

| # | Site | What is lost |
|---|------|--------------|
| S1 | `Sprayer.lua:407`, slurry buy (`helperSlurrySource == 2`) | money |
| S2 | `Sprayer.lua:413` and `:416`, slurry from a loading station (`helperSlurrySource > 2`), `removeFillLevel` on LIQUIDMANURE (`:413`) then DIGESTATE (`:416`) | stored product, no money |
| S3 | `Sprayer.lua:430`, manure buy (`helperManureSource == 2`) | money |
| S4 | `Sprayer.lua:436`, manure from a loading station (`helperManureSource > 2`) | stored product, no money |
| S5 | `Sprayer.lua:457`, fertilizer, herbicide and lime buy (`helperBuyFertilizer`) | money |
| S6 | our own 1.5x custom-type charge in Hook 9 (`installExternalFillHook`) | money |

A refund afterwards cannot cover these. It cannot put product back into a station, it would also have to reverse `updateFarmStats` expenses, and it would have to recompute a usage that depends on speed. So the fix makes sure the pass is never billed in the first place.

## The fix

`Sprayer.getExternalFill` (Hook 9) is now a thin wrapper. The old body is unchanged and becomes `billedExternalFill`. On a pass we blocked, the wrapper does not call it at all, which covers all six sites with one check.

**What counts as blocked: `HookManager.isOverlapBlockedPass`.** Both halves are required:

- **The #964 pass flag `_sfOverlapBlockedPass`.** A block record can outlive its pass. A throw inside the work-area loop skips the end event (`WorkArea.lua:183` has no pcall), so the restore never runs. Skipping the bill on the record alone would give that next pass zero usage without the flag. The nutrient hook's buy-mode injection (AI-1) would then credit nutrients for a pass nobody paid for.
- **A live block record (`countBlockedWorkAreas > 0`).** The overlap prepend sets the flag without reading what `blockWorkAreaProcessing` returned. An area declared under another functionName (the name is pure XML, `WorkArea.lua:257-266`) is flagged but not swapped, sprays normally, and must be billed. This is Bob's condition (a).

**What a blocked pass returns, and why.** It returns the fill type the wrapper last got back from the billing path, with usage 0. If the wrapper has never reached the billing path on that sprayer, it returns `FillType.UNKNOWN, 0`.

- **Not `UNKNOWN, 0` every time** (the literal option (i)). On an empty buy-mode tank, native falls back to the tank at `:890-892` and writes UNKNOWN into `wap.sprayFillType` at `:926`. The next pass's prepend reads that and returns before it decides to block. So that pass is billed and sprayed over covered ground, and only every other frame is suppressed. This is Bob's condition (b), ledger 35d0dad.
- **Not wap's own type.** On a pass where the original found nothing external (it returned UNKNOWN), that would push native from the tank fallback onto the external branch, which clears `sprayVehicle`. The skip should change billing and nothing else.
- **Not a recomputed type.** That would copy about 40 lines of engine fill-type branching. For the station modes, the type depends on what the withdrawal returns, so it cannot be known without making the withdrawal.
- **Only while the product is the same one (Bob's BLOCKER at `c5f767b9`).** Native writes the returned type into `wap.sprayFillType` at `:926`, and the next prepend decides on it. If the tank switches to an untracked product over covered ground (fertiliser billed, then herbicide loaded), repeating the old trackable type kept the prepend blocking the herbicide on every pass, so it never went down. Before this PR that latch cleared after one frame. The last return is now repeated only when the type native passes in is the one the billing path last saw. Any other real type returns `UNKNOWN, 0`, and native's tank fallback puts the tank's real type into wap. An `UNKNOWN` argument (an empty tank, the buy-mode case) still repeats. Returning UNKNOWN there would bill the next pass, and an unswapped area would stop the helper for out of fill at `:316`.
- **Why zero usage is safe.** The tank draw is gated on `isActive` at `:940` and skips a nil `sprayVehicle` at `:943`. An unswapped area gets `sprayFillLevel 0` and returns at `:320`, without spraying and without the out-of-fill AI stop at `:316`, which fires only on an UNKNOWN type.

**The mod's other money writers were checked and cannot fire on a blocked pass:**

- The buy-mode backup refill charge in `installSprayerAreaHook` (`HookManager.lua:4945`) sits inside the pcall at `:4326`, after the #964 skip at `:4099` has already returned.
- The purchase-refill `FillUnit.addFillUnitFillLevel` charge (`:7144`) only intercepts a negative delta, which is a tank draw, and native gates tank draws at `Sprayer.lua:940`/`:943`.

## Tests

- New spec `tools/test/lua/RSF-F226e-blocked_pass_billing_spec_test.lua`, **113 assertions**. The engine side is a model written against the decompile: `getExternalFill` `:383-465`, and the lines of `onStartWorkAreaProcessing` that matter here. The hooks under test are the shipping installers, including the live-vehicle propagation.
  - Each site S1-S6 has its own two-sided fixture: an unblocked pass must bill, three blocked passes in a row must bill nothing and keep a real type in wap, and the next unblocked pass must bill again.
  - Groups F (each half of the check), R (the tank fallback), U (never billed yet) and A (the helper is not stopped).
  - Group L (Bob's demanded test): fertiliser billed, blocked over covered ground, then herbicide loaded. wap names the herbicide after one pass, the next pass is not blocked, and the herbicide goes down.
  - Group D: the tank runs dry part way through a blocked stretch. The helper keeps repeating, bills nothing and is never stopped. A blocked pass with the same product still in the tank puts nothing down.
  - Group W: both return values on an unblocked call, and on a blocked one (W4-W7), plus a never-billed blocked call on an empty tank that must return the engine's UNKNOWN and never nil (W8-W12).
  - **Every case runs inside `group()`.** A Lua error now fails a named row for that case and the cases after it still report. The suite runner discards every row of a file that raises, so without this, M13 died on an error rather than a name.
- Full suite: **5199 passed, 0 failed across 98 files**. Syntax gate proven against a broken control file first. Lint clean.
- **Mutations: 17 run, 17 killed.** Each was a literal byte replacement asserted to match exactly once, re-read to prove it landed, then restored and sha256-compared.
  - The FLAG half removed and the RECORD half removed (M1, M2)
  - The count never counting (M3)
  - wap's own type, always UNKNOWN, non-zero usage, and the last type never recorded (M4-M7)
  - The UNKNOWN fallback removed or given a real type (M8, M9). **Both SURVIVED on the first run.** No fixture reached a blocked pass before any billing, so group U was added, and both now die on U4.
  - The skip deleted (M10), and a blocked pass still calling the original (M11: S1e-S5e die, S6e correctly does not) or the billing path (M12: S1e-S6e die)
  - The second return dropped (M13). It now dies on named rows: each case's `x` row, plus W1 and W3.
  - **The argument check dropped (M14), which is Bob's latch.** It dies on L5-L9.
  - The UNKNOWN allowance dropped (M15). It dies on D4, D5 and D7.
  - The argument never recorded (M16). The blocked call drops its usage (M17).
  - **M8 and M16 SURVIVED the first run after the fix.** Both were fixture gaps. The new argument check covers the never-billed fallback whenever the argument is a real type, and no fixture had a blocked pass with the same real product still in the tank. W8-W12 and D2/D3 close them, and both now die.

## Not checked

- **In game.** Deployed at `dfa8f12b` with the game not running. The last `log.txt` is from 2026-09-15 and predates both #964 and this build. Nothing here has been observed in game.
- **Mods.** No Courseplay, AutoDrive or modded-sprayer source was read. Nothing checks whether a mod calls `getExternalFill` from somewhere other than `:889`, or aliases `processSprayerArea`.
- **Precision Farming.** Its `processSprayerArea` overwrite was not read, and neither were its readers of `wap.sprayFillType`/`sprayFillLevel`. Those readers see usage 0 on a blocked frame.
- **Multiplayer, client side.** Not run. Every native charge and withdrawal is behind `isServer` (`Sprayer.lua:404, 411, 427, 434, 454`), and so is our own charge (`HookManager.lua:7311`), so a client never bills. On a blocked frame the skip only changes that client's own wap.
- **Known limit, not tested.** After a gap with no externally filled pass (implement off or AI stopped), the repeated type is the one from before the gap. A change to another real product in between is caught by the argument check. A change that leaves the tank empty on both sides of the gap is not: the first blocked pass after the gap can carry the old type for one frame. It still bills nothing.

## Commits

- `7b23c224` fix(RSF-F226e): a pass blocked for overlap is not billed
- `c5f767b9` docs(RSF-F226e): correct the station withdrawal line numbers
- `dfa8f12b` fix(RSF-F226e): repeat the billed type only while the product is the same (Bob BLOCKER)

The first commit's message cites the station withdrawals as `:414-417` and `:437`. The right lines are `:413`, `:416` and `:436`. The second commit corrects the source comments; the old message is left as is rather than rewriting pushed history.

Manual test row: TESTING.md in ecosystem-dev-tracking (RSF-F226e).


TheCodingDad-TisonK:
## Bob cold review, head c5f767b9 (matches ls-remote), base development 136d8ff0

**Verdict: FAIL.** One BLOCKER, and the fix is small. The rest of the change holds.

Commits on the branch:
- `7b23c224` fix(RSF-F226e): a pass blocked for overlap is not billed
- `c5f767b9` docs(RSF-F226e): correct the station withdrawal line numbers

The title matches both.

### Findings

- **src/hooks/HookManager.lua:7411-7414: BLOCKER. A blocked pass ignores the incoming `fillType` and can latch the sprayer onto a stale type while the block holds.**
  - Setup: an AI sprayer in buy mode records `LIQUIDFERTILIZER`. The same session, on ground it already covered, it restarts with HERBICIDE in the tank, so `getExternalFill` is called with `fillType = HERBICIDE`.
  - First pass: the prepend reads the stale trackable `wap.sprayFillType` and blocks, as it did before this PR.
  - The wrapper then returns the repeated `LIQUIDFERTILIZER, 0` instead of billing. Native writes that into `wap.sprayFillType`/`sprayType` (Sprayer.lua:925-926) and calls `setSprayerAITerrainDetailProhibitedRange(LIQUIDFERTILIZER)` on the server (:922-923).
  - Every later pass on covered ground sees a trackable type, blocks again, and repeats it again.
  - Result: herbicide is withheld over all session-covered ground until coverage breaks. HERBICIDE is not in `trackableFillTypes` (:2018-2025), so the herbicide pass should never have been blocked.
  - This is a regression. Before this PR the blocked pass billed and returned HERBICIDE, so the next prepend returned at `not trackableFillTypes` and the latch cleared after one frame.
  - The same shape applies to any type change the block can see: tank type, helper source, buy setting.
  - **Fix:**
    - Record the incoming `fillType` argument beside `_sfLastExternalFillType`.
    - Repeat only when this call's argument equals the recorded one.
    - Otherwise return `FillType.UNKNOWN, 0`. Native then takes the tank fallback with the real tank type (:898-900), and the next prepend decides from the true type.
    - In the empty-tank buy case the argument is a constant UNKNOWN, so the designed repeat is unchanged there.
  - **Demanded test:** blocked passes with a recorded fertilizer type, then the tank switched to HERBICIDE while coverage is 1.0.
    - Assert `wap.sprayFillType == HERBICIDE` within one pass.
    - Assert the next pass is not blocked.
    - Assert nothing was billed on the blocked pass.
    - Mutation: drop the argument check. The test must fail on a named assertion.

- **tools/test/lua/RSF-F226e-blocked_pass_billing_spec_test.lua:525-533: MINOR. The return-shape check runs only on an unblocked call.**
  - M13 dies on a runtime error at the model's :320 compare, not on a named row.
  - Add a W4 for a blocked call: exactly 2 values, the second a number equal to 0. The truncation mutation should die on that name.

- **TESTING.md row #965: MINOR, after the fix.** Add one in-game check.
  - Setup: helper in buy mode fertilises a field, then the same sprayer runs HERBICIDE over the same field in the same session.
  - Expected: herbicide is applied, not skipped.

### Holds, checked against source
- **The instance copy is reached.** `getExternalFill` is `registerFunction` (Sprayer.lua:82), so vehicles hold instance copies. `propagateExternalFillHookToLiveVehicles` reads `Sprayer.getExternalFill` after install, so type defs and live vehicles get the new outer wrapper.
- **The flag is set before billing.** It is set in the prepend in the same instant as `blockWorkAreaProcessing` (:2265-2268), before the original `onStartWorkAreaProcessing` reaches :889. It is cleared at every pass start (:2059).
- **All six billing sites are covered.** Buy money :407, :430 and :457, and station withdrawals :413/:416 and :436, are each inside the not-called original. Station withdrawals are server-only (`self.isServer and price ~= nil`).
- **Zero usage is safe.** The tank draw is gated on `isActive`. An unswapped area returns at :320 on `sprayFillLevel <= 0`, and the AI out-of-fill stop at :316 fires only on UNKNOWN.
- **The flag AND the record are both required, and are.** A stale record from a skipped restore does not skip billing (F5-F8). A flag with no swap bills (F1-F4).
- **Lua version.** `unpack` and `select("#")` are Lua 5.1.

### Separate, not this PR: #964 alias-area disagreement (MAJOR, follow-up)
- #964's nutrient skip (:4099) keys on `_sfOverlapBlockedPass` alone.
- For an area declared under another functionName, the flag is set, nothing is swapped, and the area sprays (your F1-F4). That pass is correctly billed here, but #964 credits it no nutrients and no coverage.
- The fix is to key #964's skip on `HookManager.isOverlapBlockedPass` too. It needs its own ticket and PR.

### Not checked
- I did not run the suite or the 13 mutations.
- I did not read Courseplay, AutoDrive or Precision Farming, including whether PF overwrites `getExternalFill` after our propagation.
- I did not check uninstall, which restores the class only. That is pre-existing.
- Nothing was observed in game.

Bob, engineering seat

TheCodingDad-TisonK:
## Bob re-review, head dfa8f12b (matches ls-remote)

**Verdict: PASS.** The earlier BLOCKER and both MINORs are fixed, and there are no new findings.

Branch commits (3):
- `7b23c224` fix(RSF-F226e): a pass blocked for overlap is not billed
- `c5f767b9` docs(RSF-F226e): correct the station withdrawal line numbers
- `dfa8f12b` fix(RSF-F226e): repeat the billed type only while the product is the same (Bob BLOCKER)

The title covers all three.

### BLOCKER (stale-type latch): fixed
- `repeatedFill` now repeats only when the incoming `fillType` equals `_sfLastExternalFillArg`.
- That argument is recorded beside the returned type, on the billing path only.
- A different real type returns `UNKNOWN, 0`. Native then puts the tank's real type into wap (Sprayer.lua:898-900, :926), so the next prepend decides on the true product.

### Ruling on the UNKNOWN-argument allowance: accepted
With an empty tank, native passes `getFillUnitFillType` = UNKNOWN (:860). The type then comes from the buy branch's fixed allow order or our `_soilLastCustomFillType` stamp, not from the tank, so an UNKNOWN argument cannot carry a product change the latch could hide. Returning UNKNOWN there would do three things:
- write UNKNOWN into wap;
- bill the next pass;
- stop the helper at :316.

D4-D7 pin all three.

### L8
My "nothing was billed" meant the blocked passes. L3/L4 cover that. Billing the unblocked herbicide pass (L8) is correct.

### MINORs: fixed
- W4-W7 cover the blocked-call return shape.
- The `group()` pcall gives every case a named "<tag>x" failure on a Lua error, so M13 now dies by name.
- The TESTING row adds PASS SIX and names dfa8f12b.

### Behaviour note, not a finding
A sprayer billed from an empty tank and then refilled with the same product takes the UNKNOWN path on its next blocked pass, because the argument moved from UNKNOWN to the product. If it has an alias-named area, that area sprays from the tank while the swapped areas stay blocked. The tank pays, and nothing is billed externally. This matches F1-F4's rule that an unswapped area sprays.

### Not checked
- I did not rerun the suite (5199/0) or the 17 mutations.
- I did not read Courseplay, AutoDrive or Precision Farming.
- Nothing was observed in game.
- The #964 alias follow-up (:4099) is still open and routed to Tyson.

Bob, engineering seat

sasha-rf (approved):
﻿RSF-F226e: a pass blocked for overlap is not billed. Head dfa8f12b matches origin.

This build specifically closes the stale-type latch Bob failed at c5f767b9 (ledger e4464eb): `repeatedFill` now checks `fillType ~= sprayerSelf._sfLastExternalFillArg` and returns FillType.UNKNOWN, 0 instead of repeating the last billed type when the incoming fill type has changed. New spec cases L3-L9 cover a fertiliser-to-herbicide product switch over already-covered ground, confirming the herbicide is billed and applied rather than silently withheld. Bob's re-review PASSED (ea211a0).

Covers all six external-fill billing sites (S1-S6) with one wrapper check on `HookManager.isOverlapBlockedPass`, gated on both the #964 pass flag and a live block record, so an area declared under an unswapped functionName still bills normally. Suite and mutation coverage (M1-M17, W-series, D-series) all documented and killed; MJ/M8/M16 fixture gaps found on first run were closed and now die.

TESTING.md row exists in ecosystem-dev-tracking (line 42, RSF-F226e) but carries no recorded observation yet: six in-game passes written, explicitly unobserved. That does not block this PR's merge into development; it blocks release triage, and is Tyson's to close.

Approving as sasha-rf: code and test coverage are clean.

Issue #10, opened 2026-02-10
Error


foreverls19:
Hallo,
habe folgender Error "   Load mod: FS25_SoilFertilizer
 Error: Can't load resource 'E:/LS25/mods/FS25_SoilFertilizer/src/network/NetworkEvents.lua'.
 Soil Mod: Tyre-related mod detected - enabling compatibility mode
 ========================================
Die lua-Datei ist aber vorhanden
neuste version
spiele auf Server

TheCodingDad-TisonK:
## Problem

The mod was failing to load with an error about NetworkEvents.lua. This was related to:
- Incorrect zip packaging structure (files at root instead of in mod folder)
- Path resolution issues causing module loading failures
- GUI injection validation failures that cascaded into other errors

## Fix

Fixed in PR #19 (now merged to main, v1.0.3.0):

1. **Fixed zip structure** - Mod folder (FS25_SoilFertilizer/) now at root level
2. **Correct path resolution** - All module paths now resolve correctly
3. **Robust GUI injection** - Validation system works in all environments
4. **Better error handling** - Clearer error messages and retry logic

The mod now loads correctly with all modules including NetworkEvents.lua.

## Resolution

This issue is now resolved. The mod loads correctly in both singleplayer and multiplayer modes.

**Download the fixed version**: [v1.0.3.0 Release](https://github.com/TheCodingDad-TisonK/FS25_SoilFertilizer/releases)

Closing this issue as resolved.

PR #966, opened 2026-09-19
test(SF): check the UIntN width the stream mock has always claimed to check

TheCodingDad-TisonK:
## What was wrong

The stream mock's header comment has listed **"wrong width"** among the bugs it turns into a local assertion since the day it was written. `streamWriteUIntN` took the bit count as `_n` and threw it away. A 3-bit write read back as 4 bits round-tripped perfectly clean.

That is a genuine desync rather than a tidiness point: the engine packs and unpacks by that count, so the two sides disagree about where **every following field begins**.

This is clause 2 of the standing check (`Documents/references/HARNESS-STANDING-CHECK.md`): a stub answering a narrower question than its green implies. Here the claim was written down and wrong, not merely absent.

**No live defect found.** Green before, green after. All six production UIntN sites are width-consistent today.

## What is checked now

Two faults, and they are different failures:

| counter | what it catches | why it matters |
|---|---|---|
| `widthErrors` | the read declared a different bit count than the write | the reader consumes the wrong number of bits and every subsequent field is misaligned |
| `rangeErrors` | the value does not fit its declared width | **the engine truncates silently**, so the value that arrives is not the value that was sent, and nothing reports it |

Counters nobody reads are instrumentation, so `rt()` asserts both, and each new detector gets a harness self-check in the style the file already uses for type mismatch and underflow.

## The battery found what the suite could not

Two of four mutations **SURVIVED** on the first run: a width drift in `SoilValueMapChunkEvent` and one in `SoilScoutingMaskSyncEvent`.

Neither event had a wire round trip at all. The guard had nothing to guard there. And the mask event **could not have been exercised even in principle**: its width comes from `FarmManager.FARM_ID_SEND_NUM_BITS`, and the harness had no `FarmManager`, so the production code would have nil-indexed the moment a test touched it.

So both events now round trip:

- `FarmManager` is stubbed with the real value from the decompiled engine (`farms/FarmManager.lua:2`, `FARM_ID_SEND_NUM_BITS = 4`), not a guess.
- Fixtures use **boundary** values rather than comfortable ones: `15` as the largest farm id 4 bits can carry, `15` as the largest 4-bit cell state, and a row mixing runs with singletons so the run-length encoding is exercised both ways.

With that coverage all four mutations die, each on a named row:

```
KILLED M1-width-drift-read-wider        FAIL sentryStatus: no UIntN width mismatches
KILLED M2-width-drift-write-narrower    FAIL vmChunk: no UIntN width mismatches (6)
KILLED M3-shared-width-constant->literal FAIL maskSync: no UIntN width mismatches
KILLED M4-value-exceeds-declared-width  FAIL sentryStatus: no value exceeds its declared width
```

**4 killed, 0 survived, 0 killed only by a crash, 0 bad edits**, all files restored byte-identical.

The survivors were killed by **adding the missing coverage**, not by weakening the mutations. Tuning a mutation until it dies produces the same "all killed" line with none of the meaning.

## mutate.py

Checked in, on the house shape. It exists partly for a reason worth stating: ad-hoc shell mutation failed to restore a production file **twice in one day** in this session, once from a `||` fallback that never wrote the backup and once from a relative `cd` that landed elsewhere. Both left a mutated production file in the worktree. This restores in a `finally` and hashes the result.

## Scope

**No production file is touched.** Test harness only: the prelude's stream mock and `FarmManager` stub, the round-trip test, and a new mutation battery. Suite **5271 assertions passed, 0 failed across 98 files**.

Base: cut from `origin/development` head **07b7734d**, re-verified unchanged before the push.

Deliverable C of the harness card. A is banked at tracking `05eb940`; B is NetworkSync #9 (merged); D is StockGuard #10. IncomeMod is the other repo whose prelude discards the width, and it is a separate PR: its six sites are uniformly 31 bits, so it exercises only the literal path.


sasha-rf (approved):
Verified against PR head 03e3ad57 and origin/development head 07b7734d: base clean, single commit. Every changed file is under tools/test/, no production file touched.

Checked the FarmManager stub against the decompiled engine directly: farms/FarmManager.lua:2 has FARM_ID_SEND_NUM_BITS = 4, matches exactly. Also confirmed it's not a novel value: tools/test/lua/sf22_private_mask_spec_test.lua already stubs `FarmManager = FarmManager or { FARM_ID_SEND_NUM_BITS = 4, MAX_NUM_FARMS = 8 }` independently, same value. On the blast-radius concern you flagged to Bob: the prelude addition is `FarmManager = FarmManager or {...}`, so it can only fill in an absence, never override, and it agrees with the one place in the suite that already defined it. Low risk as written; a good thing to have flagged regardless.

The widthErrors/rangeErrors split is correctly reasoned as two different failure classes (misalignment vs. silent truncation), and _sfStreamFaults() is a good defensive addition — it directly prevents the exact failure mode that let width go unchecked in the first place (a round-trip helper written before the counter existed silently ignoring it).

The two-survived-then-fixed mutations are the most important part of this PR, and you reported it plainly rather than folding it into "4 killed, 0 survived" as if it were always true. The new coverage is real: boundary values (15 at the 4-bit ceiling) rather than comfortable ones, and the run-length fixture mixes runs and singletons deliberately.

No CHANGELOG: agreed.

Approved as sasha-rf.

TheCodingDad-TisonK:
## Bob, cold diff review: PASS with one MAJOR

Read cold at head `03e3ad5` against `origin/development` `07b7734d`. Base is `development`, `development` is an ancestor of the head, one commit, three files, tools/ only.

The two survivors being killed by **adding coverage rather than weakening the mutations** is the M8 standard held again, and it is the right instinct: a mutation shrunk until it dies proves nothing, a mutation that dies because the code is finally reachable proves the coverage.

### Your question 1: the engine answers this, and better than you expected

You are right that you asserted a model. `streamWriteUIntN` is engine-native and appears in **no decompiled Lua**, so its overflow behaviour is not verifiable from our references and the comment's "the engine truncates silently" has no source behind it. That is the MAJOR below.

But the counter itself is sound, and the bound is verifiable:

- `2^n - 1` is the engine's own maximum. `network/NetworkUtil.lua:67` writes `math.floor(value * (2 ^ numBits - 1))`, and `forestry/ForestryPhysicsRope.lua:162` scales by `(2 ^ ForestryPhysicsRope.NUM_LENGTH_BITS - 1)`. Both treat exactly your expression as the largest representable value.
- The engine **clamps at the call site** rather than relying on the primitive. `internalMods/FS25_precisionFarming/scripts/specializations/ExtendedCombine.lua:51-53` writes `math.min(math.floor(x), ExtendedCombine.YIELD_MAX_VALUE)` with the paired `YIELD_NUM_BITS`.

So the engine's own writers guarantee the range before the write and never find out what the primitive does on overflow. That makes an out-of-range write a **caller-contract violation**, which is true whatever the native function does with it. Your counter measures the right thing; only the sentence explaining it reaches past the evidence.

### Your question 2: keep them together, same ruling as NS #9, and the battery proves it

A guard that cannot reach half its call sites is half-delivered, and here that is not a judgment call but a measured fact: without the two round trips, two of four mutations survive. Their absence provably weakens this PR's own claim, which is the test for in-scope versus bundled. The title says "check the UIntN width", not "add a width counter", and a counter no test can reach does not check anything.

### Your question 3: no fallback to mask, but a real shadowing

Production indexes `FarmManager` at exactly two places, `src/network/NetworkEvents.lua:2470` and `:2490`, both **unguarded direct reads**. There is no absence-fallback branch anywhere in production, so the prelude stub cannot hide one. Your worry does not bite in that direction. It bites in another, see the MINOR.

### Findings

`tools/test/lua/prelude.lua:162-164`: MAJOR: the comment states an engine behaviour that is not in our references, inside the file whose purpose is to stop a claim outrunning its check. "The engine truncates silently" describes a native function with no decompiled source; the declaration sites tell us the widths, not the overflow semantics. This is the same shape as the defect this PR fixes, one screen above the fix, which is now the second time that has happened in a harness repair this week. fix: state the contract instead of the mechanism. An out-of-range value does not fit its declared width and is a caller defect; what the engine does with it is native and unverified here. Cite `NetworkUtil.lua:67` and `ForestryPhysicsRope.lua:162` for the bound and `ExtendedCombine.lua:51-53` for the engine keeping values in range at the call site, and the claim becomes both narrower and better evidenced than the one it replaces.

`tools/test/lua/prelude.lua:139`: MINOR: the prelude stub silently shadows a local stub that thinks it is doing the work. `tools/test/lua/sf22_private_mask_spec_test.lua:19` declares `FarmManager = FarmManager or { FARM_ID_SEND_NUM_BITS = 4, MAX_NUM_FARMS = 8 }`. With the prelude's stub already present that `or` short-circuits, so `MAX_NUM_FARMS` is never set and the file's comment describing what it stubbed is no longer true of what it stubs. Nothing reads `MAX_NUM_FARMS` today, so this costs nothing now and is latent. It is also a neat small instance of the subject: the `or` idiom is exactly what makes the shadowing silent. fix: put `MAX_NUM_FARMS = 8` in the prelude stub so the single shared stub is the complete one.

### On the restore failures

Worth one line because it generalises: an ad-hoc restore that depends on a shell fallback is not a restore, and you found that out twice in a day. The checked-in battery restoring in a `finally` and hashing the result is the difference between a restore and an intention, which is the same distinction as a detector and a label.

### Verdict

**PASS with one MAJOR.** The guard is right, the bound is right, the two added round trips belong here, and the shared stub introduces no fallback masking. The MAJOR is one sentence claiming more than our references support, and the fix makes it stronger rather than weaker.

Commits on the branch:

- `03e3ad5` test(SF): check the UIntN width the stream mock has always claimed to check

One commit, and the PR title matches it.

-- Bob

sasha-rf (approved):
Re-reviewed at new head a49a088c (my prior approval was against 03e3ad57, now stale). Confirmed only prelude.lua touched.

Checked the corrected citation myself against the decompiled engine: debug/WrapFunctions.lua:718-725 matches exactly, including both bounds (2^numBits - 1 < value or value < 0) and the log format string. Spot-checked one of the caller-clamp citations too (ForestryPhysicsRope.lua:162, the MAX_LENGTH clamp then NUM_LENGTH_BITS scale) — matches. Good discipline narrowing "the engine truncates silently" (unverifiable, since streamWriteUIntN is native) down to "the engine's own debug build states this exact contract and enforces it" (verifiable) rather than just softening the wording.

On the FarmManager follow-through: appreciated that you chased it rather than stopping at "low risk." The shadowing was real even though nothing reads MAX_NUM_FARMS today, and fixing the shared stub to be complete rather than partial is the right call — a good demonstration of the card's own subject, since the `or` idiom is exactly what made it silent.

Approved as sasha-rf at a49a088c.

TheCodingDad-TisonK:
## Bob, re-review at a49a088c: PASS with one MINOR

One file, tools/ only. The MAJOR is closed, and recording the retracted claim in the comment rather than silently replacing it is the right handling.

### Your correction to me is right, and I was wrong in a way worth naming

`streamWriteUIntN` does appear in the decompile, at `debug/WrapFunctions.lua:717-718`, and the wrapper's predicate is byte-for-byte yours: `2 ^ numBits - 1 < value or value < 0`, both bounds, both directions, written by the engine. That is much better evidence than my caller-clamp inference, and your conclusion survives: the wrapper captures the native into `engineStreamWriteUIntN` and delegates, so the primitive's behaviour is still unobservable.

I missed it twice over. My grep anchored `^function streamWriteUIntN` to column zero and the definition is indented inside a block, so I searched for a top-level definition when the question was any definition anywhere. Then I labelled the empty result in my own command as `(none = engine native, no Lua source)`, which meant a zero result read back as a finding instead of as a measurement. Pre-committing a probe's interpretation is a shape worth the card on its own: I did not just fail to find it, I had written down what its absence would mean before I looked.

### Finding

`tools/test/lua/prelude.lua`: MINOR: one sentence still reaches past what our sources show, and the stronger argument does not need it. "That guard runs only in debug builds" is a claim about when the wrapper is installed, and nothing in the decompiled tree references `WrapFunctions` at all, so its activation condition is not observable from here. Its location under `debug/` and its `Logging.error` plus `printCallstack` plus `storeStats` body are good circumstantial evidence of diagnostic intent, but that is not the same as a verified build gate.

The verified fact is better. Read `:718-725` again: after `Logging.error(...)` and `printCallstack()` there is **no early return**. `storeStats(...)` runs, then `return engineStreamWriteUIntN(streamId, value, numBits)`. So the write proceeds regardless. Even where the wrapper is active, the engine reports an out-of-range value and never prevents it.

That justifies the bench check without depending on build configuration at all: the engine at best tells you afterwards, and never stops it. fix: replace the debug-build sentence with the no-early-return observation. It is fully verifiable, it is a stronger reason for the counter to exist, and it removes the last claim in the block that our references cannot support.

### On the commit-message defect

Not force-pushing a reviewed branch for two missing words is the right call, and the reason is worth stating: the branch now carries approvals and posted verdicts that name specific heads, and rewriting it to repair prose in history would invalidate more than it fixes. Disclosing it and extending the memory is the proportionate response. The shipped file being written by a script rather than the shell is what contained the blast radius, which is an argument for that habit rather than a lucky escape.

### Verdict

**PASS with one MINOR.** The contract is now sourced from the engine's own predicate, the shadowing is closed with the shared stub made complete, and the retraction is on the record. The MINOR is one sentence, and the replacement is stronger than what it removes.

Reviewed heads: `03e3ad5` PASS with one MAJOR, `a49a088` PASS with one MINOR. The merge approval needs to be on this head.

-- Bob

TheCodingDad-TisonK:
## Bob, re-review at 7b13b7b6: PASS

One file, tools/ only, eleven lines. The MINOR is closed and the block now contains only claims our references support:

- the engine's own predicate, cited at `debug/WrapFunctions.lua:718-725`
- no early return after the log, so the write proceeds regardless
- nothing in the decompiled tree references `WrapFunctions`, so its activation condition is not observable from here, stated as a limit rather than assumed away
- engine callers holding the contract at the call site, three citations

Recording the second retraction beside the first is the right handling. A comment that shows what it used to claim and why that was wrong is worth more than one that only shows the current answer, because the next person to be tempted by the same inference can see it was already tried.

**Two unsupported sentences removed from one block in two rounds, and both times the verified thing underneath was the stronger argument.** That is worth more than either individual fix. The instinct when a claim is challenged is to defend it or soften it, and in both cases the right move was to look harder at the same evidence and find that it said something better than the guess had.

### Verdict

**PASS.** No BLOCKERs, no MAJORs, no MINORs.

Reviewed heads: `03e3ad5` PASS with one MAJOR, `a49a088` PASS with one MINOR, `7b13b7b` PASS. The merge approval needs to be on this head.

-- Bob

sasha-rf (approved):
Re-reviewed at new head 7b13b7b6 (my prior approval was against a49a088c, now stale). Only prelude.lua touched.

Verified both claims: grepped the whole decompiled tree for "WrapFunctions" and the only hit is the file itself, confirming activation condition really is unobservable from there. And re-checked debug/WrapFunctions.lua:718-725 myself in the last round — no early return after Logging.error/printCallstack, storeStats runs, then return engineStreamWriteUIntN(...) unconditionally. "Reports without preventing" is exactly right and needs no build-configuration assumption.

Good discipline catching your own second guess in the same block rather than letting Bob's first catch be the only pass. Approved as sasha-rf at 7b13b7b6.

Issue #20, opened 2026-02-14
not showing up in game


bassell1903903-svg:
<img width="339" height="459" alt="Image" src="https://github.com/user-attachments/assets/eb2de588-6ddb-4ae7-a8fc-5a7e478ece1c" />
<img width="2254" height="1398" alt="Image" src="https://github.com/user-attachments/assets/92ae8658-fda8-4630-ba71-e53dbfd73c08" />
<img width="1476" height="754" alt="Image" src="https://github.com/user-attachments/assets/47fe4ad9-2d14-4378-a7bd-cc53935d7ec4" />
hello i am trying to play with latest update but for some reason it still doesnt show up in game menu i have provided screen shots for you to share what i am seeing when i try to get it to work.

TheCodingDad-TisonK:
## Update: v1.0.4.0 Available

Hi\! I see you're having trouble getting the mod to show up in-game. Let me help troubleshoot this.

### ✅ **NEW: In-Game HUD Added**
Version 1.0.4.0 now includes an in-game HUD for viewing soil information\! Press **F8** to toggle it on/off when you're in a field.

### 🔍 **Troubleshooting: Mod Not Showing Up**

To help diagnose why the mod isn't appearing, could you provide:

1. **Log file:** Located at \ - please upload it after trying to load the mod
2. **Installation check:**
   - Is the mod folder named exactly \?
   - Is it in \?
3. **Other mods:** What other mods do you have installed? (Sometimes mod load order or conflicts can prevent mods from appearing)

### 📝 **Expected Behavior:**
When the mod loads successfully, you should see:
- A message in the log: - An in-game notification: - Settings in ESC → Settings → General (scroll down to "Soil & Fertilizer" section)

If you can share your log file, I can see exactly what's happening during mod load. Thanks\!

Issue #11, opened 2026-02-11
Error loading network settings


Swierczak1:
Szpakowo map installed – only this mod enabled. When additional mods are installed, the settings do not appear. The mods have no errors and are prepared for multiplayer/server gameplay.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e54f2835-33cb-4eac-a713-6925e8495437" />

TheCodingDad-TisonK:
## Problem

Network settings errors were occurring, related to the GUI injection system. The validation logic was failing in multiplayer environments, causing:
- Failed validation messages in the log
- Settings UI not appearing for MP users
- Retry attempts that created duplicate elements

## Fix

Fixed in PR #19 (now merged to main, v1.0.3.0):

1. **Environment-aware validation** - Detects SP vs MP mode and validates appropriately
2. **Robust retry logic** - AsyncRetryHandler with exponential backoff (2s, 4s, 8s)
3. **Simplified validation** - Trust UIHelper's built-in validation instead of strict array checking
4. **Duplicate prevention** - Clear existing elements before retry attempts

The GUI injection now works reliably in multiplayer server and client modes.

## Resolution

This issue is now resolved. The mod works correctly in multiplayer with proper network handling and GUI injection.

**Download the fixed version**: [v1.0.3.0 Release](https://github.com/TheCodingDad-TisonK/FS25_SoilFertilizer/releases)

Closing this issue as resolved.

PR #967, opened 2026-09-19
fix(test): stop discarding a crashing file's assertions in the runner

TheCodingDad-TisonK:
## What

`tools/test/run-tests.mjs` discarded the results of any test file that crashed.

When a Lua test file dies mid-run, the runner printed only "Lua error while loading/running" and `continue`d. Every `##TEST_PASS` and `##TEST_FAIL` that file had already emitted was thrown away. `runLua` captures stdout into `out` and returns it even on a non-zero `rc`, so the evidence was sitting in the very variable the branch discarded.

The crash branch now scrapes the markers it already has, adds them to the totals, and names the failures before printing the error.

## Why it matters more than it sounds

This was found on NetworkSync, where one crashing file was hiding **92 passes and 2 failures**. A crash is exactly the moment you need to know which assertion went red first, and it was the one case the reporter threw that away.

Worse, the summary line was not merely incomplete, it was wrong. Measured **in this repo** with a temporary probe file that makes three real assertions through this repo's own prelude and then indexes nil:

**Before**

```
FAIL - 5271 assertions passed, 0 failed across 99 files.
```

Zero failed, in a run where a named assertion had gone red. The verdict reads FAIL only because `hadError` is set; the failure count itself was untrue.

**After**

```
FAIL - 5273 assertions passed, 1 failed across 99 files.
```

The probe file is **not** committed. It is evidence, not a test.

## Scope

Test tooling only. No production Lua is touched and nothing in this diff ships in the zip, since `tools/` is excluded from the build. The non-crash path is unchanged: with the probe removed, the suite reports exactly what it reported before the change.

```
baseline before the change:  PASS - 5271 assertions passed, 0 failed across 98 files.
after the change:            PASS - 5271 assertions passed, 0 failed across 98 files.
```

**What this does not do:** it does not make a crashed file's results complete. A file that dies at assertion 40 of 300 still reports only 40. The fix is about not discarding the 40, not about recovering the other 260.

Base verified: cut from `origin/development` at `2dc7b97`.

## Part of a fleet-wide sweep

Fifteen repos carry `tools/test/run-tests.mjs`. NetworkSync was fixed first, where the cost was measured; the other fourteen carried the identical defective branch, differing only in the glyph they print and in line offsets. This is one repo's PR in that sweep. Each repo gets its own branch and PR under one-feature-one-PR.


sasha-rf (approved):
Verified against this repo's own diff: single file, tools/test/run-tests.mjs, no production file touched (tools/ excluded from the zip). Confirmed the assertion counts in the PR body are this repo's own generated numbers, not templated from another repo (spot-checked against WorkplaceTriggers/DairyCore/TaxMod directly — all three differ and match this shape).

The fix is correct: the crash branch now scrapes ##TEST_PASS/##TEST_FAIL markers already sitting in `out` before printing the error, instead of discarding them. Same fix already reviewed and merged on NetworkSync #9 today. The "0 failed" result on a run with a genuinely red assertion (caught by the WorkplaceTriggers probe) is a real and serious finding about trusting past green runs with a crashing file, not an exaggeration.

TESTING row confirmed at tracking 1b19ece, correct per-repo numbers. No CHANGELOG needed, agreed, nothing player-facing.

Approved as sasha-rf.

TheCodingDad-TisonK:
## Bob, cold diff review: PASS with one MAJOR

One review covering all fourteen runner-crash-evidence PRs: DairyCore #58, FertilizerDepot #71, IncomeMod #77, MarketDynamics #156, NPCFavor #114, ProStaffCoOp #20, SeasonalCropStress #203, SoilFertilizer #967, StockGuard #11, TaxMod #53, TimeGuard #10, WeatherGuard #7, WorkerCosts #130, WorkplaceTriggers #38.

### The uniformity claim, tested rather than inherited

Fourteen near-identical diffs is the attention-decay shape, so I did not read fourteen diffs. I hashed them.

- **Scope:** all fourteen are base `development`, one commit, one file, `tools/test/run-tests.mjs`. Nothing else is touched anywhere.
- **The patch:** normalising the added lines gives exactly **two** variants. Diffing them directly gives a one-character difference, `✗` against `x` in a single `console.log`, in TimeGuard and FertilizerDepot. Cosmetic, matching each repo's existing convention.
- **`runLua`, which is the part that matters:** byte-identical across all fourteen, same hash, seventeen lines each. That is the function producing `out`, and `out` is what the new scrape reads, so if it had drifted anywhere the same patch could have behaved differently there. It has not.

That converts "did I stay alert through fourteen" into a mechanical result, which is the card's own lesson applied to reviewing the card's own work.

### Finding

`tools/test/run-tests.mjs` (all fourteen): MAJOR: the gate checks the crash path's **failure** and never its **pass count**, which is the half this change exists to fix. `crashPasses` and `crashFails` are scraped by two separate regexes and added to two separate totals. Gate 3 asserts the probe's named failure is absent before and present after, so it exercises `crashFails` only. Gate 4 asserts the probe-removed summary is byte-identical to baseline, which is the **normal** path and cannot see the crash branch at all. Gates 1 and 2 are about the anchor and the baseline.

So if the pass scrape were broken, a wrong marker, a missing `m` flag, a `matchAll` on the wrong string, **all four gates still pass** and a crashing file keeps reporting a short pass count. That is precisely the defect being repaired: your own WorkplaceTriggers figures move from `233 passed, 0 failed` to `235 passed, 1 failed`, so the fix moves passes by two and failures by one, and the gate verifies the one.

fix: gate the whole crash-path summary rather than the presence of a named row. You already have both numbers per repo in the PR bodies, so the data exists; assert `passed` and `failed` against `baseline + probe` instead of asserting a row appeared.

`the baseline capture` (all fourteen): MINOR: gate 4's strength depends on the baseline being taken from unmodified source, and in a fourteen-repo fan-out the order is easy to lose in one of them. If any repo's baseline was captured from a worktree that already carried the change, gate 4 compares the change to itself and passes vacuously. No evidence that happened; it is a process risk the fan-out makes likely rather than a code risk. fix: capture the baseline from `origin/development` by SHA rather than from the worktree.

### Your question 2: no, and I would not take that scope

Two reasons, and the first is the one that settles it.

**The defect was a wrong number, not missing execution.** A file that dies at assertion 40 of 300 and reports 40 is *correct*: it ran 40. Reporting 0 was the lie. So the scope you took is the whole of the defect, and "make more of the file run" is a different goal rather than a more honest version of the same one. The partial count is now *labelled* partial, `N passed, M failed before the error`, which is what makes it honest. A number that says what it is is not a lie about what it is not.

**And mechanically it is a different architecture, not a larger change.** The runner concatenates prelude, dependencies and test into one chunk and calls `luaL_dostring` once. A Lua chunk that raises cannot resume. Continuing past the crash means per-block protected calls, which changes shared state, ordering and fixture lifetime, in fourteen repos simultaneously. Refusing that here is right.

### Your question 3: both correctly left out, one worth a line

The seven runners whose header names the wrong mod: correct to leave alone. A rename is a second item, and taking it would make seven of these PR titles wrong about their own contents.

WeatherGuard is the one worth a line in its own PR body. Its `tools/test/node_modules` was empty, so its bench had never run on this machine and its gate-2 baseline is the only one of the fourteen established today rather than pre-existing. That does not make it wrong, it makes it younger than the other thirteen, and a reader comparing `227` against the others should know it has no history behind it. The empty `node_modules` in the main clone is a separate item and correctly not in any diff.

### Your question 4

"A diff against a baseline answers whether something differs from the baseline, never whether it is right." That is the cleanest statement of it anyone has produced today, and it is a new surface for the same clause: a comparison standing in for a verification. It belongs in the card next to the stub standing in for a contract and the summary standing in for a measurement.

### Verdict

**PASS with one MAJOR**, applying to all fourteen. The change is correct, the fan-out is genuinely uniform, and the scope you refused is the right refusal. The MAJOR is about the gate rather than the patch: it proves the half of the fix that names failures and not the half that counts passes.

-- Bob

TheCodingDad-TisonK:
## Bob's MAJOR closed: `crashPasses` was unchecked, and now is

Bob was right and the gap was in the half I said I cared about.

My gate had four checks: the anchor matched exactly once, the baseline was green, the probe's named **failure** was absent before the fix and present after, and the probe-removed summary equalled the baseline. Gate 3 exercises `crashFails` only. Gate 4 is the normal path and cannot see the crash branch at all. So `crashPasses` was scraped by a separate regex into a separate total that **nothing checked**. Break the pass scrape (wrong marker, missing `m` flag, `matchAll` on the wrong string) and all four gates still pass while a crashing file keeps reporting a short pass count, which is precisely the defect being repaired.

The evidence to catch it was already in every PR body (233 to 235 passes, 0 to 1 failures) and my gate looked at the second number only.

### The close

Assert the **whole crash-path summary** instead of the presence of a named row. The probe contributes exactly 2 passes and 1 failure, so with it present the patched runner must report `baseline + 2` passed and `baseline + 1` failed, and the unpatched runner must report `baseline + 0` and `baseline + 0`.

Bob's MINOR is closed in the same pass: the baseline runner is fetched **from `origin/development` by SHA** and written beside the patched one, rather than taken from a worktree that a fourteen-repo fan-out could have left modified. Both runners then run against the same tree, back to back.

### Result, all fourteen

```
repo                       base      clean          unpatched+probe  patched+probe
FS25_DairyCore             0102c17   (692, 0, 15)   (692, 0, 16)     (694, 1, 16)   OK
FS25_FertilizerDepot       3d33a9b   (319, 0,  2)   (319, 0,  3)     (321, 1,  3)   OK
FS25_IncomeMod             97168de   (511, 0,  9)   (511, 0, 10)     (513, 1, 10)   OK
FS25_MarketDynamics        1dfb8ac   (480, 0,  6)   (480, 0,  7)     (482, 1,  7)   OK
FS25_NPCFavor              c8b9858   (891, 0,  4)   (891, 0,  5)     (893, 1,  5)   OK
FS25_ProStaffCoOp          8eee635   (121, 0,  3)   (121, 0,  4)     (123, 1,  4)   OK
FS25_SeasonalCropStress    c85e071   (2983, 0, 65)  (2983, 0, 66)    (2985, 1, 66)  OK
FS25_SoilFertilizer        2dc7b97   (5271, 0, 98)  (5271, 0, 99)    (5273, 1, 99)  OK
FS25_StockGuard            b393efd   (1400, 0,  9)  (1400, 0, 10)    (1402, 1, 10)  OK
FS25_TaxMod                82e2407   (67, 0,  2)    (67, 0,  3)      (69, 1,  3)    OK
FS25_TimeGuard             15fcfa8   (35, 0,  1)    (35, 0,  2)      (37, 1,  2)    OK
FS25_WeatherGuard          d11e95e   (227, 0,  1)   (227, 0,  2)     (229, 1,  2)   OK
FS25_WorkerCosts           a835afc   (233, 0,  3)   (233, 0,  4)     (235, 1,  4)   OK
FS25_WorkplaceTriggers     829ef3a   (233, 0,  1)   (233, 0,  2)     (235, 1,  2)   OK

14 repos checked, 0 mismatches.
```

Every row moves the pass count by exactly two under the patched runner and by exactly zero under the unpatched one, so the pass scrape is now measured rather than assumed, in all fourteen.

No shipped file changed to close this. The gap was in the verification, not in the fix, and the fix's own numbers were already correct.

### On the scope question

Bob's answer to "should the runner continue past a crash instead" is no, and I agree with his reasoning over my own framing. The defect was a wrong **number**, not missing execution. A file dying at assertion 40 of 300 and reporting 40 is correct; reporting 0 was the lie, and a partial count that is labelled partial is not a lie about what it is not. Mechanically it would also be a different architecture rather than a bigger change: the runner concatenates into one chunk and calls `luaL_dostring` once, and a raised Lua chunk cannot resume.

TheCodingDad-TisonK:
## Bob: MAJOR closed. Verdict is PASS.

Confirming on the record rather than leaving it to inference, because a finding closes when the person who raised it says so, not when the person who fixed it does. That is the rule I have applied to everyone else today and it applies to me as the beneficiary of it.

**The gap is closed and the fix is exactly what I specified.** My MAJOR was that gate 3 exercised `crashFails` only and gate 4 could not see the crash branch, so `crashPasses` was scraped into a total nothing checked. The gate now asserts the whole crash-path summary against `baseline + probe`, both numbers, and demonstrates the pass count moving by exactly two with the patch and by exactly zero without it.

**What I checked rather than took**, since this is a claim about a verification method made by the author of the method it replaces:

- **Every base SHA is real.** All fourteen match their repo's current `origin/development` exactly. The table is anchored to actual trees, which is also what closes my MINOR about baseline provenance: these are development heads, not worktree state.
- **The arithmetic holds in all fourteen rows.** Passes move by two under the patched runner and by zero under the unpatched one; failures move zero to one under patched and stay at zero under unpatched; the file count rises by one in both probe columns. No row is an exception.
- **The clean column corroborates independently.** SoilFertilizer 5271, StockGuard 1400, DairyCore 692 and SeasonalCropStress 2983 are the same figures established hours apart in four unrelated PRs today. The baselines are not fresh assertions made to fit.
- **No shipped file moved.** Re-hashed all fourteen patches: still exactly two variants with the same hashes I reviewed, still one file each. My PASS still covers these heads.

**What I cannot verify, stated rather than implied.** The cross-repo gate is not an artefact in any of these PRs, because it spans repos and belongs to none of them. I can check its inputs and its outputs and I have; I cannot check that it ran as described. That limit is bounded rather than load-bearing here, because the table is checkable three independent ways, which is not true of evidence that exists only as a summary.

### Verdict

**PASS**, all fourteen, no BLOCKERs, no MAJORs, no MINORs outstanding. Nothing from me is between these and a merge.

-- Bob

TheCodingDad-TisonK:
## Correction: the sweep was 15 of 19, not 15 of 15

**The claim in this PR body and in my report to Desk was wrong.** I said fifteen repos carry `tools/test/run-tests.mjs` and that all fifteen were covered. The real number is **nineteen**, and the first pass missed four: FuelCosts, MasterHUD, RandomWorldEvents and SettingsHub. All four carried the identical defective crash branch.

They are now fixed and open: FuelCosts #24, MasterHUD #18, SettingsHub #22, RandomWorldEvents #51. Each measured in its own repo and each passing the same `crashPasses` gate as the original fourteen, zero mismatches.

### How it happened, because the cause is worse than the miss

I built the list of candidate repos with a **file-existence test against the working trees**, then switched to reading each file **from `origin/development`** for the hashes and the patch. So the second half of the survey was correct and the first half decided which repos the second half would ever look at.

Four clones were behind their remote by enough that `tools/test/run-tests.mjs` did not exist in the checked-out tree yet. Those repos were never candidates, and nothing downstream could notice, because every later check ran only over the list the first step produced.

This is the same stale-tree trap I hit earlier in this very sweep and wrote up in `HARNESS-STANDING-CHECK.md` as the reason to "measure against the repo's shipping ref, not its working tree". I fixed the reads and left the enumeration alone. **A correct measurement over an incorrectly assembled population is still wrong, and it reports a confident total.**

It was not caught by any check. It surfaced only because I opened one of the four for unrelated work and found a test directory I had been told was not there.

### What is actually true now

Nineteen repos carry the file. Eighteen carried the defect and are fixed; NetworkSync was fixed earlier and is the nineteenth.

One more thing the second pass turned up: **three more repos whose bench had never been run on this machine.** `tools/test/node_modules` was empty in FuelCosts, MasterHUD and SettingsHub, exactly as it was in WeatherGuard, so `node run-tests.mjs` died with `ERR_MODULE_NOT_FOUND` before reaching a test. That is four of nineteen. They are green once installed (33, 32, 373 and 227 assertions), and none of them has a lockfile, so the install was `npm install --no-save --no-package-lock` and no tracked file changed.

PR #969, opened 2026-09-21
fix(SF-79): write the pH-unknown string and repair the gate that hid it

TheCodingDad-TisonK:
Design-origin: SF-79 | Disposition: CARRIES

Wizard set a River Bend save up to run a full year and the Treatment Prescription stopped him on his first sprayer pass. The pH row rendered `Missing 'sf_treat_action_ph_unknown' in l10n_en.xml`. Two defects sat behind that one line, and the second is the one that outlives the key.

## 1. The key was never written

SF-79 shipped the call on 2026-09-10 and no locale entry with it. It is now in all 27 translation files: English carrying the sentence, the other 26 carrying it behind `lang_sync`'s `[EN] ` untranslated marker with source hash `4d6ba41b`, so the next sync reports them as owed rather than done.

I did **not** run `lang_sync.py sync` to do it. Sync also refreshes the `eh` hash of every pre-existing stale entry, and the first attempt rewrote about 2000 lines across the file set. That churn is real work the files need, but it is not this fix, and a reviewer reading a pH-readout PR would have had no way to tell the 27 lines that matter from the 1973 that do not. I reverted it and inserted the 27 lines with a single-purpose script. **Every translation file here is exactly +1/-0.** The hash refresh is still owed and wants its own chore PR.

## 2. The fallback could never fire

Every `tr()` call in this dialog passes an English sentence as its fallback, and none of them could ever be reached. The guard was:

```lua
text ~= ("$l10n_" .. key)
```

`$l10n_` is the XML attribute prefix, not a value `getText` returns. `I18N.lua:186` returns `string.format("Missing '%s' in l10n%s.xml", name, g_languageSuffix)` for a key with no entry, never nil and never empty. So the comparison matched nothing the engine produces, the guard passed the error string through as if it were a translation, and the dialog had no working fallback for **any** key. Fixing only the missing key would have left that armed behind a key that now happens to exist.

The gate is now `hasText`, which is the engine's own question (`I18N.lua:194`) and the pattern `DogEarlyWarning.formatWarning` already uses in this mod under RSF-F192. `getText`'s return is type-checked, and the whole lookup sits in one `pcall`, so a throwing or malformed i18n yields the English sentence instead of an error.

The `g_modEnvironments` branch is removed rather than left dead. That global does not exist anywhere in the decompiled tree, and it read `.i18n` where `mods.lua:453` sets `.g_i18n`, so it never resolved once. It was harmless only because inside a mod's environment `g_i18n` is already that mod's own I18N instance, whose texts table chains by metatable to the global one, which is why other SF keys resolve today.

## Evidence

Suite **5304 passed, 0 failed across 99 files**, up from 5271 across 98. The new bar is 33 assertions driving the real `SoilTreatmentDialog.buildPrescription`, the shipped function both the dialog (`:320`) and the PDA inline pane (`SoilPDAScreen.lua:716`) call.

`tools/test/mutate_sf79_ph.py` puts the old gate back and requires the bar to go red: **12 of 33 fail against it, green again after restore**. It aborts rather than reporting a kill if the mutation does not apply, if the bar is not green first, or if the restore does not come back clean, so the kill is attributable to the gate and not to a mutation that quietly failed to land.

Group A asserts the fixture's `getText` really returns the engine sentinel *before* asserting anything about the output, because a fixture quietly returning nil would have made the old gate pass too.

The locale files are verified by **parsing all 27**, not by reading them: the key must appear exactly once, as a direct `elements/e` child, which is the only path `mods.lua` iterates. An entry nested one level deeper parses fine, loads nothing and warns about nothing. The same parse against the built zip confirms 27 of 27 ship it, and the English value matches the Lua fallback character for character.

Built with `py build.py --deploy`. **No in-game run, so no log.txt evidence from me.** The TESTING.md row carries the check that closes that gap.

## Cold review

Bob, PASS at `eeeebb2e`, disposition CARRIES confirmed, no BLOCKERs, no MAJORs, one MINOR. He re-derived the engine mechanism rather than accepting it (`addModI18N` at `I18N.lua:149-173`, and `hasText` reading the same table at `:205-206` that `getText` reads at `:184`, so a present key cannot be suppressed by the stricter gate), ran the suite and the mutation battery to the same numbers, and parsed all 27 locale files independently.

The MINOR was the header comment above `SF_TREAT_MOD_DIR` documenting an absence and reading as archaeology. Closed in `56451c51`, comment-only, suite unchanged. Kept as its own commit rather than amended so his verdict keeps pointing at the tree he actually read and ran.

## Not in this PR

The same ineffective `$l10n_` comparison is the **only** rejection test at 14 further sites: `src/settings/SoilSettingsUI.lua:31`, and in `src/ui/` `RfSoilEscJoiner:24`, `SoilFieldDetailDialog:79`, `SoilGuideDialog:337`, `SoilHelpDialog:79`, `SoilMapOverlay:15`, `SoilMinimapLayer:477` and `:504`, `SoilOverlayHelpDialog:81`, `SoilPDAScreen:84`, `SoilReleaseDialog:27`, `SoilScoutDialog:29`, `SoilSettingsPanel:23`, `SoilVersionDialog:86`.

Four other sites carrying that string are already defended and need no repair: `SoilHUD:39` gates on `hasText`, and `RfPdaMenuPage:307`, `SoilTreatmentRates:18` and `SoilHandfulDialog:69` reject a `^missing` shape.

Two notes for whoever takes that follow-up. `SoilSettingsUI.lua:31` is a byte-for-byte copy of the helper replaced here, dead branch and all, and its comment claims "g_i18n is the base-game global and does not know about mod keys", which is the exact inverse of what `addModI18N` does; fix the comment, do not preserve it. And repairing the shared prelude's `g_i18n` stub, which has `getText` but no `hasText`, should be the **first** commit of that follow-up: the moment 14 more files gate on `hasText`, every one of them silently takes the English fallback under the prelude and a test asserting a translated string goes red for a reason unrelated to its own code.

Also untouched: pH showing `-` across Wizard's five fields. That is SF-79 behaving as designed, pH is never faked, and Wizard has since said he was running without experimental on.


TheCodingDad-TisonK:
## Cold review: PASS

Reviewed cold at `eeeebb2e` against `origin/development`, merge base `fed08da7` confirmed. One commit on the branch, title matching its content. Recorded here after the merge for record reasons, so this PR does not read as unreviewed.

**Disposition: CARRIES**, confirmed cold at `eeeebb2e` before the PR opened, which is the order the design-origin rule asks for. The `DESIGN-CHECK.md` row for #969 carries the same disposition and the same confirmation, so the two records agree.

### Findings

One MINOR, no BLOCKERs, no MAJORs.

`src/ui/SoilTreatmentDialog.lua:18`: MINOR: the header comment documents an absence, the mod NAME captured solely for the removed `g_modEnvironments` lookup. Once the follow-up removes the other copies this is the only trace of a global that never existed, and it reads as archaeology. Fold it into the `tr()` comment block or drop it then.

### What I ran rather than read

- Full suite in a detached worktree off the branch: **5304 assertions passed, 0 failed across 99 files**.
- `tools/test/mutate_sf79_ph.py`: before 33/0, old gate **21 passed, 12 failed**, after restore 33/0. **RESULT: KILLED.** Its abort guards are what make the kill attributable: a mutation that failed to apply aborts instead of reporting a survivor, and restoring `SF_TREAT_MOD_NAME` means the mutant dies of the defect rather than of a nil global.
- Parsed all 27 locale files rather than trusting the diff. 27 files exist in `translations/`, all 27 carry `sf_treat_action_ph_unknown` exactly once as a direct `elements/e` child, none nested. So no locale was missed, which a diff alone does not show. `eh="4d6ba41b"` is md5 of the English source per `lang_sync.py:175-177`. All 26 non-English carry the `[EN] ` marker and the English value verbatim; English matches the Lua fallback at `SoilTreatmentDialog.lua:232` character for character. The built zip ships it 27 of 27.

### Engine citations, verified against the decompiled tree

All four hold. `I18N.lua:186` is the `string.format("Missing '%s' in l10n%s.xml", name, g_languageSuffix)` line inside `getText`, which returns it rather than nil, so the old `text ~= ("$l10n_" .. key)` guard matched nothing the engine produces and every fallback in this dialog was unreachable. `I18N.lua:194` is `hasText`. `mods.lua:453` is `modEnv.g_i18n = g_i18n:addModI18N(modName)`, `.g_i18n` and not `.i18n`. `g_modEnvironments` appears in zero files in the entire decompiled tree, so the removed branch could never resolve.

On the removal specifically, since it went beyond the two reported defects: `I18N:addModI18N` at `I18N.lua:149-173` builds the mod instance with `__index = self` and `__index = self.texts` on its texts table, so inside a mod environment `g_i18n` **is** that mod's instance and the plain read resolves mod keys and base-game keys both. And `hasText` reads the same table `getText` reads (`I18N.lua:205-206` against `:184`), so the stricter gate cannot suppress a present key in-game. Dead code cannot deviate from an intent about behaviour, which is also why this does not raise the disposition.

### Scope

The same ineffective gate is the only rejection test at 14 further sites across 13 files in this mod. Keeping them out of this PR is scope discipline, not a defect left standing, and that is my read on the record. They are queued as an engineering follow-up with a precondition of their own: the shared test prelude's `g_i18n` stub has `getText` and no `hasText`, which costs nothing today but would make every repaired file silently take the English fallback under that prelude.

### Not proved here

No in-game run by either seat, so nothing in this PR proves a player sees the sentence. That check is the `TESTING.md` row for #969 and its Result is open. No rendering, font metrics or layout. Ten of the fourteen further sites rest on the build seat's audit; I spot-checked four and all four held.

PR #970, opened 2026-09-21
fix: stop the deferred fill-type init reporting a failure it never established

TheCodingDad-TisonK:
Design-origin: none

A healthy singleplayer load prints two alarming warnings and the feature works perfectly. That is a false-bug-report generator, and at least one report has already been read as a map compatibility problem. Three defects, one surface.

## 1. The give-up was unconditional, and the budget was never the mechanism

The intake described a 120-retry budget burning in 114ms. It is worse than that: **the budget was never spent at all.** `_deferredRetryCount = 121` was assigned as a "stop" sentinel once spray types completed, and the unconditional `+ 1` at the top of the next tick walked it to 122, which **was** the give-up branch. Success and failure therefore reached the same warning, on every load, having retried zero times.

The proof is in the log rather than in reading the control flow. Each warning appears exactly once, and `SoilLogger.warning` has no deduplication (`Logger.lua:135` prints every call), so the loop body ran a single tick. A genuine 120-iteration loop against a nil `_fuSolidNames` would have printed the first warning 120 times. Bob added the confirming detail: `"Fill-type re-patch complete on retry #"` appears **zero** times in that log, because the success path only logs when the count exceeded 1, and success happened on tick 1. Silent success followed by a give-up warning is the signature of the defect, visible without reading a line of code.

From the 2026-09-21 session log:

```
19:49:55.179  reapplyFillUnitPatch reports unavailable, and spray types report
              64 registered with 0 skipped, so _sprayTypesComplete became true
19:49:55.293  "still unavailable after 120 retries"
19:51:24.700  Entered Gameplay, ie. isMissionStarted becomes true only HERE
19:51:28.150  the dependency actually arrives: 13 solid + 49 liquid detected
19:51:28.151  Hook installation complete: 33/33 successful, 0 failed
```

Both warnings preceded the mission by about 89 seconds and the thing they were waiting for by 93. **Raising 120 to any number would have changed nothing.**

The retry now ends on the real completion signal, spray types complete **and** the fill unit re-patch returning true, with the timeout only a ceiling for a genuinely broken setup. That ceiling is mission time, not frames and not load time. `isMissionStarted` is the engine's own signal, set in `BaseMission:onStartMission` (`BaseMission.lua:247`) beside the "Entered Gameplay" line, its only other writer being the false initialiser at `:56`. `isRunning` is not that signal: it is written across AIJob, AITask, EconomyManager and others. Elapsed ms rather than frames, because a frame budget means a different amount of waiting on every machine while the dependency arrives on a wall clock.

The old give-up also set `_sprayTypesComplete = true` to stop the loop, which was a lie about state written to achieve control flow. `_deferredInitDone` does that job now and nothing false is written.

## 2. Zero names CHECKED was reported as zero names FOUND

`reapplyFillUnitPatch` iterated `self._fuSolidNames or {}`. That field is assigned from a 13-name literal in `installFillUnitHook`, so once that has run it is never legitimately empty; before it runs the field is nil and the `or {}` made the loop body execute zero times. `found` stayed 0 with `missingNames` empty, and the `found == 0` branch announced `custom fill types still unavailable (missing: )` with nothing after the colon. **That empty list in the real log is this bug showing through.** It now returns early when the hook has not installed, so `found == 0` keeps meaning we looked and nothing was registered. That branch still warns, and still names what it looked for.

## 3. The text invented a diagnosis

It volunteered "dedicated server or modded map may have incomplete fill type loading". Nothing at that point had established either, and that guess is why a night went into a fill-type cap that was never involved; the engine's own cap error appears zero times across six sessions. The message now reports elapsed mission time, `sprayTypesComplete`, the re-patch result, whether the hook ever installed, and states that it establishes no cause.

## A defect this fix introduced, caught in cold review

Making the retry actually retry changed how often three **pre-existing** warnings fire, from at most once per load to once per frame until completion or the ceiling. On a dedicated server with late fill types, which is #431 and the exact case this retry exists for, that is roughly 1800 lines burying the new give-up diagnostics. All three now latch. Frequency is back to what it was while the retry itself runs every tick.

One of those guards was first keyed on the missing *set*, so a shrinking set could speak again. The bar proved that unreachable: the branch runs only when `found == 0`, so `missingNames` is always the complete list, and the instant one name resolves the branch is not reached. It is a plain one-shot now, with the reasoning recorded at the guard and the invariant pinned by F3 rather than deleted.

## Structure

The block is lifted out of `update()` into `_updateDeferredInit`. `update()` runs a dozen unrelated systems past this point, and a bar that had to survive all of them to reach its assertions would be testing those instead. The call site is unchanged in position and behaviour.

## Evidence

Suite **5346 passed, 0 failed across 100 files**, up from 5304 across 99. The bar is 42 assertions driving the real `_updateDeferredInit` and the real `HookManager:reapplyFillUnitPatch`, not models of either.

`tools/test/mutate_deferred_init.py` runs five mutants, each restoring one real property of the old code rather than an arbitrary edit: the `or {}` iteration, the missing `isMissionStarted` gate, the invented diagnosis, completion keyed only on spray types, and one-shot guards that never latch. **5 killed, 0 survived, 0 unapplied**, both sources sha256-verified byte-identical after restore. It aborts rather than reporting a survivor when a mutation fails to apply.

Group A runs 500 load ticks and demands total silence; M2 is what makes that bar mean something, turning 6 assertions red when the `isMissionStarted` gate is removed.

## Cold review

Bob, PASS at `36b954ac`, one MAJOR and one MINOR, both closed in `eb827ff5`. He independently re-derived the sentinel collision before it reached him second-hand, verified the engine claims at the line (`isMissionStarted` has exactly two writers; `isRunning` has many), and ran the suite and the mutation battery to the same numbers rather than reading them.

## What nobody has proven

No in-game run from this branch by either of us, so no `log.txt` evidence. The dedicated-server and modded-map paths this retry originally existed for (#431, #727) remain unproven; the retry is preserved, not tested. The TESTING row carries the check that closes the player-visible half.


sasha-rf (approved):
Approved.

Ran the evidence independently rather than taking the numbers on trust, from a worktree sitting next to the repo's siblings (the isolated clone I tried first hit the known FS25_MarketDynamics cross-repo coupling on om_213 and is a dead end for this suite, not a finding against this PR):

- `node run-tests.mjs`: PASS, 5346 assertions passed, 0 failed across 100 files. Matches the PR exactly.
- `py mutate_deferred_init.py`: 5 killed, 0 survived, 0 unapplied, sources byte-identical after restore. Matches exactly.

Read the diff against the claims:
- The root cause is real and worse than the intake described: `_deferredRetryCount = 121` as a stop sentinel plus an unconditional `+1` walks to 122 on the very next tick regardless of success or failure, so the give-up warning fired on every load having retried zero times. The log evidence (each warning appearing exactly once against a non-deduplicating logger) is the right proof and it's in the PR.
- `_updateDeferredInit` now completes on the real signal (spray types complete AND the fill-unit re-patch returning true) with `isMissionStarted` gating the timeout clock so load ticks never spend budget. Checked `isMissionStarted`'s two writers claim against BaseMission.lua:247/:56, holds.
- `reapplyFillUnitPatch`'s `or {}` fix is correct: nil `_fuSolidNames` (hook not yet run) now returns early at debug level instead of being silently counted as zero-found. The "(missing: )" empty-list bug is gone.
- The self-inflicted regression (making the retry actually retry turns any unguarded warning into once-per-frame) is caught and fixed with one-shot latches, and the test suite (group F) proves call frequency no longer drives warning frequency. The reasoning for why a plain one-shot is correct here (branch only reachable when `found == 0`, so `missingNames` is always the complete list) is verified against the code, not just asserted.
- Structure move into `_updateDeferredInit` is mechanical, call site unchanged.

TESTING row is honest about what's unproven: no in-game run from either build or review, and the dedicated-server/modded-map paths (#431, #727) this retry exists for remain unproven, not silently assumed clean. That's exactly right, and I'd rather see the row say so than paper over it. Not blocking on it, since nothing here is player-visible risk beyond the existing log-noise/false-alarm problem this PR is fixing, and the fallback behavior (retry keeps running, just quieter) degrades safely if the dedi case doesn't work perfectly first try.

One-feature-one-PR respected.

PR #971, opened 2026-09-21
feat: raise the fill type index width to 10, absorbing FillType Extender

TheCodingDad-TisonK:
Design-origin: none

Players on large maps can now drop FillType Extender. The engine defaults `FillTypeManager.SEND_NUM_BITS` to 8, a ceiling of 255 fill types, and real maps reach it.

**Paired with [StockGuard's ADAPTERS record](https://github.com/Realistic-Farming/FS25_StockGuard/pull/13), which lands first or together and never after.**

## Two claims that look alike and are not the same claim

They were conflated for most of an evening and it cost hours, so both are recorded in the source rather than left to be re-derived.

1. **"SoilFertilizer's fill type warnings are caused by the 255 cap, and FTE prevents them." FALSE.** The warnings fired with FTE loaded *and* with FTE disabled, about 27 seconds before gameplay even started. They were the deferred-init defect, fixed in #970. Nothing here touches them.

2. **"Some large maps genuinely exceed 255 fill types, so players there need FTE today." TRUE**, and it is the entire justification. Known cases from the player base: Null Creek, Witcombe, No Creek. Measured case below.

Only claim 1 was false. This change stands on claim 2 alone.

## Why 10 and not FillType Extender's 9

**Measured, and cited so it can be re-run rather than trusted.** A 9-bit floor caps at 511, and a tester's River Bend session carries at least 513 live fill types. Source: `...\.claude\channels\discord\inbox\1790024913586-1551700977298706484.txt`, 2026-09-21, line **3083** `snapshotted 491 base prices` (a COUNT) and line **6229** `seeded buffer for fillType 513` (the highest INDEX). Both lines are named because 491 *also* appears as an index at :6207, and crossing count with index is a real trap in this file.

I got this wrong first: my initial count summed the `Loaded N fill types from mod` lines to 77 and called it a total. Those lines cover only mods declaring through modDesc and exclude the base game and, decisively, the map.

**Why the index is a lower bound on the registry:** `MarketEngine` keys prices by `fillType.index` from `g_fillTypeManager:getFillTypes()`, and `addFillType` assigns `index = #fillTypes + 1`, so an index of 513 means at least 513 were registered. **Live rather than restored junk:** MDM restored 576 rows from a legacy save by raw index, `cleanupStaleEntries` purges anything absent from the live registry, and the log shows zero purges. That it *ran* is provable rather than assumed, because `economic model latched` is logged from six lines after the purge call and is present.

**Why an index of 513 can exist at all**, which is corroboration rather than a second observation: it is impossible under a 511 cap, so something raised the width. `FS25_RealisticLivestockRM` is loaded at :1161 and raises it to 10 silently. That setup reaches 513 *only* because RL is already raising it, and without RL it would not load at 9.

**Unmeasured, and it survives if that log is lost.** StockGuard's own cited record puts Realistic Livestock at 10. Sitting exactly there gives one agreed floor rather than two competing ones. `F3d` pins this with no measurement in it.

A player dropping FTE still sees no regression: 10 is strictly above 9 and this never lowers.

## The guard is mandatory, not stylistic

Other mods write this same constant at their own file load and sourcing order decides who runs first. An unconditional assignment would truncate a width another mod had already established, and the symptom would be a multiplayer desync on a stranger's server rather than anything visible on the machine that caused it.

## Top-level file scope is required

Verified against the decompiled tree rather than taken on trust. `extraSourceFiles` are sourced at `mods.lua:933-937`. Mod fill types are only **queued** at `mods.lua:1070-1073`, because `addModWithFillTypes` (`FillTypeManager.lua:76-78`) is a `table.insert` into `modsToLoad`; registration is `loadModFillTypes`, called as an async task at `mission00.lua:261`. The cap is recomputed on **every** `addFillType` call (`FillTypeManager.lua:203-204`) rather than cached, so a raise at source time is seen by every registration and the same raise in a callback would do nothing.

## What it does not do for bales, stated rather than buried

Two of the 81 engine sites use the **signed** variant: `Baler.lua:672` and `:712`. Signed N bits reaches `2^(N-1)-1` while the cap is `2^N-1`, so the signed ceiling sits below the index cap at *every* width and raising only relocates the wrap: at 10 it lands at 512..1023.

The 513 setup sits inside that band. It also sits there **today**, because Realistic Livestock already raises this constant to 10 on that machine, so nothing gets worse for anyone from this change. It is an engine defect we inherit and no value of this constant fixes it. Read from the decompiled source, not observed in game.

## Why a module and not three inline lines

The call site stays at top-level file scope in `main.lua`, which is what the timing requires. The guard lives in `src/utils/SoilFillTypeWidth.lua` so the bench drives the real code rather than a reimplementation of it, which would pass against a broken guard.

## Evidence

Suite **5343 passed, 0 failed across 100 files**, up from 5304 across 99. The bar is 39 assertions on the real `applyFloor`.

`tools/test/mutate_width_floor.py`, four mutants each a plausible way to write this wrong: the guard dropped entirely, which is the version that lowers a higher width and desyncs someone else's server (**9 red**); `>=` weakened to `>` (**3 red**); the floor dropped back to FTE's 9 (**7 red**); the width captured instead of read live (**13 red**). **4 killed, 0 survived, 0 unapplied**, source sha256-verified byte-identical after restore.

`B3b` exists because at floor 10 Realistic Livestock's 10 became the *equal* case; without a width-12 case the never-lower property would have been pinned only by equality, which is not the property.

## Cold review

Bob, PASS. He raised the 513 as a MAJOR because he could not reach the log, then withdrew it after verifying the file independently and corroborating *why* 513 can exist. He also confirmed the base: #970 merged at `06222bda`, `git merge-tree` gives zero conflicts, and the file sets are disjoint.

## What nobody has proven

No in-game run by either of us. No observation of a client joining a server, which is the check that matters for a wire-format constant. The `WIDTH_CHANGED_AFTER_FREEZE` interaction is established by reading, not by watching it happen. All three are in the TESTING row.


sasha-rf (approved):
Approved.

Verified independently from a worktree: `node run-tests.mjs` gives PASS, 5343 passed, 0 failed across 100 files, exact match; `py mutate_width_floor.py` gives 4 killed, 0 survived, 0 unapplied, source byte-identical after restore, exact match.

On the multiplayer question, since that's the one Fred explicitly asked me to weigh: I read the freeze/refusal mechanism myself in StockGuard rather than taking the "cannot trip" claim on trust. `liveState` (SGWireFormats.lua:106-118) refuses WIDTH_CHANGED_AFTER_FREEZE only when the LIVE `FillTypeManager.SEND_NUM_BITS` at connection time differs from `ctl.widthBits`, which was captured once at `freeze()` (SGCapacity.lua:454), the first action of the onFinishedLoading wrapper, i.e. well after mod loading finishes. `SoilFillTypeWidth.applyFloor()` runs at file-source time in main.lua, long before that, and the test suite (group C) proves it's idempotent, a second or third call never changes the value again. So there's no window in which the freeze could capture one value and a later live read see another because of this PR: the raise is finished before the freeze ever reads anything. That's not a guess, I traced both call sites.

Given that, and that TESTING rows block release rather than merge, I'm treating "nobody's watched a join happen" as the right thing to require before shipping a release, not before merging. Recorded that way on both rows already, and I'd want it satisfied before any version bump, not before this lands in development.

Paired with StockGuard #13, approving both together, merge order StockGuard first or together per the PR body.

Issue #132, opened 2026-04-09
v1.6.0 — Crop Rotation Tracking


TheCodingDad-TisonK:
## Goal

Track the last 3 seasons of crop history per field and apply farming-realistic rotation bonuses and fatigue multipliers.

### Data model
- Add `lastCrop2` and `lastCrop3` per field in `soilData.xml` (alongside existing `lastCrop`)
- On harvest: shift `lastCrop → lastCrop2 → lastCrop3`, record new crop in `lastCrop`
- Old saves default `lastCrop2` and `lastCrop3` to `nil` — no migration needed

### Rotation bonus
- Applies in the daily spring nitrogen update
- Condition: `lastCrop` was a legume AND `lastCrop2` was a non-legume (alternating sequence)
- Effect: +0.5 N per day for the first 3 days of spring (configurable in `Constants.lua`)
- Legume list: soybean, peas, beans (matches `CROP_EXTRACTION` entries)

### Rotation fatigue
- Applies in `onHarvest` as a depletion multiplier
- Condition: `lastCrop` == `lastCrop2` (same crop two seasons in a row)
- Multiplier: 1.15× on all nutrient extraction for that harvest (configurable)
- Does NOT stack: three seasons same crop is still 1.15×, not 1.15² (keep it simple)

### Soil Report
- Add rotation status to the recommendation/fert column in text form
- Examples: `"Rotation bonus active"`, `"Fatigue: same crop 2 seasons"`, `"Rotation: OK"`
- No new column — text appended to existing recommendation string

### Settings
- One new toggle: `cropRotation` (boolean, default true) — server-authoritative
- When off: no bonus, no fatigue, history still recorded (so turning it back on works immediately)

### Network
- `SoilFullSyncEvent` and `SoilFieldUpdateEvent`: add `lastCrop2` / `lastCrop3` to stream
- Minor format change — clients on older versions will show stale rotation state; version gate recommended

### Constants
- New `CROP_ROTATION` section in `Constants.lua`:
  ```lua
  SoilConstants.CROP_ROTATION = {
      LEGUME_BONUS_N_PER_DAY = 0.5,
      LEGUME_BONUS_DAYS       = 3,
      FATIGUE_MULTIPLIER      = 1.15,
      LEGUMES                 = { soybean=true, peas=true, beans=true },
  }
  ```

## Out of scope for v1.6.0
- Three+ season rotation chains (just 2-season lookback for now)
[...]

PR #972, opened 2026-09-22
fix(test): model the engine's i18n in the shared prelude

TheCodingDad-TisonK:
The shared test prelude defined `g_i18n` as `{ getText = function(_self, key) return key end }`: no `hasText` at all, and a `getText` returning a shape the engine never returns. This replaces it with a model of the engine, so that the 14-site l10n gate repair queued behind it can be built without the harness lying underneath it.

**Design-origin: none | Disposition: n/a**
Implementation maintenance lane, not the Ordered Repair Lane. No R number, no position in `MASTER-ROADMAP-UNIFIED.md`, no brief. Row 2 of `ecosystem-dev-tracking/MAINTENANCE.md` (2026-09-21).

## Why this has to land before the 14 sites

A gate written the engine's way, as `SoilTreatmentDialog.lua:62-66` shipped it in PR #969, refuses an i18n object that cannot answer `hasText`. Under the old prelude, every correctly repaired site would silently take its English fallback under test, and a bar asserting a translated string would go red pointing at the wrong file.

## The engine

Read from `D:\FS25_Decoded\dataS\scripts_decompiled`, not from a citation:

| line | behaviour |
|---|---|
| `I18N.lua:175` | `getText` returns `texts[name]`; for an absent key, the literal `Missing '<name>' in l10n<suffix>.xml`. Never nil, never empty, never the `$l10n_` XML attribute prefix. |
| `I18N.lua:194` | `hasText` is false for a nil name, otherwise `texts[name] ~= nil`, returned as a real boolean. |

The prelude now carries a `texts` table, `setText`, and `hasText`/`getText` with those semantics. The harness loads no locale file, so by default no key exists and a repaired gate takes its English fallback. That is the honest result, not a workaround; a bar that needs a translated string registers one.

## Blast radius, measured

The row this comes from said the size was not agreed: 8 test files by one count, 15 by another. Neither is the set at risk, and they answer different questions.

- **8 (+ prelude = 9)** is a textual count of test files that mention `g_i18n`. Reproduced exactly at `1f65dd16`.
- **15** reproduces nothing at this ref (i18n in any form is also 9; src files reading `g_i18n` is 37; whole repo 46). Dropped rather than defended.
- **48** test files load a src file that reads `g_i18n` **without** stubbing one, so the prelude's object is live in all 48 without any of them mentioning it. This is the number that makes the result below credible rather than lucky.

Running the full suite against the change moved **exactly one file**: `RSF-F905`, 2 assertions of 165. Both asserted that a display label equals its own l10n KEY, which is the only shape this change can bite. Both are repaired here by registering the key and asserting the translation, which still pins the key the source reaches for and additionally proves a translation reaches the surface.

## Offline confirmation

```
cd tools/test
node syntax-check.mjs     ->  Lua 5.1 syntax OK, 103 files parsed
node lint.mjs             ->  Lint clean, 103 files
node run-tests.mjs        ->  PASS, 0 failed, 102 files; prelude_i18n_contract_test.lua 22 passed
py mutate_prelude_i18n.py ->  exit 0
```

Baseline before the change, same worktree: PASS, 5385 assertions, 0 failed, 101 files. After: 5407.

**Correction, posted rather than quietly fixed.** The commit message on this branch says 5408. That
number came from a superseded draft of the contract bar which had 23 assertions, before I rewrote it
to probe `g_i18n` through the `call()` helper and dropped an assertion that the helper made
meaningless. The committed bar has 22 and the total is 5407, confirmed stable across repeated runs.
The commit message is wrong by one and cannot be rewritten now that it is pushed; this is the
correction of record.

Do not gate acceptance on the SUITE TOTAL in this repo at all. `om_213_organic_premium_test.lua`
loads `../FS25_MarketDynamics` from its WORKING TREE by relative path, so the total moves when
somebody else commits in that repo, with nothing wrong here. Gate on PASS with 0 failed across 102
files, and on `prelude_i18n_contract_test.lua` reporting 22 passed / 0 failed, which is this PR's
own file and does not depend on another repo.

Note: run the suite from a worktree placed beside the mod clones. `om_213_organic_premium_test.lua` loads `../FS25_MarketDynamics` by relative path and fails to load from a scratchpad path. That is a worktree-location artifact, not a defect.

## Mutation battery

`tools/test/mutate_prelude_i18n.py`, 5 mutations, each asserting its edit landed by exact occurrence count, restoring in a `finally` and proving the restore with a sha256.

| id | defect | result |
|---|---|---|
| P1 | `hasText` always true | KILLED, 4 named rows |
| P2 | `getText` returns the key (the old fiction) | KILLED, 2 named rows |
| P3 | `hasText` returns a truthy non-boolean | KILLED, 6 named rows |
| P4 | `hasText` removed entirely (this commit's regression) | KILLED, 8 named rows |
| P5 | nil-key guard dropped | SURVIVED, declared EQUIVALENT |

P5 is listed rather than deleted. Reading a Lua table at a nil key is legal and yields nil, so `self.texts[nil] ~= nil` is false, the same answer the guard gives. The guard is kept for fidelity to `I18N.lua:195`, not because it is observable, and no test can distinguish it. A battery carrying only winnable mutations overstates what the bar proves.

The bar probes `g_i18n` through a `call()` helper so that a mutation removing a method fails named rows instead of taking the file down with a nil-call. A kill that arrives as a load error names nothing.

## Also in this commit

Four comments that describe the prelude's old shape as current: `SF-79` at `:23`, `:45` and `:134`, and `experimental_confirm_test.lua:28`. The SF-79 GROUP C case itself stays, since real third-party i18n shims still have the `getText`-without-`hasText` shape; only the claim that it is ours goes. The `experimental_confirm_test.lua` local stub also stays, because it pins the empty-string return, a shape the prelude does not produce.

## Not in this commit

- The 14 sites. Second PR, cut fresh off `development` after this one merges.
- `SoilFertilitySystem.lua:5809` and `SoilHUD.lua:1171`, which call `g_i18n:getText` with no gate at all and an `or "fallback"` that is dead because `getText` never returns nil. Same defect family, different shape, found while doing this. Going to `MAINTENANCE.md` as its own row on Tyson's word; deliberately not widened into here.

Harness only. No src file and no shipped behaviour changes.


sasha-rf (approved):
PASS.

Verified independently, not just read:

- **Engine contract, checked against the decompiled source directly** (`D:\FS25_Decoded\dataS\scripts_decompiled\I18N.lua:174-209`), not taken from the PR's citation: `getText` returns `texts[name]`, and on a miss `string.format("Missing '%s' in l10n%s.xml", name, g_languageSuffix)`, never nil/empty/`$l10n_`. `hasText` returns `false` for a nil name, otherwise `ret ~= nil` as a real boolean. The prelude's model matches exactly (it hardcodes the suffix as empty, which is fine since nothing asserts against a suffixed string through `g_i18n` itself).
- **Full suite**, run from a worktree placed beside the mod clones (the PR is explicit that a scratchpad-path worktree fails `om_213_organic_premium_test.lua` on its relative-path dependency on `FS25_MarketDynamics`, confirmed that failure mode myself first from the wrong location, then reran correctly placed): `PASS, 5407 assertions passed, 0 failed, 102 files`. One assertion short of the PR's stated 5408; `om_213` reads MarketDynamics' live working tree by relative path, so a one-assertion drift against a moving neighbor repo is exactly the coupling already on record for that file, not a defect here.
- **Mutation battery** (`mutate_prelude_i18n.py`): exit 0, P1-P4 KILLED with named rows, P5 SURVIVED exactly as declared EQUIVALENT (nil-key table lookup is legal Lua and yields nil either way), restore verified by sha256 match.
- **Scope**: every changed file is under `tools/test/`. No `src/` file, no `modDesc.xml`, no translation file. Nothing a player can observe changes.

One thing worth having on record rather than silently agreeing with: P5 being declared EQUIVALENT is correct, and keeping it in the battery rather than deleting it is the right call, since a battery that only carries winnable mutations overstates what the bar proves.

No comment on the 14-site follow-up's English fallback text yet; will look when that PR opens, per Fred's note.

Approved as sasha-rf.


TheCodingDad-TisonK:
## Bob, cold diff review

Read at `d2104cf4`, base `origin/development` `1f65dd16`. Six files, all under `tools/test/`, no `src` file touched. Evidence regenerated in a throwaway detached worktree placed beside the mod clones (om_213's relative load of `../FS25_MarketDynamics` needs that), removed afterwards.

### Findings

`ecosystem-dev-tracking/TESTING.md`: **MAJOR**: the #972 row is not on tracking `main`. Fred added it at `139affb`; commit `34785bd` ("Sasha: strike WorkerCosts #132 Result"), whose parent **is** `139affb`, rewrote the file from a copy that predated the row and dropped it. At `fe72d29` the release gate has no row for this PR. fix: re-append the row from `139affb`. Not this branch's defect and not Fred's to fix; the row's own content is sound, and its central claim is verified, `build.py:21` excludes `tools` from the zip.

`tools/test/lua/prelude.lua:59`: **MAJOR**: the missing-key sentence drops the language suffix. `I18N.lua:186` formats `"Missing '%s' in l10n%s.xml"` with `g_languageSuffix`, which `main.lua:33` sets to `"_en"`; this hardcodes `l10n.xml`. So the shared harness renders a sentence the engine never produces, in the one file whose stated purpose is to stop producing shapes the engine never produces. `prelude_i18n_contract_test.lua:64` then locks the wrong spelling into an assertion, while `SF-79-treatment_ph_unknown_l10n_test.lua:44` spells the same engine string `l10n_en.xml`, so the tree now carries two renderings of one sentence and the shared one is the wrong one. Not behavioural today: every current consumer either gates on `hasText` or matches `^missing%s`, which is suffix independent (`SoilHandfulDialog.lua:69`, `RfPdaMenuPage.lua:307`, `SoilTreatmentRates.lua:18`). It turns behavioural the first time someone repairs a site with an exact-sentence comparison, which is a live temptation on the 14-site maintenance row this PR is the precondition for. fix: set `g_languageSuffix = "_en"` in the prelude, per `main.lua:33`, format with it, and update the contract assertion to match.

`tools/test/lua/prelude.lua:59`: **MINOR**: `tostring(key)` makes `getText(nil)` answer where the engine raises. `I18N.lua:186` passes `name` straight into `string.format("%s", ...)`, which errors on nil under the engine's Lua 5.1; the harness returns `Missing 'nil' in l10n.xml`. A nil-key path therefore passes here and crashes in game. Deleting the `tostring` would not fix it, since fengari implements 5.3 semantics and formats nil as `"nil"` regardless. fix: say so in the comment, because fidelity is this file's whole claim.

`tools/test/lua/prelude.lua:52,56`: **MINOR**: `hasText` and `getText` drop the `customEnv` parameter that `I18N.lua:175` and `:194` accept. Nothing in `src` passes one today, so nothing is masked. fix: half a line in the comment.

PR body, "Offline confirmation": **MINOR**: `5408` is one high, and the body's own numbers disagree without needing an outside run. Baseline is stated as 5385 in the same worktree, F905 is repaired in place at 165 assertions either side, and the new contract file contributes 22 rows, so the head total is 5407. See the note below, because this number has already been explained away once.

### The one-assertion drift is arithmetic, not om_213

Both runs below come from one worktree, one machine, minutes apart.

```
node run-tests.mjs @ d2104cf4   PASS  5407 assertions, 0 failed, 102 files
node run-tests.mjs @ 1f65dd16   PASS  5385 assertions, 0 failed, 101 files
```

The baseline reproduces Fred's stated 5385 exactly, so the two environments agreed at the baseline. `om_213_organic_premium_test.lua` reports 21 assertions on both sides of that pair, so its MarketDynamics coupling is constant here and cannot account for a difference that appears only at the head. The delta is +22 and the new contract file is 22 rows. 5385 + 22 = 5407. The attribution to om_213 does not hold; the head total is simply one over.

### What the re-run confirms

`git diff --stat` says F905 is the only existing test file whose assertions changed, and the per-file counts say it did not change size: F905 165 both sides, SF-79 33 both sides, experimental_confirm 31 both sides. "Exactly one file moved, 2 assertions of 165, repaired in place" is confirmed rather than taken.

```
py mutate_prelude_i18n.py   exit 0
  KILLED    P1-hastext-always-true
  KILLED    P2-getext-returns-the-key
  KILLED    P3-hastext-returns-truthy-not-boolean
  KILLED    P4-hastext-removed
  SURVIVED  P5-nil-key-guard-dropped   (declared EQUIVALENT)
  restore: sha256 ... MATCHES
```

Every kill arrived as named rows, not as a load error.

### The four judgement calls

**1. `getText` returning the engine's sentence rather than the key. Right, and it is the load-bearing half of the commit.** The defect class this harness serves is people writing guards against strings `getText` never returns; a prelude that returns the key teaches exactly that error and is the reason the `$l10n_` comparison survived review for so long. The mutation battery settles it empirically: P2 puts the key back and the bar goes red on two named rows. The smaller alternative, keeping the key and adding only `hasText`, would have left the harness asserting a shape the engine never produces, moved nothing, and left no bar that could ever notice a regression back to it. Keep it. The one thing I would hold you to is the MAJOR above: you bought fidelity and then hardcoded the single non-engine constant in the file.

**2. F905 is not weakened, and the reason is sharper than "only that key maps to that string".** Under the old prelude `getText(key) == key`, so `T.eq(notices[1], "sf_notify_lime_crop_title")` also passed if the code never called `getText` at all and passed the raw key straight through. Registering the key and asserting the translation removes that degenerate pass, so the new bar proves something the old one could not. Nothing was dropped to make it green: the file's count is 165 on both sides. Leakage is a non-issue: `run-tests.mjs:36` calls `luaL_newstate` per test file, and the two keys appear in no other file in the suite.

**3. Keep the equivalent mutant.** I checked the claim rather than the label: reading a Lua table at a nil key yields nil, so `self.texts[nil] ~= nil` is false either way and the guard is genuinely unobservable. The script's `as_expected` makes SURVIVED the pass condition for P5, so it cannot rot into a silent false green, and a battery carrying only winnable mutations overstates what the bar proves. It is honesty, not clutter.

**4. The `call()` helper earns its readability cost, and P4 is the receipt.** Removing `hasText` produced eight named failing rows, including `hasText is a function :: got nil` and `<absent>` answers from the probes, instead of taking the file down. The script even carries a distinct `KILLED?` verdict for a kill that arrives with no result row, which is the failure mode the helper prevents. Keep it.

### Verdict

**PASS with MAJORs.**

Neither MAJOR is a reason to hold the code: one is a one-line constant, the other is not in this diff at all. Both want doing before the 14-site row is started, because that row is where the wrong sentence would first bite.

Commits on the branch, for the title check:

```
d2104cf4  fix(test): model the engine's i18n in the shared prelude
```

One commit, title matches it, base `development`, one feature one PR clean.

**Design origin: none, confirmed.** Implementation maintenance lane, no R number, no brief, so no DESIGN-CHECK row is owed and none exists. Fred's `n/a` disposition is confirmed as declared, not raised.

**Not checked:** the 48-file figure. I confirmed the diff-level claim it supports (only F905 moved) directly from per-file counts, so I did not rebuild the population behind it.

TheCodingDad-TisonK:
**MAJOR 1 is closed.** The TESTING.md row for this PR is back on tracking main at `26f76c5`. Verified rather than taken: the row shows no +/- at all in a net diff of `139affb` against `main`, which is byte identity with what Fred originally wrote, and the only remaining change to the file is the unrelated WorkerCosts #132 strike.

The remaining MAJOR (`prelude.lua:59`, the missing language suffix) is a one-line constant, not a merge blocker, and is best done before the 14-site row it is the precondition for.

Verdict stands: **PASS with MAJORs**, ready for merge.

PR #973, opened 2026-09-22
fix(l10n): gate 14 translation lookups on hasText, not on the $l10n_ prefix

TheCodingDad-TisonK:
Fourteen sites across thirteen files rejected a translation by comparing `getText`'s return against `("$l10n_" .. key)`. That is the XML attribute prefix, not a string `getText` can produce. `I18N.lua:186` returns `Missing '<key>' in l10n<suffix>.xml` for an absent key, never nil and never empty, so the comparison always passed, the English fallback below it was unreachable, and the player was shown the engine's diagnostic.

**Design-origin: none | Disposition: n/a**
Implementation maintenance lane, no R number, no brief. Row 1 of `ecosystem-dev-tracking/MAINTENANCE.md` (2026-09-21). Row 2, its precondition, shipped as #972 and is closed.

What it actually looks like, from the mutation battery rather than from argument. With the old guard restored, the minimap layer label renders:

```
Missing 'sf_map_layer_n' in l10n_en.xml [N]
```

## Two commits, in this order

**1. `5e21384` interpolate `g_languageSuffix` in the prelude's missing sentence.** Bob's MAJOR from the cold review of #972, ruled by Tyson into this PR rather than a separate one. #972 left the harness spelling one language suffix into the string it models. Harmless there, because every consumer gated on `hasText` or `^missing`; not harmless in a branch that repairs 14 gates against that harness. It lands first and alone for the same reason the prelude preceded the sites at all.

**2. `82d3378` the 14 sites.**

## The rule, stated because it is the part worth keeping

**No repaired site compares an exact sentence, ever.** Not a caveat, a rule, and the reasoning is Bob's from the #972 review:

1. **It is wrong for 25 of 26 players on the day it is written.** The sentence carries `g_languageSuffix`, which `main.lua:1187` sets from each language entry's own `#suffix`. This mod ships 26 languages. And it fails OPEN: the guard stops matching, the fallback never fires, and the engine's error text reaches the player. That is the defect being repaired, reintroduced by the repair.
2. **It is not constant within a session either.** `NPCManager.lua:358` assigns `g_languageSuffix` inside a loop over every installed language. Whether the engine restores it afterwards cannot be read off the decompiled text (`:363` is a self-assignment where a restore would sit, which most likely means the decompiler lost the local). What survives the uncertainty is enough: a global a runtime path reassigns in a loop is not something a mod may treat as fixed.
3. **It is the original defect in a new costume.** The `$l10n_` guard failed because it was a bet on what the engine renders on failure. An exact sentence is the same bet with better odds. `hasText` is not a better bet, it is the engine's own answer to the question actually being asked.
4. **It covers one failure shape.** `hasText` plus type and non-empty also catches a key that exists with an empty or malformed value.
5. **It duplicates an engine private at 14 sites.** One Giants change would disable 14 guards silently, with nothing going red, because the harness would be reconstructing the same wrong string and agreeing with itself.

The `^missing` shapes are not the primary gate either. They are defensive, for an untrusted third-party i18n object, and belong after `hasText`, never instead of it. The three sites that rely on them today are defended and owe nothing, but they are not the model these 14 copy.

## Also in commit 2

- **The dead `g_modEnvironments` branch is gone** from the twelve `tr` helpers. That global appears in **zero** files of the decompiled engine. The eleven now-orphaned `SF_*_MOD_NAME` locals go with it; `SoilSettingsPanel` keeps its own, still used at `:390`.
- **`SoilSettingsUI`'s comment** claimed `g_i18n` "does not know about mod keys", the inverse of what the engine does. `mods.lua:453` sets `modEnv.g_i18n = g_i18n:addModI18N(modName)`, and that instance chains to the global.
- **`RSF-F219`'s local i18n fixture** had `getText` and no `hasText`. After this repair every row it drove would have read as English, and three assertions that pin WHICH key the overlay reaches for went red. It now declares a locale where every key is present and translates to its own name, which keeps the assertions pinning the key and satisfies the gate. Same degenerate-pass problem Bob identified on F905 in #972: under a key-returning i18n, an assertion that the output equals the key also passes when the code never calls `getText` at all.

## Evidence

New bar `tools/test/lua/l10n_gate_sites_test.lua`, 17 assertions, driving `SoilMinimapLayer.buildLayerLabel`, a shipped public function whose only text path is one of the fourteen sites. A copy of the gate inside a test file would pass with no src file touched, so this drives the real one. It is the only repaired site reachable without a GUI fixture; `SoilMapOverlay`'s repaired helper is driven by `RSF-F219`, and the other eleven are dialog constructors.

`tools/test/mutate_l10n_gate_sites.py`, all four KILLED with named rows, restore sha256 verified:

| id | defect | result |
|---|---|---|
| S1 | the shipped `$l10n_` guard | KILLED, 10 named rows |
| S2 | `hasText` dropped, type and empty checks kept | KILLED |
| S3 | truthy `hasText` accepted instead of `== true` | KILLED |
| S4 | type check dropped | KILLED |

Offline confirmation, from a worktree placed BESIDE the mod clones (`om_213_organic_premium_test.lua` loads `../FS25_MarketDynamics` by relative path and errors from anywhere else, which looks like a real failure and is not):

```
cd tools/test
node syntax-check.mjs            ->  Lua 5.1 syntax OK, 103 files
node lint.mjs                    ->  Lint clean, 103 files
node run-tests.mjs               ->  PASS, 0 failed, 103 files
                                     l10n_gate_sites_test.lua 17 passed
                                     prelude_i18n_contract_test.lua 25 passed
py mutate_l10n_gate_sites.py     ->  exit 0, 4 killed with named rows
py mutate_prelude_i18n.py        ->  exit 0, 5 killed, 1 equivalent as declared
```

Do not gate on the suite total. It is a function of another repository's working tree, for the `om_213` reason above.

## Found here and deliberately not taken

**The class was under-enumerated.** `SoilHarvesterPanel.lua:667`, `:820`, `:835` and `SoilSprayerInfoPanel.lua:639`, `:683` carry the same defect in a fifth shape: their only rejection is `find("^%$l10n_")`, with no `hasText`. Five sites in two files, on neither the queue row's list of fourteen nor its list of four defended, so the recorded class of nineteen was short by five. `SoilHarvesterPanel:667` is the worst of them: past the failed guard it runs `string.format` on the engine's sentence, which carries no specifier, so the field id is silently dropped as well.

**`SoilMinimapLayer`'s `sfTr` helper (site 14) has no callers.** Repaired for consistency, so the file does not carry one gate of each kind. Deleting dead code is a separate judgement.

Both recorded for `MAINTENANCE.md` rather than widened into this PR.


sasha-rf (approved):
PASS.

Verified independently, not taken on the PR's word:

- **The `g_modEnvironments` claim.** Grepped the full decompiled engine tree; it appears in zero files, confirming the comment fix in `SoilSettingsUI.lua` (the old comment had the claim backwards) is correct rather than just plausible.
- **The `mods.lua:453` citation.** Read it directly: `modEnv.g_i18n = g_i18n:addModI18N(modName)`, matching the claim that a mod-scoped i18n chains to the global rather than being a separate unreachable thing.
- **The pattern itself**, spot-checked in full on `SoilMinimapLayer.lua` (two sites) and `SoilSettingsUI.lua`: every one of the 14 sites now gates on `hasText(...) == true` before calling `getText`, never inspects the returned string, and falls back safely through a `pcall` on both calls with a type + non-empty check after. Consistent shape across all 13 files, which is what you'd want from a mechanical repair rather than 14 independent judgment calls.
- **Full suite**, run from a worktree placed beside the mod clones (the PR is explicit that a scratchpad-path worktree fails `om_213` on its relative MarketDynamics dependency, learned the hard way on #972, applied correctly here): `PASS, 5427 assertions passed, 0 failed, 103 files`. The two named files match exactly: `l10n_gate_sites_test.lua` 17 passed, `prelude_i18n_contract_test.lua` 25 passed.
- **Mutation battery** (`mutate_l10n_gate_sites.py`): exit clean, all 4 (S1-S4) KILLED with named rows, restore verified by sha256 match.

The rule stated in the body (no repaired site ever compares an exact sentence, only `hasText`, because the sentence carries a language suffix that isn't fixed even within a session per `NPCManager.lua:358`) is the right generalization from the #972 review, and this PR is a faithful application of it at all 14 sites rather than a partial one.

One thing noted rather than blocking: the `RSF-F219` fixture repair (giving the local i18n a locale where every key translates to its own name) is the same shape as the F905 degenerate-pass problem from #972, an i18n that returns the key would also pass an assertion checking the output equals the key. The PR names this explicitly rather than hiding it, which is the right way to carry a known-imperfect test forward.

Approved as sasha-rf. Since this is the PR that changes player-visible text and the non-English #969 tester check is already live, I'm watching for either failure shape it predicts: a "Missing '...' in l10n_XX.xml" sentence, or a bare key name like "sf_map_layer_n", from any non-English report; both route to Fred per his last message.


TheCodingDad-TisonK:
## Bob, cold diff review (POST-MERGE)

This landed after the merge. It is a record, not an objection: #973 merged as `516a4d06` before this was posted. Stated plainly so a later reader knows which it was. **Had it been in time it would have been PASS.** Nothing below would have blocked the merge. Two MINORs, one of which is outside the diff.

Read at `82d3378a`, base `origin/development` `4014197c`. Two commits, the suffix fix first and alone as ruled. 13 src files, 2 harness files, 2 new test files. Evidence regenerated in a throwaway detached worktree, removed afterwards.

### Verification

```
node syntax-check.mjs        Lua 5.1 syntax OK, 103 files
node lint.mjs                clean, 103 files
node run-tests.mjs           PASS 5427 assertions, 0 failed, 103 files
  l10n_gate_sites_test.lua        17 / 0
  prelude_i18n_contract_test.lua  25 / 0
  RSF-F219-handful_ph_grain...   118 / 0
py mutate_l10n_gate_sites.py exit 0, S1-S4 all KILLED with named rows, restore sha256 MATCHES
py mutate_prelude_i18n.py    exit 0, P1-P4 + P6 KILLED, P5 SURVIVED as declared, restore sha256 MATCHES
```

Every number reproduces. The gate repair is uniform where it should be: the three canonical lines (`pcall(i18n.hasText, ...)`, `has ~= true`, the type-and-empty check) each occur exactly **14** times in `src`, so one executed bar plus textual uniformity covers the set rather than one site of fourteen. No repaired file retains a live `$l10n_` comparison; the ones that remain are the defended sites and the five below.

`l10n_gate_sites_test.lua` is the right shape. It drives `SoilMinimapLayer.buildLayerLabel`, a real shipped function, and A5 asserts the output is not the raw key name, so it cannot pass on a code path that never called `getText`. A copy of the gate would not have done that.

### Findings

`tools/test/lua/RSF-F219-handful_ph_grain_spec_test.lua:364`: **MINOR**: the repaired fixture keeps the degenerate pass that this PR removes everywhere else. `getText` returns the key, so the three assertions (`:402`, `:416`, `:436`, all `any(out, "sf_...")`) still pass if the code never called `getText` and emitted the literal key string. That weakness is pre-existing and this PR does not deepen it, but it is the exact shape removed from F905 in #972 and ruled out at A5 of the new bar in this one, so the standard is now inconsistent inside a single diff. fix: return a decorated key, `"<" .. key .. ">"`, and keep asserting containment of the key. The assertion still pins which key, and additionally proves `getText` was reached.

`src/ui/SoilTuningPanel.lua:16`: **MINOR**: `SF_TUN_MOD_NAME` is a twelfth orphaned mod-name local and it survives. Nothing reads it: the only remaining readers of any `SF_*_MOD_NAME` are `SoilHandfulDialog.lua:65` and `SoilSettingsPanel.lua:395`, both legitimate. Outside this diff and pre-existing, so not a regression, but the commit's claim is that these were cleaned up and one was missed. fix: delete the declaration when the file is next opened.

### The three things you asked me to attack

**1. The under-enumeration framing is NOT too strong. It is not strong enough, and I have the number.** Your five sites are real, all five verified: `SoilHarvesterPanel.lua:667`, `:820`, `:835` and `SoilSprayerInfoPanel.lua:639`, `:683` reject on `find("^%$l10n_")` with no `hasText`. Three of them are worse than the fourteen in a way worth recording: `:667`, `:820` and `:639` pass the unvalidated return into `string.format` as the FORMAT STRING, so on a missing key the engine's diagnostic becomes the format template. It does not crash, because the missing sentence carries no `%` specifier, but the player gets the diagnostic where a field number belongs.

The deeper problem is the method. I enumerated every `g_i18n` `getText` call site in `src` at the post-merge head `516a4d06`, because a fixed population can be counted and a shape cannot:

```
132  getText call sites in src
 27  within reach of a hasText gate
  7  comparing against $l10n_   (your 5, plus the 2 defended by a ^missing match)
 36  ungated with an `or` fallback that can never fire  (MAINTENANCE row 3)
 62  raw, no gate of any kind
```

So the enumerated class of 24 covers 24 of 132, and **105 call sites are not hasText-gated**. Not all 105 are defects: a raw `getText` on a key this mod ships itself renders correctly in any language whose file has it, and the defect only bites where a key is absent. But nobody has read them, so the defective subset is unknown and cannot be quoted.

That is the answer to whether 24 is the real number or merely the next one: **it is the next one.** Two rounds of enumeration have each found a shape the previous round's grep could not see, and that will keep happening, because a grep written for the last shape is blind to the next by construction. The row should stop counting shapes and instead close over the call sites: 132 is fixed, countable, and shrinks as each is classified. That converges. Shape-grepping does not.

**2. Removing the eleven locals was right, keep it.** A local that nothing reads is worse than two lines of diff, because the next reader has to prove it is dead before touching the file, and the thing it fed (`g_modEnvironments`) appears in zero files of the decompiled engine. I checked the survivors rather than the removals: exactly two readers remain, `SoilHandfulDialog.lua:65` and `SoilSettingsPanel.lua:395`, and both are genuine. The twelfth is the MINOR above.

**3. You have NOT papered over a regression, and I want that stated plainly so it is not over-corrected.** The three red assertions were a fixture limitation, not a code fault: a fixture carrying `getText` and no `hasText` is refused by the repaired gate, which is the gate behaving exactly as designed, and every row then reads as English. Giving the fixture a locale where every key is present is faithful to what the old key-returning fixture already implied. The diagnosis is right and the change is right. What it leaves behind is the MINOR above, which is a different and smaller point.

### Verdict

**PASS**, as a post-merge read. No BLOCKER, no MAJOR. Two MINORs, neither of which would have held the merge.

Commits:

```
5e213846  fix(test): interpolate g_languageSuffix in the prelude's missing-key sentence
82d3378a  fix(l10n): gate 14 translation lookups on hasText, not on the $l10n_ prefix
```

Two commits, title matches the second, suffix fix first and alone as ruled.

**Design origin: none, confirmed.** Implementation maintenance lane, no R number, no brief, so no DESIGN-CHECK row is owed and none exists. The `n/a` disposition is confirmed as declared, not raised. TESTING.md row for #973 is present on tracking main.

**Not checked:** whether any of the 105 ungated call sites is actually reachable with an absent key. That is the sweep the row needs, not something a diff read can answer.

Issue #205, opened 2026-04-20
AI buy fert still not working


antler22:
### Mod Version

1.9.6.0

### Farming Simulator 25 Version

Current

### Game Mode

Singleplayer

### What happened?

Courseplay fertilizing field, menu option for fertilizer set to BUY.  But, the fill unit depletes instead of money being taken out.

### Steps to Reproduce

Hired worker or Courseplay running any fertilizer, set game menu option to BUY, and watch as tank level still depletes.

### Log Snippet

```text
NA
```

### Other Mods Loaded

450 

<img width="1908" height="689" alt="Image" src="https://github.com/user-attachments/assets/397607e4-d876-443d-ad05-50de0292fa26" />

### Anything Else?

Rates are also an issue, I let AI run for 30 seconds.  Total field N showed a +30 increase potential.  After only doing less than 1% of the field, total N had increased +22 leaving only a +8 potential left.  

I should've had to complete the entire field for the +30 gain to take effect, not just 1/2 of the boundary pass.

TheCodingDad-TisonK:
@antler22 Can you try this ZIP please and let me know if it changed something
[FS25_SoilFertilizer.zip](https://github.com/user-attachments/files/26904381/FS25_SoilFertilizer.zip)

antler22:
No change, still depleting, still abnormally high nutrient uptake..

<img width="1886" height="711" alt="Image" src="https://github.com/user-attachments/assets/e29a9c32-918f-4e06-97a3-046ef9b6041c" />

TheCodingDad-TisonK:
2026-04-20 19:15:13.758 [SoilFertilizer] DEBUG: Sprayer/Spreader hook: Field 1, STARTER, 0.1L (x1.00 rate)
2026-04-20 19:15:13.758 [SoilFertilizer] DEBUG: Fertilizer: Field 1, STARTER, 0L
2026-04-20 19:15:13.803 [SoilFertilizer] DEBUG: Sprayer/Spreader hook: Field 1, LIQUIDFERTILIZER, 0.1L (x1.00 rate)

TheCodingDad-TisonK:
It still goes back to LIQUIDFERTILIZER even though i made sure to caught this...

TheCodingDad-TisonK:
And that other part... that is a bug in how field areas are calculated during the map scan

TheCodingDad-TisonK:
Another round (trying to fix this problem) @antler22 

[...]

PR #974, opened 2026-09-22
feat(RSF-F196): custom-product identity, and the density refusal every consumer honours by name

TheCodingDad-TisonK:
A dry product's rate is stated in kg/ha and can only be interpreted through its density. This PR establishes custom-product **identity** independently of price, records a **refusal** at registration for a product whose density is nonsense, and makes every consumer the engine reaches honour that refusal by name: nothing painted, drained, credited or charged for a refused product, on any peer. Slices 1 and 2 of the v0.9 brief, shipped together because "refuses the right product" is not a checkable claim until identity is established.

**Design-origin: RSF-F196 v0.9 | Disposition: CARRIES** (declared in the handback to Bob; he confirms or raises it in the cold verdict). Brief: `Office Tyson/RSF-F196-dry-rate-unit-contract-IMPLEMENTATION-BRIEF.md`, travelled in the September 19 wave. Unified position R4, listed first on its row. Lane FAST TRACK.

## Three commits

| | |
|---|---|
| `461dff5c` | one name population; `customProductIndices` (identity only); `refusedProducts`; the one bounded resolver R1b |
| `d5231850` | density in kg/L from the engine's tonnes-per-litre field; the refusal written at registration, cleared on a valid retry |
| `4f04835c` | R2 the refusal append; R3a-R3d its consumers; the stamp retired; price no longer identity anywhere; V18/L1 |

## What it fires on today: nothing, stated rather than discovered

All twelve dry products declare a valid density (0.60 to 1.10 kg/L). Bob established that **every one of the 60 fill types in `fillTypes.xml` declares `massPerLiter` explicitly**, so the engine default is never relied on. The gate guards a future or third-party mis-declaration. The bench uses a synthetic refused product because there is no real one to point at.

**And density is not what protects the dose.** 28 fill types declare `massPerLiter="0.001"`, one gram per litre, every crop-protection product and every blend. That is positive, so a density test accepts it. If one were ever mass-dosed the conversion would be wrong by a thousand and the gate would not object. **The twelve-product catalogue is the real boundary**; the density test only catches nonsense. The bench pins that directly.

## The unit, because it is a factor of a thousand

The engine stores `massPerLiter` in **tonnes** per litre: `FillTypeDesc.lua:71` reads `physics#massPerLiter` (documented in kilograms at `:290`, default 1) and multiplies by 0.001, over a default of 0.001 at `:13`. `FillTypeManager.MASS_SCALE` is 1. `densityOf` does that scale once, in one place.

**What refusal cannot detect, and why that is correct.** A fill type declaring no density is stored as 0.001 t/L, exactly 1 kg/L, indistinguishable from one declaring `1.0`. "Absent" is unreachable through the loader, and re-parsing the XML to reach it would create a second density authority beside the engine's, the parallel-mechanism mistake this brief argues against everywhere else. The detectable set is a density that is **present and nonsense**: non-positive or NaN. A nil is therefore *not* refused: the loader always writes a number, so nil only appears on an object that never went through it, and refusing it fires the gate on fixtures rather than products. The F187 bar proved that immediately when a first draft refused all twelve solids.

## R2, the refusal, and the clause that was easy to get wrong

An appended `Sprayer.onStartWorkAreaProcessing`, installed **last**, after native's final assignment (`Sprayer.lua:926-931` in our tree; the brief's `:1274` is the official lane) and after the five appends the mod already installs. It zeroes `sprayFillLevel`, `usage` and `usagePerMin` for a refused driving product and leaves `sprayVehicle`, the fill units and the tank alone. Paint gates on `sprayFillLevel` at `:320`; drain reads `usage` at `:942`.

**It is not server-only.** `processSprayerArea` runs on clients while the mod's end hook is server-only. A refusal folded into the server-only rate multiplier would leave every nearby client painting a product the server refused. R2 runs on both, and each peer reaches the same verdict from its own loaded descriptors: no refusal state is transmitted.

## The consumers, in the order the engine reaches them

- **R3a `getExternalFill`**, first, because native calls it from inside `onStartWorkAreaProcessing` (`:889`) before the assignment R2 follows. Identity now comes from the one resolver, never from `prices[]`. A catalogue member in reach is **never delegated to native**, whose `allowFertilizer` cascade sells FERTILIZER for a UREA tank (#205). A refused member, or one with no usable price, returns `UNKNOWN, 0`. The `or 1.0` that charged an unpriced product at a made-up rate is gone.
  **Bob's condition is met by construction**: this is the existing wrapper edited in place, so Hook 9's propagation carries it to every type table and live instance. An already-loaded machine gets the new money writer at install, not at slice 4.
- **R3b AI-1**: its entry condition is exactly what R2 leaves behind, so a numeric refusal was indistinguishable from an empty buy-mode tank and the `or 1` fallback would fabricate a dose. Gated by name before the injection.
- **R3c the residual snap (#764)**: its negative delta reaches the purchase intercept and either charges or drains, in the frame the dose was zeroed. Skipped by name. A valid product keeps the incumbent behaviour, intermittency and AI stop included.
- **R3d the secondary enumeration**: a secondary unit holding a refused product costs only that unit.
- **R3e**: the backup refill needs no refusal gate (R2 zeroed usage first), but its *identity* is now the catalogue rather than the presence of a price, as is the FillUnit purchase intercept's.

**The `_soilLastCustomFillType` stamp is retired**, two writers and two readers. The empty-tank answer it existed for is the resolver's third step: the engine's own retained `lastValidFillType`, which FillUnit syncs in its own stream (`FillUnit.lua:482`, `:541`, exposed at `:699`), so a late-joining client reaches the same product with no private field and no new event.

## V18 / L1, the one new player string

`sf_notify_density_refused`, hand-inserted in all 27 language files in the `<e k= v=>` form this mod uses (never `lang_sync.py`), bytes in and out, every file exactly +1/-0. Rendered: **"Soil Update: UREA has no usable density and was skipped. Use Drain Vehicle before refilling."** Once per product per mission; technical reason logged at the existing throttle; `showNotifications` respected. "Skipped" over "not applied" is Sasha's reading as a player: the notice does not recur, and "not applied" reads as permanent.

## Two fixtures repaired, and the 14 failures split by cause

"The fixture was wrong" is the most comfortable sentence a build seat can say when tests go red, so here is the evidence rather than the sentence.

**Production builds the manager with the class metatable**: `HookManager.lua:11` `HookManager_mt = Class(HookManager)`, `:40` `setmetatable({}, HookManager_mt)` inside `HookManager.new()`, called from `SoilFertilitySystem.lua:143`. The hooks capture `self` and call its methods. `RSF-F226d` and `RSF-F226e` built the manager as a bare table with no metatable.

| cause | rows | cleared by |
|---|---|---|
| no class metatable (resolver method nil) | F226d's whole file (load error) + 9 of F226e's 14 | `setmetatable(..., { __index = HookManager })` alone |
| identity modelled through the retired stamp | F226e's other 5, all in S6 | giving its stub `getFillUnitLastValidFillType`; the stamp line kept so the bar still proves a stale stamp is ignored |

Each repair applied alone, and the count moved exactly as stated.

## Evidence

- `RSF-F196-catalogue_resolver_test.lua`, 61 rows: one population, membership only, candidate order, non-members are not intent, no guessing, nothing throws, density and its unit, the catalogue boundary, refusal at registration and the clearing retry.
- `RSF-F196-density_refusal_test.lua`, 54 rows, driving the **real installed hooks** on an engine model faithful to `Sprayer.lua:926-931`, `:320` and `:942`: valid unchanged, refused is nothing, on every peer, last append is load-bearing, the nurse tank, the late join, no stamp, no fabricated price, stand-down, told once. Its own first draft had two fixture faults the bar caught, named in the file: a vehicle not listed before propagation (Hook 9's exact gap), and `SoilLogger.warning` stubbed as a colon-call.
- `mutate_prelude_i18n.py` unchanged; `mutate_f196_refusal.py`: **8 mutations, all KILLED with named rows**, restore verified by sha256: R2 server-only, zeroing dropped, usage left armed, R3a refused not checked, price fabricated, delegates when a member is in reach, stamp consulted, notice every frame. Slice 1's rows were proved to bite the same way.

```
cd tools/test
node syntax-check.mjs      ->  Lua 5.1 syntax OK, 103 files
node lint.mjs              ->  Lint clean, 103 files
node run-tests.mjs         ->  PASS, 0 failed, 105 files (5542; do not gate on the total, om_213 reads another repo's tree)
py mutate_f196_refusal.py  ->  exit 0, 8 killed
py build.py                ->  zip builds, untracked
```

## Not in this PR, each its own seam

U5b's secondary-unit identity correction (under the unit rule, not the refusal); V13's token and wrapper ownership per dispatch slot (slice 4, **which RSF-741 item 9 extends rather than inventing its own**); Drain Vehicle recovery (slice 5); V19's #764 speed family. Per Tyson's sequencing the N/P/K/OM scalar-stamp fix follows this PR before those slices, because both edit inside `applyFertilizer` and share `:5905`'s `factor`.


sasha-rf (approved):
PASS.

Checked state (OPEN) and the commit list before fetching anything, fetched 4f04835c by exact SHA, reviewed from a fresh worktree.

Verified independently rather than taken from the body:
- **The unit conversion**, `FillTypeDesc.lua`: `massPerLiter` is stored as `xmlFile:getValue(... , default) * 0.001`, documented in kilograms at the schema (`"Mass per liter/unit in kilograms"`, default 1) but held internally as tonnes/L, default `0.001`. Exactly the factor-of-a-thousand the PR names.
- **R2's shape**, read in full: `Sprayer.onStartWorkAreaProcessing` appended with `Utils.appendedFunction` (runs after native, after every existing append), zeros `sprayFillLevel`/`usage`/`usagePerMin` only, leaves the tank/fill-units/`sprayVehicle` alone, and carries **no server/client gate anywhere in the function**, confirmed by reading it end to end, not by trusting the comment above it. That's the load-bearing "not server-only" claim, and it's real.
- **The retired stamp's replacement**: `FillUnit.lua` genuinely tracks and syncs `lastValidFillType` (`:482` stream write, `~:700` getter), so the claim that a late-joining client reaches the same product with no private field checks out.
- **The fixture diagnosis**: `HookManager.new()` does use `setmetatable({}, HookManager_mt)` with `Class(HookManager)`, confirming the metatable claim rather than assuming the PR's own account of "the fixture was wrong" is honest by default.
- **The new string**, spot-checked in `en` and `de`: English carries the real text, German (and by the PR's own count, all 26 others) carries the `[EN]` placeholder. "Skipped" is in, matches what shipped in #973's sibling work today.
- **Suite**: syntax/lint clean, `PASS, 5542 assertions, 0 failed, 105 files`, matching exactly.
- **Mutation battery**: `mutate_f196_refusal.py`, all 8 (R2-server-only, R2-zeroing-dropped, R2-usage-left-armed, R3a-refused-not-checked, R3a-price-fabricated, R3a-delegates-when-member-in-reach, R1b-stamp-consulted, V18-notice-every-frame) KILLED with named rows, restore sha256-verified.

TESTING.md row (Fred flagged it for my eye specifically): good as written. It correctly frames this as a non-regression check since the refusal fires on nothing today, tells testers explicitly not to go looking for the notice, and names the one path whose identity genuinely moved (an AI helper on an emptied custom tank, now identified through the engine's own `lastValidFillType` instead of a retired private stamp) with a clean falsification condition. No changes needed.

Approved as sasha-rf.


TheCodingDad-TisonK:
## Bob, cold diff review

Read at `4f04835c`, base `origin/development` `516a4d06`. Three commits, one src file, two new bars, one battery, two repaired bars, 27 translation files. Evidence regenerated in a throwaway detached worktree, removed afterwards.

**Disposition CARRIES is CONFIRMED, not raised.** What is built implements the clauses declared, and none of it moves intent. The one thing I would have raised, the "absent density" question, is correctly a limit rather than a WITHHELD: `FillTypeDesc.lua:13` and `:71` mean the loader always writes a number over a schema default of 1 kg at `:290`, so absence is unreachable rather than undecided, and refusing a nil would fire the gate on fixtures instead of products.

### Verification

```
node run-tests.mjs      PASS 5542 assertions, 0 failed, 105 files
  RSF-F196-catalogue_resolver_test.lua   61 / 0
  RSF-F196-density_refusal_test.lua      54 / 0
  RSF-F226d-overlap_nutrient_skip        21 / 0
  RSF-F226e-blocked_pass_billing        113 / 0
py mutate_f196_refusal.py  8 of 8 KILLED with named rows, restore sha256 MATCHES
```

**Every engine claim in the code comments checks out against `D:\FS25_Decoded`, and the citations are OURS rather than the brief's official lane.** That matters because the brief cites `Sprayer.lua:1274-1284` for native's final assignment and `FillUnit.lua:1571-1580 / :593-605 / :645-675 / :920-925` for last-valid, none of which land in our tree. The code cites `Sprayer.lua:926-931` and `FillUnit.lua:699 / :482 / :541`, which do:

- native's final dose assignment is `:925-931`, `sprayFillType` `:926` through `sprayVehicleFillUnitIndex` `:931`
- `processSprayerArea` gates on `sprayFillLevel <= 0` at `:320`, and that return is NOT server-gated, so zeroing it stops paint on clients too
- `onEndWorkAreaProcessing` reads `usage` at `:942` under `self.isServer` at `:940`
- `getExternalFill` is called from inside native start at `:889`, before the final assignment, which is why R3a has to refuse on its own
- `getFillTypeIndexByName` `FillTypeManager.lua:305`, `getSprayerFillUnitIndex` `Sprayer.lua:88`, `Utils.appendedFunction` `:380`, `prependedFunction` `:387`

**The "last append" claim is true, and file order does not show it.** There are six appends and two prepends on `Sprayer.onStartWorkAreaProcessing`, and one append sits at `:7860`, textually far below the refusal at `:2244`. Install order is what counts: `installAll` calls `installSprayerStartHook` (that `:7860` append, the rate multiplier) at its relative line 138 and `installDensityRefusalHook` at 229, so with `appendedFunction` the refusal genuinely runs last. The two things installed after it are both prepends on this function: `installOverlapPreventionHook` (`:2337`) and `installSectionStatePreserver` (`:2657`, whose only append is on `onEndWorkAreaProcessing`, after the drain). Nothing installed later can re-arm a zeroed dose.

**The not-server-only decision CARRIES the brief rather than improving on it.** The brief says it twice: "Not on the server, not on a nearby client" and, in the round-four rationale, "The append-order rule exists because a refusal inside the server-only hook leaves clients painting." The code matches, and `Sprayer.lua:320` is why it works.

**The new l10n key is clean:** `sf_notify_density_refused` in 27 of 27 translation files, in the `<e k= v=>` form this mod actually uses (1386 of them in `translation_en.xml`, zero `<text name=>`), English real and the other 26 carrying the established `[EN] ` marker.

### Findings

`ecosystem-dev-tracking/DESIGN-CHECK.md`, the #974 row: **MAJOR**: the row's "not carried yet" list is incomplete, and the row is the artifact whose whole job is that accounting. It names U2/U3, the residual-drain family, slice 4 wrapper ownership, slice 5 Drain Vehicle recovery and U5b. **V7 is in neither list.** The brief's V7 requires `onFertilizerApplied` to validate before `applyFertilizer` and return exactly `false` before the nutrient buffer, agronomy, organic input, overlay refresh or multiplayer broadcast, and its own build note warns "the boolean has to be wired at both call sites or the gate is inert". On this branch `onFertilizerApplied` at `SoilFertilitySystem.lua:926` still calls `applyFertilizer` and returns nothing. A reader of the row concludes V7 is carried, because the row says what is not.

**The code is fine, and I want that separated from the record.** Nothing reaches a credit for a refused product today: R2 zeroes `usage`, so the area hook's `liters` is 0, so `factor` at `:5905` is 0 and every delta is 0; and the secondary caller at `:4688` is gated by R3d on `fu.fillType`. So V7 would be a second, explicit gate over paths already closed, and its absence is not a live defect. fix: add V7 to the row's not-carried list, or state why it is considered carried. No commit needed on the branch.

`src/hooks/HookManager.lua:7597-7601`: **MINOR**: the retired stamp's instructions survive above their own retraction. The original three-step priority comment, including "3. Fall back to `_soilLastCustomFillType` (stamp set by the Sprayer-area hook; covers the empty-tank AI case...)", is still in place, and the new RSF-F196 note explaining that those three steps were replaced begins on the next line. A reader meets the instruction before the correction. Every other `_soilLastCustomFillType` mention in `src` is now comment-only, so nothing reads the field; this is the one place the stale text reads as current. fix: delete the superseded four lines.

### Verdict

**PASS with MAJORs.** The MAJOR is on the DESIGN-CHECK row rather than the diff, and the MINOR is four lines of comment. Nothing here blocks the merge.

```
461dff5c  feat(RSF-F196): the custom-product catalogue and the one intent resolver
d5231850  feat(RSF-F196): density, and the refusal written at registration
4f04835c  feat(RSF-F196): the density refusal, its consumers, and the stamp retired
```

Three commits, base `development`, one item, title matches. **Design origin: RSF-F196 v0.9**, DESIGN-CHECK row present with CARRIES, **confirmed cold here**. TESTING.md row for #974 is present on tracking main.

**Not checked:** in-game behaviour on any path; the unit-rule conversions U2/U3, which are not in this slice; and whether the twelve dry products' registered LPS values are right, which is the unit contract rather than the identity contract.

PR #933, opened 2026-09-10
feat(SF-79): sprayed-area lime and chemical pH (WIP)

TheCodingDad-TisonK:
## SF-79 Sprayed-area lime and chemical pH — host build

**Draft.** `FIXED_FOR_STAGE_6B`, APPROVED by Bob 2026-09-09, recipient
IMPLEMENTATION. Companion reader: `FS25_FarmTablet/SF-79-soil-ph-reader.md`
(Wizard). Release stays LOCKED. Merging not requested.

Makes the existing `sfSoilMap_PH.grle` layer the chemical-soil authority; the
field number becomes a **report** derived from its written pixels over the
complete cultivated parcel domain. A report never paints chemical pH. No new
grid, product, target, rate curve, clock, app or automatic repeat-job policy.

### Slices (all 6 landed)
- [x] Slice 1 — `src/PositionalPH.lua`: `_applyPHFootprint` over
  `DELTA|SET|NORMALIZE` × `POINT|STRIP|POLYGON|FIELD`; PH carrier grain;
  complete cultivated parcel domain; saturation cohorts established **before**
  the interior add (`240+10 -> 250`, not 255); pure kernel statics.
- [x] Slice 2 — derived `_phReport`/`_ensurePHReport`
  (CURRENT/EMPTY/UNAVAILABLE), preservation-first migration from the frozen
  scalar, domain-keyed sub-step remainders, `soilData` XML (`#sf79PHSchema`,
  `#sf79PHSeed`, `sf79PHPending(i)`) + StateLedger table.
- [x] Slice 3 — every pH writer routed (boom strip, narrow-tool dot, daily +
  meadow normalize, rain, burn, scorch, admin). pH removed from
  `VM_NUTRIENT_KEYS` scalar replay and the `SpatialNutrients` banded second
  paint; sub-step whole-field deltas bank as remainders instead of dropping.
- [x] Slice 4 — `getFieldInfo` contract (`pHStatus`/`pHGrainMetres`/
  `pHRevision`/`pHLastKnown`; pH nil when stale/unavailable;
  `needsFertilizationKnown`) + ~20-consumer nil sweep; `updatePHWorkAuto`
  pH-aware AUTO rate applied once before the multiplier.
- [x] Slice 5 — `pHReportValid` beside the pH report on
  `SoilFullSyncEvent`/`SoilFieldBatchSyncEvent`/`SoilFieldUpdateEvent` and the
[...]

PR #975, opened 2026-09-22
fix: the application dot adds this tick's delta instead of stamping the field scalar

TheCodingDad-TisonK:
The narrow-tool application dot in `applyFertilizer` wrote the **field scalar** over a 5 m square through `writeValueAtWorld`, an unfiltered set. Every pixel there was levelled to the field average, erasing the per-pixel record the boom strip had built. It now **adds this tick's delta** through `addPaintStrip`, the additive primitive the strip already uses, on the same square. Four layers (N, P, K, OM) plus the guarded pH fallback.

**Design-origin: none | Disposition: n/a.** A defect fix inside the REFINED per-pixel write; no brief, SDS or RSF specified the dot. Sequenced by Tyson after F196 slice 2 and before slices 3 to 5, because it edits inside `applyFertilizer` next to the `factor` F196's unit rule will redefine. Lane FAST TRACK. One commit.

## The defect, and who reached it

The dot is the NARROW-TOOL fallback: a tool whose span is under one cell yields no boom points, so the strip never paints and this dot is its only per-pixel write. But the defer flag (`field._vmBoomPaintTime`) is set by the strip, which runs **after** the dot in the same tick, so the dot reads the previous tick's flag and **every tool's first tick of a pass reaches it**. Bob established that in intake. So: a narrow tool crossing a boom strip flattened the strip, and a wide sprayer starting a pass stamped the field average over the 5 m square under it.

## The fix

```lua
local sx, sz, wx, wz, hx, hz = sprayX - r, sprayZ - r, sprayX + r, sprayZ - r, sprayX - r, sprayZ + r
local dotN = (entry.N or 0) * factor * tunFert
if dotN ~= 0 then vm:addPaintStrip("nitrogen", sx, sz, wx, wz, hx, hz, dotN) end
```

- **The primitive** is `addPaintStrip`, on the parallelogram `writeValueAtWorld` built for the same radius: `(x-2.5, z-2.5), (x+2.5, z-2.5), (x-2.5, z+2.5)`. One writer, one expression, no drift from the strip.
- **Not `addValueAtWorld`**, which the intake first named. Its own SF-934 header forbids it for incremental field writes: it samples one pixel and levels the result across the square, the same defect family under an additive name. Bob verified and accepted the correction; the class is reading a body from its signature and not the twenty lines above it.
- **The delta** is the same expression as the scalar update above it and the pH dot beside it, `entry.X * factor * tunFert`. A zero coefficient or a zero-litre tick (every pass's first, dt 0) paints nothing, where the stamp used to write the scalar regardless.
- **The SF-79 pH route** through the positional writer is untouched. The guarded fallback for a system without it (never live in production) is swept to the same additive square rather than left as a stamp.

## Seeding: none, like the strip. On a normally seeded field nothing changes.

This is the question Bob asked to have brought back rather than picked, and the pack's "Unwritten nutrient pixels" section names the tradeoff. The three siblings, read from source:

| sibling | seeds? | why |
|---|---|---|
| birth seeder `vmSeedField` (`:3894`) | the whole field polygon, at load, for `VM_NUTRIENT_KEYS` = N, P, K, OM (`:3770`) | layers restored from savegame are skipped |
| boom strip `paintBoomStrip` (`:4443`) | nothing | relies on the birth seeder; `addPaintStrip` bands from `def.rawFloor or RAW_MIN`, so raw 0 is skipped |
| pH `_seedPHFootprint` (`PositionalPH.lua:487`) | **dead code**: no callers in `src` or `tools` (Bob, verified by name) | pH is not in the birth seeder's list; brief 3.B specified a footprint seed that was never invoked |

So pH was not a sibling to match; "match pH" would have meant building pH's unbuilt seeding for four layers it was never specified for. Bob is raising the dead seeders separately under SF-79; they are not folded in here.

**What actually changes.** Today's stamp wrote unrecorded pixels too, at the field average, only as a side effect of being unfiltered. After this PR the dot skips them like the strip. On a normally seeded field that is **no behavioural change at all**: every pixel inside the polygon is recorded at birth. On ground the seeder never covered (a field whose polygon did not resolve, or worked ground outside it) the narrow-tool dot no longer invents records the boom strip on the same ground would not. Both tools now leave the same population on the same ground, and SF-73 sees that ground as unwritten share. Seeding it from the **live** field average mid-pass would recreate the pack's "one pass grades itself as the field average rises" hazard; the frozen pass baseline that avoids it is SF-73's artifact to define, not this fix's.

## Magnitude: shape only, Tyson's ruling

The dot lands the field-average delta on a 25 m2 square. The strip rescales the same delta by `sd.area / areaHa` to a local dose; the dot has no travel geometry to rescale by, and consecutive squares overlap tens of times at driving pace, so "divide by the square's area" is not the obvious answer either. The B1 MAINTENANCE row already flags the pH dot's magnitude as an unmeasured under-dose with "measure before fixing". **The two dots have different footprints**: the pH dot passes no radius and lands on one grain cell (about 2.25 m2); this one passes 2.5 and lands on 25 m2, about eleven times the ground. The row is widened in the same turn to say the four layers share that shape **by this PR**, not by discovery of an older instance.

**Residual, stated rather than found:** a wide sprayer's first tick still reaches the dot, now as one tick's field-average delta into a 5 m square instead of a flattening stamp.

**Bob's cold-review finding, and the sharper question it leaves (record, not code).** `addPaintStrip` quantises before it writes: `rawDelta = floor(delta / upr)` and a zero raw delta returns silently (`SoilValueMaps.lua:751-752`). One raw step is `(maxVal - minVal) / 254`, about 0.394 for N, P and K, 0.039 for OM, 0.0098 for pH. The dot's delta is one tick's field-average delta with no area rescaling, so for FERTILIZER (N = 41.1) on a 5 ha field at default multipliers one raw step needs about 48 litres in a single tick, which never happens; the most concentrated product on 1 ha with both multipliers at 2.0 crosses it at roughly 0.12 L per tick. **Across most of the realistic range the new write lands nothing**, and the population this path serves (narrow tools, the smallest per-tick usage of anything) is its own worst case. The same arithmetic makes the SF-79 pH dot this PR's expression matches inert across the whole realistic range since #933, so "same expression as the pH dot" above is faithful but not reassurance. The change is therefore from "always writes, the wrong value" to "usually writes nothing, the correct value when it writes": strictly better, which is why this is a MAJOR and not a blocker. The B1 MAINTENANCE row now carries the sharper question: not "how much too small" but at what product, field size and multiplier the dot crosses one raw step at all. The answer is frequently never, and that invites a different fix from division: whether the dot should bank a sub-step remainder the way `_phApplyField` does for field deltas (`PositionalPH.lua:441-462`), which is a design question. **The bench cannot see this by construction**: the primitive is a declared stub, so the 49 rows prove the arguments are right, and the sub-step return is the one behaviour that decides whether a right argument becomes a write.

## Evidence

- `sf_application_dot_delta_test.lua`, **49 rows**, driving the **real `applyFertilizer`** with a recording stand-in for the value maps: no stamp ever; one additive write per non-zero layer; the painted delta equals the scalar's own increment this tick; the three corners; a zero coefficient and a zero-litre tick paint nothing; consecutive ticks and two VWW sections of one tick each paint their own delta (the tick stash belongs to the strip); the deferral window unchanged at both edges and against a future flag; no position and no value maps; the pH fallback and the pH positional route; the scalar ceiling (the primitive saturates per pixel, as for the strip).
- `mutate_sf_dot_delta.py`: **10 mutations, all KILLED with named rows**, restore verified by sha256: the stamp restored, the scalar fed to the add, the tick stash read, the zero guard dropped, the radius shrunk, the defer gate dropped, the window doubled, the pH fallback stamp restored, OM dropped, P fed N's delta.

```
cd tools/test
node syntax-check.mjs      ->  Lua 5.1 syntax OK, 103 files
node lint.mjs              ->  Lint clean, 103 files
node run-tests.mjs         ->  PASS, 0 failed (do not gate on the total; om_213 reads another repo's tree)
py mutate_sf_dot_delta.py  ->  exit 0, 10 killed, restore MATCHES
py build.py                ->  zip builds, untracked
```

## Not in this PR

The dose magnitude (MAINTENANCE, measure first). The dead pH seeders (SF-79, Bob's separate finding). The AUTO whole-rate item (two defects, queued behind this). F196 slices 3 to 5 and the two comment-only #974 MINORs owed to slice 3.


sasha-rf (approved):
PASS.

Checked state (OPEN) and got the head SHA fresh before fetching. One commit, e55b90e4, based directly on #974's merge. Reviewed from a fresh worktree.

Read the actual diff rather than the commit message's account of it: confirmed the old code called `writeValueAtWorld` (an unfiltered SET of the field scalar) for N/P/K/OM/pH-fallback, and the new code computes `entry.X * factor * tunFert` per layer (the exact same expression as the scalar update immediately above it and the pH dot beside it) and writes it through `addPaintStrip` on the same 5m-radius parallelogram `writeValueAtWorld` used to target. Zero-guarded per layer (`if dotN ~= 0 then ...`), so a zero coefficient or a zero-litre tick genuinely paints nothing, where the old stamp wrote regardless.

Suite: syntax/lint clean, PASS, 5591 assertions, 0 failed, 106 files. 5542 (from #974) + 49 new bench rows = 5591, matches exactly, and the file count is +1 for the new bench file, also exact.

Mutation battery: `mutate_sf_dot_delta.py`, all 10 (N-stamp-restored, delta-is-the-scalar, dot-reads-the-tick-stash, zero-guard-dropped, radius-shrunk, defer-gate-dropped, defer-window-doubled, pH-fallback-stamp-restored, OM-write-dropped, P-fed-N-delta) KILLED with named rows, restore sha256-verified.

Design-origin none, correctly stated for a defect fix inside an already-REFINED per-pixel write with no brief behind the dot itself. Sequencing (after F196 slice 2, before slices 3-5, since it sits next to the `factor` F196's unit rule touches) is sound and doesn't conflict with anything in #974.

One claim I did not independently re-derive, noted rather than skipped: that the strip's `field._vmBoomPaintTime` write happens AFTER the dot's read in the same tick (making every tool's first tick reach the dot once). I confirmed the write site (SoilFertilitySystem.lua:4464) and the read site (:6019-6020) both exist and are consistent with that claim, but didn't trace the calling order across a full frame to independently prove it. This was already established context from prior work in this area rather than a new claim unique to this PR, so I'm not blocking on it, but flagging that I didn't re-derive it from scratch.

Approved as sasha-rf.


TheCodingDad-TisonK:
## Bob, cold diff review

Read at `e55b90e4`, base `origin/development`, which carries #974 (`c66e4905` is an ancestor). One commit, one source hunk, two new test files.

**Design origin: none, CONFIRMED.** The changed code is the pre-SF-79 fallback and the N/P/K/OM stamps; the SF-79 positional-writer branch keeps its original expression. Nothing here implements a brief, so no DESIGN-CHECK row is owed and none exists. `n/a` disposition confirmed as declared.

### Verification

```
node run-tests.mjs      PASS 5591 assertions, 0 failed, 106 files
  sf_application_dot_delta_test.lua   49 / 0
py mutate_sf_dot_delta.py   10 of 10 KILLED with named rows, restore sha256 MATCHES
```

The geometry substitution is exact, which is the thing most likely to have been almost right. `writeValueAtWorld` builds `(worldX-r, worldZ-r)`, `(worldX+r, worldZ-r)`, `(worldX-r, worldZ+r)` with `POINT_POINT_POINT` at `SoilValueMaps.lua:646-650`; `addPaintStrip` passes its three points to `setParallelogramWorldCoords` with the same coord type at `:755`. The hunk's `sx, sz, wx, wz, hx, hz` reproduce those corners for `r = 2.5`, so the new write covers the identical square. The raw-floor skip is real too: `rawLow = def.rawFloor or RAW_MIN` at `:758`, so unrecorded pixels are left alone, which is the seeding answer.

### Findings

`src/SoilFertilitySystem.lua:6003-6013`: **MAJOR**: across most of the realistic range the new write is **dropped to quantisation and lands nothing**, so the dot changes from "always writes, wrong value" to "usually writes nothing, correct value when it writes". That is still an improvement, and it is not what the PR body says the change does.

`addPaintStrip` quantises before it writes: `rawDelta = floor(delta / upr)` at `:751` and `if rawDelta == 0 then return 0 end` at `:752`, silently. The step is `(maxVal - minVal) / 254`: **0.3937** for nitrogen, phosphorus and potassium (`minVal 0, maxVal 100`, `SoilValueMaps.lua:50-52`), 0.0394 for organic matter, 0.0098 for pH. The dot's delta has no area rescaling, deliberately and correctly, so it is the FIELD-AVERAGE delta: `entry.X * factor * tunFert` with `factor = (liters/1000) / areaInHa * rrMult` and `areaInHa = field.fieldArea` (`:5855`).

Worked, with `liters` being ONE TICK's usage:

- Ordinary case, `FERTILIZER` (N=41.1) on a 5 ha field at default multipliers: `dotN = 0.00822 * liters`, so crossing one step needs about **48 litres in a single tick**. Not reachable.
- Most favourable case in the shipped data, the N=793.6 product on a 1 ha field with `rrMult` and `tunFert` both at their 2.0 maximum: `dotN = 3.174 * liters`, needing about **0.12 L/tick**, roughly 7 L/s. Reachable.

So it writes only for the most concentrated products, on a small field, at boosted settings. **And the population this path exists for is the worst case for it**: the dot is the narrow-tool fallback, tools whose span is under one cell, which have the smallest per-tick usage of anything.

The precedent is worse than the thing that copied it. The same arithmetic on the SF-79 pH dot, which this hunk matches by design: `LIQUIDLIME` (pH=1.07) on 1 ha at 4x needs about **2.3 L/tick**, near 140 L/s. The pH dot has therefore been inert across the whole realistic range since #933, and it is not this PR's job to fix that, but it does mean "the same expression as the pH dot below" is a faithful description of something that does not land either.

**The bench cannot see this, structurally, and that is worth stating rather than treating as an oversight.** The PR body already says `addPaintStrip` is a recording stub, so the 49 rows prove the ARGUMENTS are right, which is exactly what they should prove. The primitive's sub-step return is the one behaviour that decides whether the argument becomes a write, and stubbing it is what makes the bar unable to distinguish this. fix: no code change asked for here. State it in the PR body beside the magnitude note, and carry it into the B1 MAINTENANCE row, because that row currently frames the dot as an unmeasured UNDER-dose and the measurement now has a sharper question: at what product, field size and multiplier does the dot cross one raw step at all. Whoever measures should be told the answer is frequently "never" rather than "about eleven times too small".

Nothing else. No BLOCKER, no MINOR.

### Verdict

**PASS with MAJORs.** The change does what it set out to do, the geometry is exact, the seeding decision is the one we agreed with the corrected premise recorded, and nothing regresses: writing nothing is strictly better than levelling the square to the field average. The MAJOR is that the effect is much smaller than the description, and it belongs in the record rather than in a commit.

```
e55b90e4  fix: the narrow-tool application dot adds this tick's delta instead of stamping the field scalar
```

One commit, base `development`, one item, title matches. TESTING.md row for #975 is present on tracking main and names a real in-game check with a falsification condition.

**Not checked:** in-game behaviour; the density-execution path, which no bench in this repo exercises; and whether the B1 row's widening has landed, which Fred reported as in flight this turn.

PR #976, opened 2026-09-22
feat(RSF-F196): the rate contract and the unit rule (slice 3)

TheCodingDad-TisonK:
A dry product's public rate is **kilograms per hectare**; every stored, drained, billed and transmitted quantity is **engine litres**. Before this slice the registration treated the kilograms as litres and every interpretation site compared raw litres with a kg/ha number, so a hectare of a 0.60 kg/L product drained 60 percent of its configured mass and was credited as if it were all of it. This slice carries the rate contract (C2) and the unit rule (U2 to U5b), and the two comment-only MINORs from the #974 cold review.

**Design-origin: RSF-F196 v0.9 | Disposition: CARRIES** (declared in the handback to Bob; he confirms or raises it in the cold verdict). Brief: `Office Tyson/RSF-F196-dry-rate-unit-contract-IMPLEMENTATION-BRIEF.md`. Slice 3 of five; lane FAST TRACK. One commit.

## Why C2 and U2/U3 are one item and not two

Each half alone is wrong in the **opposite** direction. C2 alone drains the right volume (rate/density litres per hectare) and the nutrient factor then over-credits by 1/density. U2/U3 alone converts litres that were never scaled and under-credits by the density. Either half passes a bench that only checks its own side. Together, one hectare at 1.0x drains exactly the configured mass and lands exactly the configured mass at the factor: the brief's first invariant, and the agronomic identity a full pass had before the repair (bench row C5).

A note on the #974 row: it listed C1 and C2 in neither its implemented list nor its not-carried list. C2 is genuinely not carried until this PR (the solid loop on development registers `customLPS = customRate / 36000` with no density term). That row is rebuilt from the brief's full clause list in this PR's tracking commit, and a DEVIATES row is added for C1's default-equivalent tolerance, both per Bob's ruling in the slice 3 handback.

## What changed

- **C2, registration.** `litersPerSecond = (rateKgHa / kgPerLiter) / 36000`, with `kgPerLiter` from `HookManager.densityOf`, the same authority the refusal uses. COMPOST at 0.60 kg/L draws 8,333 L/ha where it drew 5,000; GYPSUM at 1.10 draws fewer. Prices, the worker premium and the three billing owners are untouched (H2); a hectare of compost costs about two thirds more because the farmer now pays for the material he actually gets (H2a).
- **U2, the one conversion.** `HookManager.rateKgPerLiter(fillType)`: the density for one of the twelve dry products, nil for everything else. `HookManager.massEquivalent(fillType, liters)`: litres times density, or the litres unchanged for a passthrough product (U4: liquids, base-game FERTILIZER and LIME, the crop-protection buffer keys keep their exact arithmetic).
- **U3, the four interpretation sites**, each named so none is left half-converted: the nutrient factor in `applyFertilizer`; the fully-treated comparison; `trackSprayerCoverage`'s litre fallback (it receives a name, so the descriptor is looked up); and the HUD ghost bar, through the **same function** as the threshold, so bar and threshold can never disagree (row F8 proves it at one buffer). `SoilFertilitySystem` wraps the call so a bench loading it without HookManager still runs, at passthrough.
- **U5.** A secondary fill unit is credited under **its own** product's density. No new code: the secondary path applies under `secFillTypeIndex`, so the factor converts by that product. U5a's three open quantity decisions stay recorded, not solved.
- **U5b, the incumbent double credit.** The secondary enumeration excluded the unit at `workAreaParameters.sprayFillUnitIndex`, a field nothing assigns. Verified in our decompiled `Sprayer.lua`: it appears nowhere, and native writes `sprayVehicleFillUnitIndex` (`:231`, `:859-865`, `:895`, `:931`). So the exclusion was always nil, every local unit passed it, and a single-tank custom pass drawing from its own tank was enumerated again as its own secondary: **drained and credited twice, live today**. Now excluded by identity, split on the source (`:863-881`, `:895-896`): when `sprayVehicle == self`, the unit native named and the unit the resolver reads (`getSprayerFillUnitIndex`, `:581`) are both local and both excluded; when the source is another vehicle or nil, its index numbers a unit over there and can collide with an unrelated local secondary, so nothing local is excluded by index and the stale local reading is not used. Identity only: two tanks holding the same product remain the open question the brief names.
- **The two #974 MINORs.** The retired three-step priority comment above `billedExternalFill` is gone; the install-order rationale sits above the `installDensityRefusalHook` call in `installAll` (file order argues the opposite; call order settles append order).

## What a player will see, so nobody files the fix as a bug

- A single-tank dry pass credits nutrients **once** and drains the tank **once**. The previous build did both twice.
- The tank drains **rate/density** litres per hectare: UREA about 218 L/ha (was 168 native plus a second 168 from the double drain), COMPOST about 8,333 (was 5,000 plus 5,000).
- The "fully treated" notice fires at 90 percent of the configured **mass**, which is later in litres than before for a product under 1 kg/L.
- Helper BUY mode on COMPOST costs about two thirds more per hectare; GYPSUM about nine percent less.
- Liquids, base-game FERTILIZER and LIME, and crop protection: unchanged.

## Evidence

- `RSF-F196-unit_rule_test.lua`, **73 rows**: the conversion contract and the twelve-product boundary (A); C2 through the **real** `registerCustomSprayTypes` with the one-hectare mass invariant for three densities (B); site 1 through the **real** `applyFertilizer`, including the pre-repair full-pass identity and two products converted by their own densities (C); site 2 at the litre boundary and the mass boundary (D); site 3 for a dry name, a crop-protection name and no manager (E); site 4 through the **real** `drawNutrientRow` with a recording native bar, plus the row that proves bar and threshold agree at the same buffer (F); U5b through the **real** `installSprayerAreaHook`: single tank credited once and not drained again, a real second tank credited once under its own product, an externally fed pass with a colliding foreign index still crediting the local secondary, native's and the resolver's units both excluded when both are local, no source vehicle (G). Bob's scope note: no existing bar asserts an exact positive credit count, so G1 contradicts nothing.
- `mutate_f196_unit_rule.py`: **10 mutations across three shipped files, all KILLED with named rows**, every file restored and proved by sha256: kilograms as litres, every product converts, divides instead, exclusion never built, excludes on external too, resolver unit not excluded, and each of the four sites unconverted.

```
cd tools/test
node syntax-check.mjs       ->  Lua 5.1 syntax OK, 103 files
node lint.mjs               ->  Lint clean, 103 files
node run-tests.mjs          ->  PASS, 0 failed, 107 files (do not gate on the total; om_213 reads another repo's tree)
py mutate_f196_unit_rule.py ->  exit 0, 10 killed, 3 restores MATCH
py build.py                 ->  zip builds, untracked
```

**What the bench cannot see, stated:** the engine's own spray-type registration and tank drain (native drains from the registered LPS; the bar proves the number registered), density-map execution, rendering. The in-game check in TESTING.md is the drain and the single credit.

## Not in this PR

V7's boolean (own PR after this merges: it edits the same secondary block as U5b, so it is cut after rather than in parallel). Slice 4 wrapper ownership (which RSF-741 item 9 extends). Slice 5 Drain Vehicle recovery. V19. U5a's three quantity decisions.


sasha-rf (approved):
PASS.

Checked state (OPEN) and got the head SHA fresh, f3f75cb5, based directly on #975's merge. Reviewed from a fresh worktree.

Verified the two most load-bearing claims independently, not taken from the body:
- **The double-credit exclusion field.** Grepped the decompiled `Sprayer.lua` for `sprayFillUnitIndex`: zero occurrences. The engine only ever writes `sprayVehicleFillUnitIndex` (`:231`, `:859-865`, `:895`, `:931`). The old SF code was checking a field name the engine never assigns, so the exclusion was always nil and a single-tank pass drawing from its own tank was genuinely enumerated and credited twice, live on development today. The fix (`wapU5b.sprayVehicleFillUnitIndex`) uses the real field name.
- **C2's density division.** Read the actual registration site: `local customLPS = (customRate / kgPerLiter) / 36000`, correctly scoped to the dry/solid path only, the liquid path (`customLPS = customRate / 36000`) untouched.
- Spot-checked `rateKgPerLiter`/`massEquivalent`, both match the described contract (nil for passthrough products, so a dry product with an unusable density never reaches the four interpretation sites, having already been refused at registration by #974/#975's own work).

Suite: syntax/lint clean, PASS, 5664 assertions, 0 failed, 107 files. 5591 (carried from #975) + 73 new rows = 5664, exact; file count +1, exact.

Mutation battery: `mutate_f196_unit_rule.py`, all 10 (C2-kilograms-as-litres, U2-every-product-converts, U2-divides-instead, U5b-exclusion-never-built, U5b-excludes-on-external-too, U5b-resolver-unit-not-excluded, U3-site1 through site4-unconverted) KILLED with named rows, all 3 touched files restored and sha256-verified.

**On the #974 regression Fred flagged:** verified it myself before writing anything here, and it's real. `git grep -n "rebuildCustomProductCatalogue"` across `origin/development` returns exactly three hits: the function definition and two calls from `RSF-F196-catalogue_resolver_test.lua`. Zero production callers. `customProductIndices` stays empty at runtime, so `isCustomProduct` is always false and helper BUY mode for custom products is broken on development right now. This is a real miss in my own #974 approval: I verified the consumer logic (R3a etc.) thoroughly but never checked whether the producer side was actually wired into a load path. Owning that plainly rather than letting it sit as "Fred found it first." Doesn't change my read on #976 itself: #976's own mechanism (density-based conversion by fill-type name) doesn't depend on catalogue membership, so it stands on its own regardless of the hotfix's timing.

Approved as sasha-rf.


TheCodingDad-TisonK:
## Bob, cold diff review

Read at `f3f75cb5`, base `origin/development` `b1b0d1e6`. One commit, three source files, one new bar, one battery.

**Disposition CARRIES is CONFIRMED.** After the C1 lesson on #974 I checked the clause text rather than the row's account of it, clause by clause:

- **C2** says "A valid product registers `litersPerSecond = (rateKgHa / kgPerLiter) / 36000`". The code is `local customLPS = (customRate / kgPerLiter) / 36000`. Exact.
- **U2** says the litres are multiplied by `kgPerLiter` "at the point of use with no stored state". `massEquivalent` does exactly that and stores nothing.
- **U3** names four interpretation sites; the diff converts four, and they all route through the one function.
- **U4** says nil `kgPerLiter` means factor 1. `rateKgPerLiter` returns nil for anything outside the twelve and `massEquivalent` returns the litres unchanged.
- **U5** needs no code, and the reason given is right: the secondary applies under `secFillTypeIndex`, so site 1 inside `applyFertilizer` already converts with the secondary's own density.

### Verification

```
node run-tests.mjs      PASS 5664 assertions, 0 failed, 107 files
  RSF-F196-unit_rule_test.lua   73 / 0
py mutate_f196_unit_rule.py  10 of 10 KILLED with named rows, restore proved
```

**U5b is a real bug fix and I verified the premise in our decompiled tree rather than taking it.** `sprayFillUnitIndex`, the field the old code read, appears NOWHERE in `Sprayer.lua`. What the engine writes is `sprayVehicleFillUnitIndex`, at `:231` (nil), `:865`, `:895` (nil) and `:931`. So the old exclusion was always nil, every local unit passed it, and a single-tank custom pass was enumerated as its own secondary. The double credit was real. The three-way split matches the engine's own source selection at `:864-865` (`sprayVehicle = self` with the local index) and `:875` (another vehicle), and `getSprayerFillUnitIndex` is at `:581` as cited.

The registration fixture is faithful in the detail that would have been easy to get wrong: `COMPOST massPerLiter = 0.0006` is the engine's TONNES-per-litre storage for 0.60 kg/L, not the kg number, so the bar exercises the same unit the loader produces.

### Findings

`src/SoilFertilitySystem.lua:53-58`, `src/ui/SoilHUD.lua:1954-1957`, `src/SoilFertilitySystem.lua:6320-6325`: **MINOR**: every degradation path in the unit rule converts silently by 1, and converting by 1 is precisely the defect this PR repairs. Three instances of one shape: the local `massEquivalent` wrapper returns the raw litres when `HookManager` is absent; the HUD's inline guard does the same; and site 3's `pcall` lookup yields a nil descriptor on failure, which is also passthrough. None is wrong for a passthrough product, and all three are wrong only for the twelve dry ones.

**Production is safe and I checked rather than assumed it.** `src/main.lua` sources `HookManager.lua` at `:140`, `SoilFertilitySystem.lua` at `:225` and `SoilHUD.lua` at `:259`, so the guards cannot fire in a running game. And `getFillTypeByName` upper-cases at `FillTypeManager.lua:312` and is fed engine-sourced names, so site 3's lookup is case-safe.

The residual is the bench, and it is the same shape as the two findings that preceded it. **33 bars load `SoilFertilitySystem.lua` without `HookManager`.** A future bar written that way, asserting a dry-product conversion through the wrapper, would pass at factor 1 and prove nothing, exactly as #974's hand-populated catalogue hid an unwired rebuild and #975's stubbed primitive hid a dropped write. This PR's own bar is not exposed: it calls `HookManager.massEquivalent` directly and pins a non-identity result, `1000 L of UREA is 770`.

fix: keep the guard. Removing it costs 33 `--!load` edits and pulls a large file into bars that have no use for it, which is its own risk. Make the fallback observable instead: record that the passthrough branch was taken, so a bar can assert it was not. Two lines, and it converts a silent degradation into a testable one.

### Verdict

**PASS.**

```
f3f75cb5  feat(RSF-F196): the rate contract and the unit rule (slice 3)
```

One commit, base `development`, one item. **Design origin: RSF-F196 v0.9**, DESIGN-CHECK row present, **CARRIES confirmed cold**. TESTING.md row present on tracking main.

The two #974 comment MINORs are both discharged: the stale three-step stamp comment is gone, and the install-order rationale now sits above the `installDensityRefusalHook` call and says what settles append order is the call order in `installAll`, not file position.

**Not checked:** in-game behaviour; the G group's `customProductIndices` is hand-populated, which is the disclosed limit that the separate catalogue-wiring bar exists to close, so it is not counted against this PR; and I did not re-derive the twelve products' registered LPS values against their public kg/ha rates beyond confirming the formula.

PR #977, opened 2026-09-22
fix(RSF-F196): build the identity catalogue at registration; it had no production caller

TheCodingDad-TisonK:
#974 made `customProductIndices` the one identity fact and pointed every consumer at it. Its only writer, `rebuildCustomProductCatalogue`, had **one caller in the repository, a bench**. Zero in `src`, on development. At runtime the catalogue was `{}` and every custom product resolved as "not custom". This PR wires the rebuild into registration, which runs at install and on every dedi retry, and adds the bar #974 was missing: one that starts from production's own entry point and populates nothing by hand.

**Design-origin: RSF-F196 v0.9 | Disposition: CARRIES.** V12's catalogue and V12a's mechanism (rebuild on every registration attempt so a late fill type is a product the moment it exists), as written. A repair of #974's own wiring, no intent moved. Lane FAST TRACK. One commit.

## What was broken on development, from #974's own code

- **Billing off.** Before #974 both billing gates keyed on the price map, which the purchase hook populates. #974 required the catalogue as well, so custom-product purchase billing and the AI premium charge were off, and R3a delegated every external fill to native's cascade, the #205 shape it exists to prevent.
- **Refusal dead.** Every step of `resolveCustomProductIntent` gates on `isCustomProduct`, so the resolver always returned nil and every refusal it feeds could not fire: R2 itself, R3b, R3a's refusal, the opt-in refusal. Only R3c and R3d survived, because they test the fill type directly against the refused table, which registration does populate. Masked rather than harmful today (all twelve densities are valid), but "the refusal fires on nothing" and "the refusal could not fire" are different claims and only the second was true. Bob's radius, verified independently by Bob and by Sasha.

## The fix

One call at the end of `registerCustomSprayTypes`. It runs on a partial registration too, deliberately: a catalogue built only once every name has loaded would leave a dedi's first attempts empty, and the mutation that gates it on completion is killed.

## Why three bars missed it, because the next one has to know

The catalogue resolver bar calls the rebuild itself; the density refusal bar, F226e and the slice 3 bar set `customProductIndices` by hand. Every bar proved the hooks **given** a catalogue and none proved production builds one. A mutation battery cannot see it either: it mutates the code under test while the fixture supplies the world, so "nothing populates this in production" is invisible to every mutation. **A bench proves the mechanism given its inputs and can never ask where the inputs come from. At least one bar per feature has to start from production's own entry point.** Three seats missed it the same way: authorship, cold review, approval, each verifying the consumer side and none asking who produces the input at runtime.

## Evidence

- `RSF-F196-catalogue_wiring_test.lua`, **25 rows, nothing populated by hand**: (A) after production registration alone, a dry product, a liquid and a tank mix are members, base-game FERTILIZER and an undeclared name are not, a nonsense density is refused **and** a member, and a partial registration still catalogues; (B) the dedi retry, a late fill type becomes a product; (C) **real billing** on the production-built catalogue with the price map built by the real purchase hook: UREA identified, native's cascade never consulted, the helper charged, the dose armed; (D) Bob's row: a synthetic refused product registered for real and **R2 proven to zero its dose**, paint nothing, drain nothing, tell the farmer once, and leave a valid product alone.
- `mutate_f196_catalogue_wiring.py`: **2 mutations, both KILLED with named rows** (14 rows each), restore proved by sha256: the caller removed, the rebuild gated on a complete registration.
- Two existing fixtures touched the same way: `rsf_f187_gypsum_identity_test.lua` and `RSF-F196-unit_rule_test.lua` (its group B) drive the real registration without the engine's `FillType` global, which the wired rebuild compares against. Each gains the global (three lines); their 18 and 73 rows are unchanged. `FillType` exists before any mod loads, so the fixtures are fixed rather than the production read guarded, which would be a fail-open with no production case.

```
cd tools/test
node syntax-check.mjs               ->  Lua 5.1 syntax OK, 103 files
node lint.mjs                       ->  Lint clean, 103 files
node run-tests.mjs                  ->  PASS, 0 failed, 107 files (do not gate on the total; om_213 reads another repo's tree)
py mutate_f196_catalogue_wiring.py  ->  exit 0, 2 killed, restore MATCHES
py build.py                         ->  zip builds, untracked
```

## What the bar cannot see

The engine's own spray-type registration and the real `FillUnit` chain (the purchase hook is installed against a stub `FillUnit`); the in-game check is the #974 TESTING row's helper BUY mode on a custom product, which this PR makes runnable.


sasha-rf (approved):
PASS.

Checked state (OPEN) and got the head SHA fresh, a354a96b, based directly on #976's merge. Reviewed from a fresh worktree.

Given what this PR is fixing, verified the exact thing #974's review missed: not just that `rebuildCustomProductCatalogue` exists and works when called, but whether anything in production actually calls it. `git grep -n "rebuildCustomProductCatalogue"` on this head now returns a real production call at `HookManager.lua:1057`, inside `registerCustomSprayTypes`. Traced that function's own two callers: `HookManager.lua:620` (inside the install sequence, right after the "Hook installation complete" log line) and `SoilFertilityManager.lua:1444` (inside `_updateDeferredInit`, the dedi retry loop). Both are genuine production paths, not test-only, so the catalogue is actually built now, at install and on every dedi retry.

Spot-checked the new bench's billing claim directly: it builds its `HookManager` instances through the real `registerCustomSprayTypes` call rather than hand-populating the catalogue, then drives an emptied UREA tank through the actual purchase hook and asserts the external fill identifies as UREA rather than falling through to the cascade's FERTILIZER default. That's testing the wiring, not just the function.

Suite: syntax/lint clean, PASS, 5689 assertions, 0 failed, 108 files. 5664 (from #976) + 25 new rows = 5689, exact.

Mutation battery: `mutate_f196_catalogue_wiring.py`, both mutations (rebuild-not-wired, rebuild-only-when-complete) KILLED with named rows, restore sha256-verified.

This closes the gap cleanly: the miss in #974 was three seats verifying the consumer side and nobody checking the producer's wiring, and this PR's own bar is built specifically not to repeat that, sourcing its catalogue from the real registration path rather than supplying it by hand.

Approved as sasha-rf.


TheCodingDad-TisonK:
## Bob, cold diff review

Read at `a354a96b`, base `origin/development` `181fc1f0` (the merged #976). Eleven source lines, one new bar, one battery, two repaired fixtures.

**Disposition CARRIES is CONFIRMED.** V12's catalogue and V12a's rebuild-on-every-attempt, as written. Nothing withheld, no intent moved. This repairs #974's own wiring, and I record plainly that I reviewed #974 cold and passed it without asking who populated the catalogue at runtime.

### The chain, verified to production rather than to the function

This is the question whose absence caused the regression, so I traced it end to end instead of reading the call in isolation:

```
SoilFertilitySystem:initialize()          :248
  -> hookManager:installAll(self)         SoilFertilitySystem.lua:364
     -> self:registerCustomSprayTypes()   HookManager.lua:620   (installAll begins :438)
        -> self:rebuildCustomProductCatalogue()   the new call

SoilFertilityManager:update(dt)           :1481
  -> self:_updateDeferredInit(dt)         :1501
     -> hm:registerCustomSprayTypes()     :1444   (the dedi retry, until complete)
        -> the same rebuild
```

Both ends are real production entry points, `initialize` and `update`, not helpers. The catalogue is therefore built on the install path and rebuilt on every retry, which is what V12a asks for.

### Verification

```
node run-tests.mjs      PASS 5689 assertions, 0 failed, 108 files
  RSF-F196-catalogue_wiring_test.lua   25 / 0
  RSF-F196-unit_rule_test.lua          73 / 0   (unchanged count after the fixture repair)
  rsf_f187_gypsum_identity_test.lua    18 / 0   (unchanged count)
py mutate_f196_catalogue_wiring.py  2 of 2 KILLED with named rows, restore sha256 MATCHES
```

Both mutations are the right two. `rebuild-not-wired` is the regression itself, and `rebuild-only-when-complete` catches the tempting wrong placement, gating the rebuild on `skipped == 0` so a partial registration would leave the catalogue empty. The second is the one a careful author would most plausibly have written.

**The bar closes the whole blast radius, not just the symptom.** Group A proves membership from real registration with nothing hand-populated, and that a refused product stays a member. Group B proves the retry catalogues a fill type that appears between two calls. Group C drives the real purchase hook to build the price map and then the real external-fill wrapper on an emptied UREA tank, which exercises the resolver's third step, the engine's own `lastValidFillType`. Group D registers GYPSUM with a genuinely nonsense density (`-0.001`, negative, not merely odd) and proves R2 zeroes the dose, which is the half of the radius that was masked because nothing is refused in shipped data.

**The FillType fixture decision is right and consistent.** Fixing the two fixtures rather than guarding the production read is the same call made on #976: `FillType` exists before any mod loads, so a guard there would be a fail-open with no production case. Both fixtures kept their assertion counts, 73 and 18, so the repair added the global and changed nothing else.

### Findings

None. No BLOCKER, no MAJOR, no MINOR.

### Verdict

**PASS.**

```
a354a96b  fix(RSF-F196): build the identity catalogue at registration; it had no production caller
```

One commit, base `development`, one item. **Design origin: RSF-F196 v0.9**, DESIGN-CHECK row present, **CARRIES confirmed cold**. TESTING.md row present, and it is the #974 helper-buy check, which this PR makes runnable for the first time.

**One scope statement, so the bar is not read as covering more than it does.** The bar proves the rebuild happens WHEN `registerCustomSprayTypes` runs. What proves `registerCustomSprayTypes` runs in production is the call chain above, which I verified by reading source, not by executing it. No bench in this repo can reach the engine's own registration, as the PR body says. That division is fine and disclosed; it is worth stating because the regression this PR fixes lived exactly in the gap between "the mechanism works given its inputs" and "something in production supplies the inputs".

**Not checked:** in-game behaviour, including the helper-buy row the TESTING entry names; and the engine's real `g_sprayTypeManager` and `FillUnit` chains, which no bar here reaches.

PR #978, opened 2026-09-22
feat(RSF-F196): V7, the application result gate at both callers; observable unit-rule fallbacks

TheCodingDad-TisonK:
V7 of the F196 brief: the resolved result enters the fertilizer application path. `onFertilizerApplied` validates before `applyFertilizer`, returns exactly `false` for a refused product before any side effect, and exactly `true` after a valid pass completes. Both callers gate the fertilizer-specific burn, scorch, boom paint and litre coverage on `result == true`. Crop-protection-only passes are outside the gate. It also carries Bob's #976 MINOR: the unit rule's three passthrough fallbacks are now observable.

**Design-origin: RSF-F196 v0.9 | Disposition: CARRIES** (declared in the handback to Bob; he confirms or raises it in the cold verdict). V7 as written, including the build note's requirement that the boolean be wired at both call sites or the gate is inert. Lane FAST TRACK. One commit.

## V7 is the second fence, and the bar says so

In production R2 zeroes a refused product's dose before the area hook runs, and R3d skips a refused secondary before it is applied. This gate is reached only if a first fence fails. The bar models exactly that: it installs the area hook **without** the R2 append, and hands the soil system a refused table the enumeration's manager does not share, so a refused secondary reaches `applyMulti`. Both fences down is the case V7 exists for.

**The bar's limit, stated (Bob's cold review):** it proves V7 **catches** the state a failed first fence would produce, not that the state is **reachable**. Two managers disagreeing is not reachable in production, since there is one manager; the state it creates, a refused secondary arriving at `applyMulti`, is exactly the state a bug in R3d would create. Mechanism artificial, condition under test real. It says nothing about whether R2 or R3d can actually fail, and nobody should later cite it as evidence that they can.

**Nothing is refused on shipped data**, so nothing changes in play. What this PR can break is a **valid** pass: a wrong return would switch off burn, scorch, boom paint and litre coverage for every fertilizer product. The bar pins that the valid path returns exactly `true` and every effect still runs, and the battery's "valid path returns nothing" mutation is killed. The TESTING row is a non-regression check for the same reason.

## What changed

- **`onFertilizerApplied`**: `self.hookManager:isRefusedProduct(fillTypeIndex)` before anything else, `return false`; `return true` at the end. A system with no manager or no refused table in reach refuses nothing (today's behaviour, and every bench that builds a bare system).
- **Primary caller** (`applySingle`): a `fertResult` in the hook's scope, cleared when a section's application returns anything but `true`; burn requires it; scorch requires it **for a fertilizer product only**; the boom paint and litre coverage block after the section loop requires it for a fertilizer product only.
- **Secondary caller** (`applyMulti`): the same shape with its own `fertResult2`; burn, scorch, boom paint and litre coverage gated for a fertilizer secondary only.
- **The #976 MINOR**: `SoilFertilitySystem.unitRuleFallbacks` (the wrapper's fallback when HookManager is absent), `SoilFertilitySystem.unitRuleUnresolvedNames` (a name the coverage site could not resolve), `SoilHUD.unitRuleFallbacks` (the HUD guard). Read by nothing in production; they exist so a bar can assert they did not move. Bob's ruling to keep the guard rather than hard-depend stands (33 bars load `SoilFertilitySystem` without HookManager).

## Evidence

- `RSF-F196-v7_result_gate_test.lua`, **44 rows**. Group A: the **real** `onFertilizerApplied` with every side effect spied: exactly `false` before `applyFertilizer`, the organic input, the overlay refresh and the broadcast; exactly `true` after all four ran; no manager, or a manager without the refused table, refuses nothing; the result is a boolean on both paths. Groups B and C: the **real** `installSprayerAreaHook` with the real `onFertilizerApplied` over spied effects: a refused driving product is not applied, burned (rate 2.0 is above the risk threshold, so only the gate stops it), scorched, painted, cell-marked or coverage-advanced; the same pass with a valid product does all six; a herbicide-only pass keeps its direct effect, scorch and overlay cells even when the refused table names it; VWW sections agree; a refused secondary past R3d is stopped by V7 while the primary and a valid secondary are untouched.
- `RSF-F196-unit_rule_fallback_observable_test.lua`, **18 rows**: with HookManager present the counters never move through a dry conversion and a resolved name; without it the value is unconverted **and** the count is exactly the two sites reached (the factor and the fully-treated comparison); an unresolved name is counted separately; the HUD guard the same way.
- `mutate_f196_v7.py`: **11 mutations across three shipped files, all KILLED with named rows**, every file restored and proved by sha256.

```
cd tools/test
node syntax-check.mjs   ->  Lua 5.1 syntax OK, 103 files
node lint.mjs           ->  Lint clean, 103 files
node run-tests.mjs      ->  PASS, 0 failed, 110 files (do not gate on the total; om_213 reads another repo's tree)
py mutate_f196_v7.py    ->  exit 0, 11 killed, 3 restores MATCH
py build.py             ->  zip builds, untracked
```

## Not in this PR

Slice 4 wrapper ownership (V12c, V13; which RSF-741 item 9 extends). Slice 5 Drain Vehicle recovery (R7). X1. V19.


sasha-rf (approved):
PASS.

Checked state (OPEN) and got the head SHA fresh, fc16f109, based directly on #977's merge. Reviewed from a fresh worktree.

Read the actual gate rather than the description: `onFertilizerApplied` checks `hm:isRefusedProduct(fillTypeIndex)` first and returns `false` before `applyFertilizer` is even called, so nothing (nutrient buffer, organic compliance, overlay refresh, MP broadcast) runs for a refused product; the tail of the function returns `true` after every side effect completes. Both callers (`fertResult`/`fertResult2`) check `~= true` and gate burn/scorch/paint/coverage accordingly, with `not isFertilizer or fertResult` at each site, confirming crop-protection-only passes bypass the gate entirely rather than being silently swept in.

Suite: syntax/lint clean, PASS, 5751 assertions, 0 failed, 110 files. 5689 (from #977) + 44 + 18 = 5751, exact; file count +2, exact.

Mutation battery: `mutate_f196_v7.py`, all 11 (V7-refusal-check-removed, V7-returns-nil-on-refusal, V7-valid-path-returns-nothing, V7-primary-caller-ignores-result, V7-primary-scorch-ungated, V7-primary-scorch-gates-herbicide-too, V7-primary-paint-and-coverage-ungated, V7-secondary-caller-ignores-result, OBS-system-fallback-silent, OBS-unresolved-name-silent, OBS-hud-fallback-silent) KILLED with named rows, all 3 touched files restored and sha256-verified.

The "gates herbicide too" mutation specifically proving the crop-protection exemption is real, not assumed, is the one I'd have gone looking for myself if it weren't already there.

Approved as sasha-rf.


TheCodingDad-TisonK:
## Bob, cold diff review

Read at `fc16f109`, base `origin/development` (the merged #977). One commit, three source files, two new bars, one battery.

**Disposition CARRIES is CONFIRMED**, checked against V7's clause text rather than the row's account of it. V7 says the function "returns exactly `false` before nutrient buffer, agronomy, organic input, overlay refresh or multiplayer broadcast on refusal, and returns exactly `true` after valid side effects complete". The refusal return sits above `applyFertilizer` (which owns the buffer and agronomy), above the organic input call, above the overlay refresh and above the broadcast. All five, in order, below the fence.

### The property that matters most, checked directly

`onFertilizerApplied` has **exactly two exits**: `return false` at the refusal check and `return true` as the last statement. There is no intermediate return anywhere in the body. That is what makes "a valid pass returns true" a structural fact rather than a hope, and it is the one thing that had to be right, because a nil on the valid path would silently switch off four effects for every fertilizer product.

**The six gate sites are consistent, and the one asymmetry is justified rather than an oversight.** Five use the `not isX or result` form. `applySingle`'s burn uses bare `fertResult`, and that is correct because it is already predicated on `entry and (entry.N or entry.P or entry.K)`, so only a product with an NPK profile reaches it; `applyMulti`'s burn has no such precondition and therefore does need `not isFert2 or fertResult2`. Different guards because different preconditions, not an inconsistency.

**Scope checked, because a misplaced declaration here would be catastrophic and silent.** `fertResult` and `local function applySingle` and the post-loop block all sit at the same indent in the same closure; `fertResult2`, `applyMulti` and its post-loop block likewise. Had the declaration been nested one level deeper, the post-loop read would have seen nil, `not isFertilizer or nil` would be false for every fertilizer, and boom paint plus litre coverage would have been off for every fertilizer pass. It is not. The bar covers both directions independently: the ungated mutation is killed, and the valid-path rows pin all six effects on.

### Verification

```
node run-tests.mjs      PASS 5751 assertions, 0 failed, 110 files
  RSF-F196-v7_result_gate_test.lua                 44 / 0
  RSF-F196-unit_rule_fallback_observable_test.lua  18 / 0
py mutate_f196_v7.py  11 of 11 KILLED with named rows, three restores proved by sha256
```

The battery reaches the two mutations that matter most: `V7-valid-path-returns-nothing` and `V7-primary-scorch-gates-herbicide-too`. The second is the one a careful author would most plausibly have written by accident, and the herbicide row kills it.

### On taking both first fences down, which you asked about

**It is legitimate, and it is the only way to test a second fence.** Leaving R2 and R3d in place would give a bar that passes whether or not V7 exists at all, which is the "green because the fixture exits early" failure this office has hit repeatedly.

The distinction worth drawing is between a fixture that supplies an impossible world and one that produces, by an artificial route, the state a failed fence would produce. Two managers disagreeing is not a state production can reach, since there is one manager. But the state it creates, a refused secondary arriving at `applyMulti`, is exactly the state a bug in R3d would create. So the mechanism is artificial and the condition under test is real, which is correct defence-in-depth testing.

The limit that follows, and it belongs in the body: this bar proves V7 CATCHES that state, not that the state is reachable. It says nothing about whether R2 or R3d can actually fail, and it should not be read as evidence that they can.

### Findings

`src/SoilFertilitySystem.lua:59`: **MINOR**: the counter's unit is wrapper invocations, not applications, and that is the one non-obvious thing about a field whose entire purpose is to be asserted on. One `applyFertilizer` call increments it TWICE, at the nutrient factor and at the fully-treated comparison, which you found and stated in the handback. For the intended use, asserting the branch was not taken, the unit does not matter. For anyone later asserting a per-pass count it does, and they would be wrong by a factor of two with no way to see it from the declaration. fix: one clause in the existing comment saying it counts wrapper calls and that one application makes two.

### Verdict

**PASS.**

```
fc16f109  feat(RSF-F196): V7, the application result gate at both callers; observable unit-rule fallbacks
```

One commit, base `development`, one item. **Design origin: RSF-F196 v0.9**, DESIGN-CHECK row present, **CARRIES confirmed cold**. TESTING.md row present on tracking main.

The #976 MINOR is discharged in the shape agreed: the fallback is kept, the branch is counted at all three sites, nothing in production reads the counters, and all three `OBS-` mutations are killed, so the tripwires are themselves load-bearing rather than decorative.

**Not checked:** in-game behaviour; whether R2 or R3d can fail in production, which this bar deliberately does not address; and the accumulation behaviour of the three counters across a long session, which is harmless while nothing reads them.

PR #979, opened 2026-09-22
fix(SF-79): route the plow's pH normalisation through the positional writer

TheCodingDad-TisonK:
`onPlowing` computed a pH normalisation toward 7.0 and wrote `field.pH` **directly**. The block had no positional call anywhere in it. `_phRefreshScalar` republishes `field.pH` from the map-derived report after every application write, so the **next spray pass overwrote the scalar from a report that never saw the plow**: the plow's pH effect was discarded. This routes the plow through the positional writer, the way the meadow normaliser already is, and keeps the brief's number.

**Design-origin: SF-79 v1.0 | Disposition: CARRIES** (declared in the handback to Bob; he confirms or raises it in the cold verdict). Section C: "Route physical lime/gypsum, normalization, leach/rain/irrigation, **plow**, burn/scorch, reroll/admin/recovery through the writer" and "Plow targets 7.0 at existing intensity `0.1 * clampedAcceptedAreaHa / footprintAreaHa`". The number was implemented; the routing was not. Lane FAST TRACK (Tyson approved Bob's SF-79 sweep split; this is item 1's first entry, the only one a player is currently losing something to). One commit.

## What changed

One block in `onPlowing`, the shape of the meadow normaliser at the daily drift:

```lua
if type(self._phApplyField) == "function" then
    self:_phApplyField(fieldId, PositionalPH.OP_NORMALIZE, phNormalization, phTarget, phTarget, 'plow')
    if (field.pH or phBefore) ~= phBefore then changed = true end
else
    -- today's scalar fallback, unchanged, for a system without the writer
end
```

- **Magnitude carried, not re-derived**: `phNormalization = 0.1 * factor` with `factor = clampedArea / fieldAreaHa`, and `phTarget = 7.0`, exactly as before.
- **Equal bounds** are the brief's "plow's equal 7.0 bounds" (section A). The writer's NORMALIZE handles a band of one value: pixels below step up, above step down, at 7.0 untouched.
- **Source token** `'plow'`, its own, so the writer's attribution and the banked remainder are the plow's.
- **The writer banks a sub-step amount** (brief 3.B), so a small accepted area is not dropped at the write; `changed` is set when the republished scalar moved.

**Bob's cold-review finding on the shared bank, recorded rather than folded in.** The bank this PR deposits into does not keep remainders separate by cause or kind: `_phAddPending` stores both, `_phTakePending` filters on domain alone, and `_phApplyField` sums every taken remainder into one total applied under whichever operation runs next. So a plow NORMALIZE remainder (a magnitude meaning "toward 7.0") can be taken by the next rain DELTA and applied as a signed change, the wrong direction for a field above 7.0. Bounded under one raw step (about 0.0098 pH) per mixing event, because a remainder is sub-step by construction, which is why it is a MAJOR and not a blocker. The brief anticipated it: "Sub-step scheduled FIELD remainders stay SEPARATE PER CAUSE/KIND/DOMAIN"; storage complies, use ignores it. A partial plow's remainder is usually sub-step, so this PR makes the plow a frequent depositor into that bank without being wrong itself. The fix (filter the take on cause and kind too) is a change to the shared bank and its own item, in MAINTENANCE for Desk to sequence.
- The existing daily area cap and the `plowingBonus` gate are untouched and still gate the call.

## What a player will see

Plow a field with pH away from 7.0, note the pH, then spray it. Before this PR the plow's change vanished on the spray. After it, the plow's change survives, because it lives in the map the spray's report reads.

## Evidence

- `SF-79-plow_ph_routing_test.lua`, **22 rows**. (A) the **real** `onPlowing` with the writer spied: called once, for the field, as a NORMALIZE, at `0.1 * 0.25` for half a hectare on a 2 ha field, toward equal 7.0 bounds, with source `plow`; the scalar no longer written directly; the daily area cap and `plowingBonus` still gate it. (B) the **real** `onPlowing` without the writer: today's fallback from both sides of 7 and at the clamp. (C) the consequence with the **real** `_phApplyField` and `_phRefreshScalar` and only the map stubbed: a direct scalar write is **erased** by the next refresh (the defect); a routed plow moves the scalar through the report and **survives** the next refresh.
- `mutate_sf79_plow.py`: **7 mutations, all KILLED with named rows**, restore proved by sha256: routing removed, intensity doubled, target not 7.0, bounds not equal, source token borrowed, DELTA instead of NORMALIZE, fallback removed.

```
cd tools/test
node syntax-check.mjs    ->  Lua 5.1 syntax OK, 103 files
node lint.mjs            ->  Lint clean, 103 files
node run-tests.mjs       ->  PASS, 0 failed, 109 files (do not gate on the total; om_213 reads another repo's tree)
py mutate_sf79_plow.py   ->  exit 0, 7 killed, restore MATCHES
py build.py              ->  zip builds, untracked
```

**What the bar cannot see:** the real map write (the footprint is a stub that moves one value), the polygon domain, and the in-game report. The TESTING row is the plow-then-spray check.

## Not in this PR

The other SF-79 item 1 entries (AUTO's two defects, `isTrueLime`, the write-only schema key) follow separately. Bob's #978 MINOR (one clause on the unit of the fallback counter) rides the next commit on this branch after #978 merges, since the counter does not exist on this base.


sasha-rf (approved):
PASS.

Checked state (OPEN) and got the head SHA fresh, 38788aba, based on the current true development tip (#977, c8e00edb; #978 hasn't merged yet despite my earlier approval, just a timing note, not a problem). Reviewed from a fresh worktree.

Verified the actual defect mechanism, not just the description: `_phRefreshScalar` (PositionalPH.lua:418) unconditionally sets `field.pH = rep.value` from the map-derived report whenever it runs, and it's called after every spray write. So a direct `field.pH = phAfter` scalar write (the old plow code) that never touched the map genuinely would get overwritten on the next spray, since the report the refresh reads from never learned about the plow's change. The fix routes through `_phApplyField` with `PositionalPH.OP_NORMALIZE`, own source token `'plow'`, same magnitude (`0.1 * factor`, target 7.0) as before, fallback preserved for a system without the writer. Read the actual block, matches the body exactly.

Suite: syntax/lint clean, PASS, 5711 assertions, 0 failed, 109 files. 5689 (from #977) + 22 = 5711, exact; file count +1, exact.

Mutation battery: `mutate_sf79_plow.py`, all 7 (routing-removed-direct-write-restored, intensity-doubled, target-not-seven, bounds-not-equal, source-token-borrowed, delta-instead-of-normalize, fallback-removed) KILLED with named rows, restore sha256-verified.

Design-origin SF-79 v1.0, disposition CARRIES declared, Bob's cold review confirms or raises separately. This is the item a player is currently losing something to, so worth saying plainly: the fix is real, the mechanism that erases the old write is real, and the routing genuinely closes the gap rather than just adding a parallel path.

Approved as sasha-rf.


TheCodingDad-TisonK:
## Bob, cold diff review

Read at `38788aba`, base `origin/development` `c8e00edb`. One commit, one source hunk, one new bar, one battery. This repairs the defect my SF-79 sweep found, so I have read it against the sweep's own evidence rather than against my memory of it.

**Disposition CARRIES is CONFIRMED.** Section C names the plow among the writers to route and gives the magnitude; section A gives the equal 7.0 bounds. The number was already implemented and the routing was not, which is what the sweep found. The hunk carries the magnitude unchanged (`0.1 * factor` toward `7.0`) and adds the routing in the meadow normaliser's exact shape: writer first, legacy scalar only in the `else` when `_phApplyField` is absent.

### Verification

```
node run-tests.mjs      PASS 5711 assertions, 0 failed, 109 files
  SF-79-plow_ph_routing_test.lua   22 / 0
py mutate_sf79_plow.py  7 of 7 KILLED with named rows, restore proved
```

`bounds-not-equal`, `target-not-seven` and `delta-instead-of-normalize` are the three that matter, and all three are killed. The C group's shape is right: it shows a direct scalar write erased by the next refresh and the routed plow surviving one, which is the defect and its repair demonstrated against each other rather than asserted.

### Your question (a): the equal-bounds arithmetic reads as you read it

I worked it rather than agreeing. With `lowRaw == highRaw == R`, the NORMALIZE branch runs four bands (`PositionalPH.lua:386-389`):

```
runSet   [R-step+1, R-1]  -> R      pixels below, within one step: onto the bound
runDelta [RMIN,     R-step] +step   pixels further below: stepped up
runSet   [R+1,  R+step-1]  -> R     pixels above, within one step: onto the bound
runDelta [R+step,   RMAX]  -step    pixels further above: stepped down
```

**A pixel at exactly R falls in no band**, because band 1 ends at `R-1`, band 2 ends at `R-step`, band 3 starts at `R+1` and band 4 starts at `R+step`. So pixels at 7.0 are untouched, which is what you said. The bands also do not overlap, so no pixel is processed twice, which is what section A's "near-target SET before farther ADD" asks for. The reference model `normalizeRawCohorts` at `:125-142` builds the same four bands, so model and production agree.

### Findings

`src/PositionalPH.lua:640` and `:442-444`: **MAJOR**: the sub-step bank does not keep remainders separate by cause or kind, so this PR's plow remainder can be flushed as somebody else's DELTA. **This is pre-existing and not introduced here**, but the PR body states the flush behaviour as correct ("any later NORMALIZE or DELTA on that domain flushes it"), and it is not, so it should not be recorded as intended.

`_phAddPending` stores `cause` and `kind` (`:626-631`). `_phTakePending` filters on `domainKey` ALONE (`:640`). `_phApplyField` then sums every taken remainder into one `total` (`:442-444`) and applies it under whichever `operation` is running now. So a plow NORMALIZE remainder, whose magnitude means "move toward 7.0", can be added to a rain DELTA and applied as a signed pH change. For a field ABOVE 7.0 the plow wanted the value to fall; its banked positive magnitude makes the rain DELTA less negative, which pushes the value up. Wrong direction, silently.

The brief anticipated exactly this: "Sub-step scheduled FIELD remainders stay SEPARATE PER CAUSE/KIND/DOMAIN". The implementation complies in storage and ignores it in use. `cause` and `kind` are written to XML at `:663-664` and projected to StateLedger at `SoilFertilitySystem.lua:8238` and `:8374`, and **no decision anywhere reads either field**. That is the same shape as three other findings this week: the machinery to comply exists, is persisted faithfully, and nothing consults it.

Bounded, so not a blocker: a remainder is sub-step by construction, under 0.0098 pH, so each mixing event is wrong by less than one raw step.

**Why it belongs on this PR even though the PR did not cause it:** a partial plow's `0.1 * factor` is usually sub-step, so this change makes the plow a frequent new depositor into a bank that mixes kinds, and the daily rain DELTA is a frequent withdrawer. The exposure goes up materially even though the code here is correct. fix: filter `_phTakePending` on cause and kind as well as domain, which is a change to the shared bank and its own item, not this one.

Nothing else. No BLOCKER, no MINOR.

### Verdict

**PASS with MAJORs.** The hunk is correct, matches the sibling it was told to match, and the MAJOR is in shared machinery it merely uses more often. Nothing here blocks the merge.

```
38788aba  fix(SF-79): route the plow's pH normalisation through the positional writer
```

One commit, base `development`, one item. **Design origin: SF-79**, DESIGN-CHECK row owed with CARRIES, **confirmed cold here**.

One note on sequencing rather than a finding: this branches from `c8e00edb`, the same base as #978, so the two are siblings rather than a chain. They touch `SoilFertilitySystem.lua` in different functions, `onFertilizerApplied` near `:926` and `onPlowing` near `:1401`, so they should merge without conflict in either order. Deferring the #978 counter clause to the first commit that touches the wrapper after #978 merges is the right call; it cannot land here, because the counter does not exist on this base.

**Not checked:** in-game behaviour; the real pH map, which the C group stubs and says so; and whether any remainder currently sits in a save from a pre-#979 field, which would be a savegame question rather than a source one.

PR #936, opened 2026-09-13
fix(hud): resolve the applied product through attached implements (#809)

TheCodingDad-TisonK:
## What
Fixes #809: Auto rate stays at 1.00x with a tractor plus pulled sprayer or spreader.

`getApplicatorVehicle` returns the root vehicle (the tractor), and `updateAutoRates` passes it to `SoilHUD:getSprayerFillType`. With no spray work area live and no tanker spec on the tractor, the Priority 4 fallback walked only the tractor's own fill units and took the first non-empty one: the fuel tank. The rate calc read DIESEL, found no `FERTILIZER_PROFILES` entry, and held 1.0x (the "DIESEL NOT in FERTILIZER_PROFILES - holding 1.0x" log line in the issue).

This ports **only** the Priority 4 hunk from c776536b (`src/ui/SoilHUD.lua`), byte for byte (the hunk diffs identical against `git show c776536b:src/ui/SoilHUD.lua` after CRLF normalisation):
- a `NON_PRODUCT` set (DIESEL, DEF, ADBLUE, AIR, ELECTRICCHARGE, METHANE) that is never treated as an applied product
- an attached-implement walk via `getAttachedImplements()` (AttacherJoints.lua:1005, entries carry `.object`) to depth 3, plus the `rootVehicle` and its implements
- a two-pass choice: a unit holding a product with a `FERTILIZER_PROFILES` entry wins, otherwise the first other non-fuel product

Not carried, per the Design containment ruling: the wider Auto algorithm rewrite (FIX-8 under-machine sensing) and the MULT_MIN retune. `SoilUtils.resolveSprayerFillTypeIndex` (#780) is untouched; the only source file changed is `src/ui/SoilHUD.lua`.

## Evidence
- New bench `tools/test/lua/attached_product_resolution_809_test.lua` (loads the real `src/ui/SoilHUD.lua` against fengari with vanilla-shaped FillUnit/AttacherJoints mocks). Locks: tractor+spreader resolves LIME, tractor+trailed sprayer resolves LIQUIDFERTILIZER, tractor alone with diesel+DEF resolves nil, self-propelled sprayer resolves its own product tank, empty implement falls back to the root's product, profiled product beats an unprofiled one regardless of order, unprofiled non-fuel still resolves, depth 3 found / depth 4 not found, throwing `getAttachedImplements` fails safe, and the #708/#780 physical-tank rule still wins ahead of the fallback.
[...]

PR #980, opened 2026-09-22
fix(SF-79): the pH AUTO whole-rate, three defects: the index, the coordinate, the caller

TheCodingDad-TisonK:
AUTO had never changed a rate. SF-79 section D specifies the pH AUTO whole-rate, and the curve, the cache and the gates were all in place. Three defects kept the factor at 1.0 on every real path, each hiding the next. This fixes all three and changes no target, curve, limit or cadence.

**Design-origin: SF-79 v1.0 | Disposition: CARRIES** against section D (declared in the handback to Bob; he confirms or raises it in the cold verdict against both briefs). Section D: "Use current nodes/section state, never the previous interval's", "Keep autoRateControl/getAutoMode, existing targets/curves/limits and the existing five-second cadence", "An unavailable required sample retains the selected rate", "Apply the whole multiplier once". Lane FAST TRACK, SF-79 item 1 of Tyson's approved split. One commit.

**Who found what, and the register row this touches.** Defects 1 and 2 are **Iris's**, found 2026-09-13 and recorded in DESIGN-CHECK row 61 (#936, SF-73 auto-rate lane, 163a8d9): "treats native's numeric `sprayFillType` index as a table inside a swallowed `pcall`, then reads the previous spray coordinates". Bob's intake today re-derived them. Defect 3, the caller, is new today. Row 61 is WITHHELD under SF-73, and its own text separates two halves: a mechanical instruction, "**repair what is there, do not write it fresh**", which is exactly what this PR does; and a design half, "no SDS mechanism, curve value or UI contract is approved yet", which the WITHHELD protects and which this PR does not touch (row 62, the `MULT_MIN` retune, stays WITHHELD untouched). One of the row's two gates cleared today: the "Field application and machine work areas" knowledge pack landed on tracking main at a8adde7; no Stage2 record for SF-73 was found, so that half is unverified rather than open. **This PR opened on Tyson's ruling**, and row 61's Result records the mechanical half repaired with the design half still withheld.

## The three defects

| | where | what | why it hid the next |
|---|---|---|---|
| 1 | `PositionalPH.lua`, `updatePHWorkAuto` | `workAreaParameters.sprayFillType` is a fill-type **index** (`Sprayer.lua:926`, compared to `FillType.UNKNOWN` at `:316`, `:922`); the function tested `.name` on it and raised inside the caller's `pcall` | with the factor stuck at 1.0, nothing downstream was ever exercised |
| 2 | same function | the only coordinate was `_lastSprayX/Z`, written by the **area** hook, which runs **after** this start hook in the same tick: the previous tick's point, or nil on the first; the passed work areas were ignored | fixing 1 alone ships a working AUTO reading the wrong ground |
| 3 | `HookManager.lua`, the rate-multiplier append | passed `spec.workArea and spec.workArea.workAreas` with `spec = spec_sprayer`; the engine's sprayer spec has **no** `workArea` member (none in the decompiled `Sprayer.lua`; the list is `spec_workArea.workAreas`, `WorkArea.lua:74`), so the caller handed AUTO nil on every real vehicle | with 1 and 2 fixed, AUTO would still have retained 1.0 for want of a list; the bar that drives the real start hook is what caught it |

## What changed

- **Defect 1**: the descriptor is resolved through `g_fillTypeManager:getFillTypeByIndex`, as every index consumer in HookManager does. UNKNOWN, a non-number and an unresolvable index retain the selected rate.
- **Defect 2**: `PositionalPH.sampleWorkAreasPH(soilSys, vehicle, workAreas)`: the mean local pH under each active, non-auxiliary work area's centre. The engine's work area is the parallelogram start / width / height (`WorkArea.lua:289-291`), whose centre is the midpoint of the width and height nodes, read through `getWorldTranslation` **now**. `getIsWorkAreaActive` (`Sprayer.lua:726`) supplies current section state. No area, no node or no numeric sample retains the selected rate. The function **no longer reads `_lastSprayX/Z` at all**; nothing falls back to the stale point.
- **Defect 3**: the appended function takes the third argument the engine raises the start event with (`WorkArea.lua:126` passes `spec.workAreas`; `Utils.appendedFunction` forwards every argument) and falls back to the vehicle's own `spec_workArea.workAreas`.
- **Unchanged**, as section D requires: the server gate (pure-client AUTO is display-only), `getAutoMode`, the curve and its limits, the five-second cache per working vehicle, and the multiplier applied once together with the rate manager's.
- **Also carried**, on Desk's ruling that it rides the next SoilFertilizer commit: Bob's #978 MINOR, one clause on the unit-rule fallback counter's comment.

## What a player will see

With AUTO on and a pH product loaded, the applied rate now follows the ground under the boom: boosted (up to 1.5x) where the local pH is below the neutral band for lime, reduced (0.5x) where it is above; the reverse for gypsum. Before this PR AUTO applied the selected rate everywhere. **If the coordinate were stale rather than current** (defect 2 alone), the rate would follow the ground the boom was over on the previous tick, which at driving speed is a few centimetres behind, and on the first tick of a pass nothing at all; the TESTING row says what that looks like.

## Evidence

- `SF-79-ph_auto_whole_rate_test.lua`, **38 rows**. The engine boundary (`getWorldTranslation`) is a node-to-coordinate map, because a bench has no scene graph; what the rows pin is **where the input comes from**. (A) a numeric index resolves and the factor moves; UNKNOWN, an unresolvable index, a product without pH and a descriptor table are neutral. (B) the stale point planted over different ground from the areas; the stale point absent; the sprayer's own list over different ground from the argument; no list; an empty list; auxiliary and inactive areas skipped; the mean over two areas; unrecorded ground under every area; a missing node; a vehicle without the activity method. (C) the existing curve and limits from both sides of the band for LIME and GYPSUM. (D) the cache within and after five seconds, per vehicle. (E) the gates. (F) the **real** rate-multiplier append: the dose multiplied by the factor once, not compounding across frames; the raised list winning over the vehicle's own; the engine's real spec shape (no `spec_sprayer.workArea`) using the vehicle's own list.
- `mutate_sf79_auto.py`: **10 mutations across two shipped files, all KILLED with named rows**, restores proved by sha256: each of the three defects put back, the caller discarding the raised list, the sample from the sprayer's own list, a stale-point fallback, auxiliary counted, inactive counted, the cache dropped, the curve altered.

```
cd tools/test
node syntax-check.mjs    ->  Lua 5.1 syntax OK, 103 files
node lint.mjs            ->  Lint clean, 103 files
node run-tests.mjs       ->  PASS, 0 failed, 112 files (do not gate on the total; om_213 reads another repo's tree)
py mutate_sf79_auto.py   ->  exit 0, 10 killed, 2 restores MATCH
py build.py              ->  zip builds, untracked
```

**The bar that starts from production's own entry point** (Tyson's standing rule, 2026-09-22): group F of `SF-79-ph_auto_whole_rate_test.lua`. It installs the real rate-multiplier append through `HookManager.installSprayerStartHook` and drives it from `Sprayer.onStartWorkAreaProcessing` the way the engine raises it, with no hand-populated coordinate or spec; the sprayer carries only what the engine's spec shape carries. That group is what found defect 3.

**Two readings, stated as decisions rather than quotations (added after the merge on Bob's cold review, and said to be a post-merge edit).** The mean over each active area's centre is a *reading* of section D's "use current nodes/section state": the clause does not say how several areas reduce to one factor; the mean of the areas the boom actually covers was chosen over one read at the mean centre, which can fall between sections and be sprayed by neither. And the section state sampled is **native's at append time**: the mod's own section hooks (section control, see-and-spray) are installed after this append and may deactivate sections it has already counted, so "current" here means native's current, not the mod's final. Both are in MAINTENANCE with Bob's two MINORs.

**What the bar cannot see:** the scene graph (node positions are a fixture map), the real value-map read, and the rate manager's own multiplier. The in-game check is AUTO on a field whose pH sits away from the band.

## Not in this PR

The other SF-79 item 1 entries (`isTrueLime`, the write-only schema key) and the sub-step bank mixing kinds (Bob's #979 MAJOR, in MAINTENANCE for Desk to sequence). No curve value moves: the #936 rows on `MULT_MIN` and the curve retune stay WITHHELD with SF-73.


sasha-rf (approved):
PASS.

Checked state (OPEN) and got the head SHA fresh, 5e0d4733, based on the current true development tip (#979). Reviewed from a fresh worktree.

Verified all three fixes against the actual code, not the description:
- **Defect 1**: `sprayFillType` is resolved through `g_fillTypeManager:getFillTypeByIndex` with a type/UNKNOWN guard before it, rather than indexing it for `.name` directly.
- **Defect 2**: read the whole rewritten function; `_lastSprayX`/`_lastSprayZ` appear exactly once in it now, in the explanatory comment, no longer as an actual read. The function takes `workAreas` and samples current node positions via `PositionalPH.sampleWorkAreasPH`.
- **Defect 3**: the caller now prefers the raised `workAreas` argument and falls back to `self.spec_workArea.workAreas`, not the old `spec.workArea` (no such field). Confirmed `spec_workArea` is the real decompiled engine field name directly against `WorkArea.lua` myself (`self.spec_workArea` used throughout, `:296` for one example) rather than taking the citation on trust.

Suite: syntax/lint clean, PASS, 5811 assertions, 0 failed, 112 files. 5751 (from #978) + 22 (from #979) + 38 = 5811, exact; file count +2, exact.

Mutation battery: `mutate_sf79_auto.py`, all 10 (defect1-index-read-as-descriptor, defect2-stale-point-restored, defect3-caller-passes-nil, caller-ignores-the-raised-list, sample-from-sprayers-own-list, stale-point-as-fallback, auxiliary-not-skipped, inactive-sections-counted, cache-dropped, curve-altered-above-band) KILLED with named rows, both touched files restored and sha256-verified.

Design-origin SF-79 v1.0, disposition CARRIES against section D, correctly scoped: the mechanical repair only, the WITHHELD design half (row 61's gate, row 62's MULT_MIN retune) untouched, matches the PR's own account of what it does and doesn't touch.

Separately, and not a reason to hold this: I'm mid-tracing a real player report (Wizard, via Sunflower) of a pH read-path discrepancy elsewhere in this same subsystem, handful vs HUD/Esc-page disagreeing on the same spot. Already flagged to Desk and routed to Bob rather than sat on. AUTO here reads through the same positional path (`sampleWorkAreasPH`), so if Bob's trace finds something that touches this PR, it'll land on the review; per Desk, that's not a reason to withhold approval now.

Approved as sasha-rf.


TheCodingDad-TisonK:
## Bob, cold diff review

Read at `5e0d4733`, base `origin/development` `3c118efb`. One commit, three source files, one new bar, one battery. Reviewed against **both** briefs, as agreed: SF-79 section D, and DESIGN-CHECK rows 61 and 62.

**Disposition CARRIES against SF-79 section D is CONFIRMED**, and the SF-73 claim is confirmed too, which here is a claim about what does NOT move. I checked the register before the code, per the rule adopted today: rows 61 and 62 in DESIGN-CHECK, and the bank and counter-clause rows in MAINTENANCE.

- Row 61's mechanical half named two defects on 2026-09-13, "treats native's numeric `sprayFillType` index as a table inside a swallowed `pcall`, then reads the previous spray coordinates" (Iris, `163a8d9`). Both are repaired here. Its instruction was "repair what is there, do not write it fresh", and that is what this is.
- Row 61's design half and row 62 are untouched: no target, curve value, limit, cadence, mechanism or UI contract moves. I verified `MULT_MIN` and the boost/reduce caps are unchanged, and the battery's `curve-altered-above-band` mutation is killed, so the curve is pinned rather than merely unedited.
- **Attribution is right in the body**: defects 1 and 2 are Iris's, re-derived by my intake nine days later; defect 3 is new today. My intake missed row 61 entirely, which is recorded.

### Verification

```
node run-tests.mjs      PASS 5811 assertions, 0 failed, 112 files
  SF-79-ph_auto_whole_rate_test.lua   38 / 0
py mutate_sf79_auto.py  10 of 10 KILLED with named rows, restores proved
```

**The entry-point bar is named and it is the reason this PR is right:** group F drives the real start hook, and it is what found defect 3. Without it the other two fixes would have shipped and changed nothing, because the caller handed AUTO nil on every real vehicle. That is the new standing rule earning its place on its first PR.

### Your question (a): the centre is right, confirmed two ways

`(width + height) / 2` is correct for the engine's three-node parallelogram. Algebraically the fourth corner is `width + height - start`, so the centre of the four is `(start + width + height + (width + height - start)) / 4`, which collapses to `(width + height) / 2`, and `start` drops out exactly as you said. I also checked it against a known square: `writeValueAtWorld` with radius `r` builds start `(x-r, z-r)`, width `(x+r, z-r)`, height `(x-r, z+r)`, and your formula returns `(x, z)`, its true centre.

**Question (b):** the timing holds. `WorkArea:onUpdateTick` raises the start event at `:126` and calls `getIsWorkAreaActive` for its own loop at `:143`, so the state your append reads is this tick's, and `Sprayer` overrides the predicate at `:726` as you say.

**Question (c): your reading is the better one and I would keep it.** Section D says to use current section state and does not specify how to reduce several areas to one factor. For a whole-rate factor on a sectioned boom over a gradient, the mean of each active area's centre represents the ground the boom actually covers; one read at the mean centre represents a point that may lie between sections and be sprayed by neither. Worth saying in the body that this is a reading of section D rather than a quotation of it, so a later seat knows it is a decision and not a transcription.

### Findings

`src/PositionalPH.lua:3`: **MINOR**: `local aux = (WorkAreaType ~= nil) and WorkAreaType.AUXILIARY or nil`, and a nil `aux` makes every area usable, so a missing global silently INCLUDES auxiliary areas in the mean. `WorkAreaType` is an engine global used bare by the engine itself (`WorkArea.lua:141`, `:240`, `Mower.lua:417`, `:481`), so it cannot be absent in production. That makes this the same shape you declined twice today for `FillType`: guarding production against a missing global that has no production case, where the guard's failure mode is a wrong answer rather than a loud one. fix: either read it bare and let a bench supply it, as you did for the `FillType` fixtures, or keep the guard and count a skip the way the unit-rule fallbacks now count theirs.

`src/PositionalPH.lua:7-10`: **MINOR**: the section state sampled is native's at append time, not the state after the mod's own section hooks. `installAll` installs the start hook at its relative line 138 and `installSectionControlHook` and `installSeeAndSprayHook` at 217 and 220, so those run after this append and may deactivate sections it has already counted. For a whole-rate factor computed before the multiplier that is defensible and probably right, but "current section state" then means native's current, not final, and only the body can say so.

### The reported HUD/handful discrepancy, since it lands in this subsystem

Desk asked for this on the PR. **#980 is not affected as a defect, and Desk's expectation is confirmed: AUTO acts correctly while the HUD says unknown.** `sampleWorkAreasPH` reads through `valueMaps:readValueAtWorld`, the positional path, exactly as row 61's own text requires. It never consults the aggregate the HUD and the Esc page use.

**That is a tester-confusing state and this PR's TESTING row should name it**: in AUTO on such a field, the rate will visibly respond while the pH readouts say unknown, and a tester who assumes the readout reflects what AUTO sees will report the rate as wrong. The row should say that the HUD reading unknown does not mean AUTO is reading nothing.

The mechanism, in short, with the full trace going to Sasha: the handful is a positional read and returns a value if and only if that pixel is written (`decode` returns nil for raw 0, so a real handful value does prove a written pixel, which was the premise worth checking). The aggregate requires the farmland's field polygons to resolve and counts only pixels inside them. Those can disagree, and the strongest sub-case is the polygons not resolving at all, which empties the report regardless of what is written.

### Verdict

**PASS.**

```
5e0d4733  fix(SF-79): the pH AUTO whole-rate, three defects: the index, the coordinate, the caller
```

One commit, base `development`, one item. **Design origin: SF-79 section D**, DESIGN-CHECK row present, **CARRIES confirmed cold**; row 61 stays open with its design half WITHHELD under SF-73, and row 62 is untouched. Opened on Tyson's ruling, which the body records.

**Not checked:** in-game behaviour; whether AUTO's restored factor is agronomically right, which is the curve question row 62 still owns; and Stage 2 of row 61's gate, which I could not find a record of and did not claim.

Issue #437, opened 2026-05-27
Ph/OM Amendment lock out period


Arissani:
### What problem does this solve?

 i have no idea why FS designed it this way but never in the history of the world has anyone put lime directly on crops

### Proposed Solution

 lock out all Ph and OM soil amendment while crops are growing, i have no idea why FS designed it this way but never in the history of the world has anyone put lime directly on crops nor compost or manure, rain also plays a role with lime,  we apply lime in light rain it actually helps, but never in a heavy rain just washes in to clumps, this also adds some actual tension factor, like having to make sure you plan your soil amendment properly or you'll miss out on either planting to adjust Ph/OM, or miss adjusting Ph/OM to plant, makes a nice gameplay choice to the player to navigate, wont happen to often either since Ph is relatively slow to leach

### Alternatives Considered

cause the crop to "burn " causing - 80% yield if you put lime on a growing crop, -20% if you apply OM to a growing crop, meaning if you wrongly apply lime and OM you can burn/ruin the entire crop -100% yield 

### Which part of the mod does this relate to?

Fertilizer & Organic Matter

TheCodingDad-TisonK:
Good suggestion — a lockout period after pH/OM amendment would add realism. This is noted as a planned enhancement. The main design question is whether the lockout should be calendar-days or in-game-days, and whether it should block application entirely or just show a warning. Will circle back on this for a future version.

TheCodingDad-TisonK:
Implemented in development branch (commit 66b4573). LIME/LIQUIDLIME applied to a growing crop now flags an -80% yield penalty at harvest; OM amendments (manure, compost, digestate, etc.) flag -20%. A blinking warning notification fires on first application per crop cycle. The penalty is tracked in field save data so it survives a session reload.

PR #981, opened 2026-09-22
feat(SF-79): isTrueLime, the native-credit classification on the footprint request

TheCodingDad-TisonK:
## What

SF-79 section 3.A lists `isTrueLime` among the footprint request fields: "diagnostic classification, not permission or a request to fabricate native credit". Before this PR nothing set it and nothing read it (zero hits in the repo at d7d5cd99).

`applyFertilizer` now classifies once per tick from the engine's own axis, `sprayType.isLime` through `g_sprayTypeManager:getSprayTypeByFillTypeIndex(fillTypeIndex)`. That is the flag `FSDensityMapUtil.updateSprayArea` tests before it calls `updateLimeArea` (decompiled `utils/FSDensityMapUtil.lua:1296-1297`; `misc/SprayTypeManager.lua:62` sets it only for a spray type registered as LIME). A missing spray type is false. The value rides the dose record, which resets on a product change, so the boom strip request carries the same product's answer as the narrow-tool dot. The existing 1000 L milestone line, the one diagnostic reader, now ends in `trueLime=<bool>`; its frequency is unchanged.

**The writer stays blind.** `PositionalPH.lua` gains a four-line comment naming the field and no code. The pH-sign classifiers at the #437 burn gate (`SoilFertilitySystem.lua:5795`) and the VR section curve (`HookManager.lua:2126`) are untouched: the brief keeps them (3.D), and switching them to `isTrueLime` would turn the diagnostic into permission. On shipped profiles the two axes agree (LIME +0.16 and LIQUIDLIME +1.07 are LIME spray types; GYPSUM at -0.10 is FERTILIZER-typed) and remain separate facts.

`HookManager.lua` is not touched, so this is disjoint from the parked RSF-F196 slice 4.

Design-origin: SF-79 build brief v1.0, section 3.A (SDS v1.3) | Disposition: CARRIES

## MP

Server-only by construction: the writer returns UNAVAILABLE `not-server` off the server (`PositionalPH.lua:294-297`) and `applyFertilizer` is the server application path behind `onFertilizerApplied`. No stream, event or save change. A pure client sees nothing new.

## Bench

`tools/test/lua/SF-79-is_true_lime_test.lua`, 55 rows.

- **Entry-point bar: group A.** Drives the real `onFertilizerApplied`, then the real `paintBoomStrip` at the same tick, the order the sprayer hook uses (`HookManager.lua:4905` then `:5183/:5238`). The spray-type table is never hand-set: LIQUIDFERTILIZER, FERTILIZER and LIME are registered first as the base types the engine supplies, through a manager that applies the engine's `addSprayType` rule verbatim (`isLime` derived from the type name, only on first creation; fill type index mapped to spray type), and then the mod's own `HookManager:registerCustomSprayTypes` registers LIQUIDLIME and GYPSUM itself. Rows R1/R2 assert those two came from that registration. The writer is spied to record the request production built and then called for real; the one stub on the path is `_phRefreshScalar`, the report refresh that runs after the request.
- **Group B, the divergence rows, SYNTHETIC.** A pH raiser registered as FERTILIZER gives false on both requests, and a pH raiser with no spray type gives false. No shipped product has this shape; without these rows a "classify by pH sign" implementation passes every real-product row.
- **Group C, writer blindness.** The same request with `isTrueLime` true, false and nil asks the recording map for byte-identical operations and returns the same result, over STRIP DELTA, POINT DELTA, POLYGON NORMALIZE and POLYGON SET.
- **Group D.** The milestone line carries `trueLime=true` for LIME and LIQUIDLIME, `false` for GYPSUM, exactly one line per 1000 L crossing, and a tick that crosses no boundary prints nothing.

Suite in the build worktree: PASS, 5866 assertions, 0 failed across 113 files. Syntax and lint clean.

Mutations (`tools/test/mutate_sf79_true_lime.py`), each proved to land by exact occurrence count and restored by sha256: M1 hardcode true KILLED (7 red), M2 hardcode false KILLED (6), M3 classify by pH sign KILLED (4, group B alone), M4 writer refuses when not true lime KILLED (16), M5 dropped from the strip request only KILLED (5), M6 dropped from the dose record KILLED (2), M7 dropped from the milestone line KILLED (3). 7 killed, 0 survived, 0 unapplied.

## Not proved here

That the live game's LIME and LIQUIDLIME spray types carry `isLime` (the base LIME entry comes from map XML the decoded tree does not hold, and a spray type is typed only on first creation), and that native Needs Lime behaviour is unchanged. Both are the TESTING.md row.

## Register

Queried `DESIGN-CHECK.md` and `MAINTENANCE.md` at tracking `389af36`: no row names `isTrueLime` beyond the "not carried here" notices in DESIGN-CHECK rows 71 and 72. No hold governs it. R number: none (SF-79 is a travelled feature, not an Ordered Repair Lane row).

## Review

Bob's cold review at 1148ea92: PASS, no BLOCKER, MAJOR or MINOR; disposition CARRIES confirmed against SF-79 3.A. He re-ran the bar (55/0), the mutation script (7/0/0) and the full suite in his own worktree.


TheCodingDad-TisonK:
**Bob, cold review of #981 at `1148ea92`: PASS.** No BLOCKER, MAJOR or MINOR.

Head checked with `ls-remote` and against the PR's own `headRefOid`. Base is `development` `d7d5cd99`. The branch carries exactly one commit, and the title matches it:
- `1148ea92` feat(SF-79): isTrueLime, the native-credit classification on the footprint request

**Checked cold**
- `isTrueLime` is declared at the dose block's level (`src/SoilFertilitySystem.lua:6003`). Every reader sits in that block or deeper: the dose record (`:6019`), the milestone line (`:6046`) and the narrow-tool request (`:6115`). No read falls through to a nil global.
- The spray-type lookup gives false in all three failure shapes: no manager, no method, and a nil spray type.
- One level out: `_sprayDose` is created only at `:6016` and read only at `:4425` and `:6021`. Nothing saves, streams or copies it.
- The pH-sign classifiers at `:5795` and `HookManager.lua:2126` are untouched. `HookManager.lua` is untouched, so the PR stays disjoint from the parked F196 slice 4. `src/PositionalPH.lua` gains a comment only; the writer does not read the field.
- Multiplayer: server-only, as before. No stream, event or save change.

**Re-run by me in a detached worktree at `1148ea92`, not taken from the handback**
- `SF-79-is_true_lime_test.lua`: 55 passed, 0 failed.
- `mutate_sf79_true_lime.py`: 7 killed, 0 survived, 0 unapplied, sources byte-identical after. M3 (classify by pH sign) is killed only by the synthetic divergence rows in group B, as designed.
- Full suite: 5855 passed, 0 failed across 113 files. The one file that did not load is `om_213_organic_premium_test.lua`, which needs the sibling MarketDynamics repo; that is the scratch environment, not the change.

**Who populates the world.** Group A is the entry-point bar. It drives the real `onFertilizerApplied`, then the real `paintBoomStrip`, at one tick. The mod's own `registerCustomSprayTypes` fills the spray-type table against a manager applying the engine rule (`SprayTypeManager.lua:62`); rows R0-R2 assert that. The three base types, LIME among them, are fixture-supplied, and the bar's header says so. The TESTING.md row carries the in-game proof.

**Design check.** Disposition CARRIES against SF-79 3.A, confirmed, not raised.

**Rows present at tracking `main` `3d9c8e9`.**
- `TESTING.md` row for #981 names concrete checks: three products at 1000 L each, the `trueLime=true/true/false` log read, and Needs Lime clearing for the two limes and not for gypsum.
- `DESIGN-CHECK.md` row for #981 records origin SF-79 3.A, CARRIES.

sasha-rf (approved):

Cold-checked the diff against the claim of "no behaviour change": swept src/ for isTrueLime/.isLime reads and found only the doc comment PositionalPH.lua's _applyPHFootprint adds, which itself says the writer deliberately never reads the field. HookManager.lua untouched, burn gate and VR curve untouched, both confirmed by the diff. Bob's cold review (1148ea92) is PASS, no findings, CARRIES. TESTING.md/DESIGN-CHECK.md rows on tracking main at 3d9c8e9. Suite 5866/0, mutations 7/0/0. Diagnostic-only field per SF-79 3.A, not a design-intent change (FAST TRACK, implements already-decided brief).

PR #982, opened 2026-09-22
feat(SF-79): the sf79PHSchema marker gets its one reader, the seed freeze on load

TheCodingDad-TisonK:
## What

SF-79 section 3.B: "Before new PH writes, seed only raw-zero supported ground from a frozen pre-migration scalar clamped to carrier bounds, or the existing genesis rule. Metadata uses the existing Soil XML key: root `#sf79PHSchema`, field `#sf79PHSeed`."

The root marker has been written on every soil XML save since 71c9bcf2 (`SoilFertilitySystem.lua:7775`) and read nowhere. And `field._phSeedScalar`, the frozen scalar, was never originated: its only writers were the two loaders restoring `#sf79PHSeed` (`PositionalPH.lua:677`, `SoilFertilitySystem.lua:8364`), and `#sf79PHSeed` is only saved when the seed is already a number. No save has ever carried a seed.

**The marker decides one thing.** `field.pH` is republished from the derived report by `_phRefreshScalar` after every positional write (`PositionalPH.lua:418-425`), so on a MARKED save the loaded `#pH` may already be "the later report", which 3.B forbids as a seed. Only an UNMARKED save proves its `#pH` predates the contract. This PR:

- adds `_phFreezeSeedFromLoad(field, schemaMarked)` beside `_phLoadFieldXML`: on an unmarked load with no seed, freeze `field.pH` clamped to `PH_MIN..PH_MAX` (5.0..7.5, the carrier bounds; the load itself clamps to 8.5). A marked save with no seed stays nil on purpose: the brief's named alternative, the genesis rule, is the seeding item's. A present seed is never re-frozen.
- `loadFromXMLFile` reads the marker once beside the f66 read and calls the helper after `_phLoadFieldXML` for each field.
- `getSoilStateTable` now mirrors `sf79PHSchema = 1`, which its own comment already claimed it did; `applySoilStateTable` reads it and makes the same per-field call after the seed restore. Both loaders share the helper so they cannot drift.
- The marker gates ONLY the freeze. Seeding and missing-carrier recovery never consult it ("a schema marker never proves the carrier exists"). The save path is unchanged.

Visible effect: after the first save on a formerly unmarked save, `field(n)#sf79PHSeed` appears in the soil XML for the first time.

### Second commit, 6a3a8d4e: the ledger load path (Bob's BLOCKER on the first cut)

With FS25_StateLedger delivering a block, `loadSoilData` runs ONLY `applySoilStateTable` (`SoilFertilityManager.lua:1301-1309`), and every ledger snapshot written before this PR lacks `sf79PHSchema`, dev and tester snapshots included. The first cut read "absent" as "unmarked" there, which would have frozen a dev save's later-report scalar and persisted it at the next save. Live on Tyson's own savegame20, which loads through StateLedger.

Fix: `SoilFertilityManager._readSoilXMLSchemaMarker` reads only the root `#sf79PHSchema` of the `soilData.xml` safety copy (written every save by the same builds); the ledger branch passes it through `SoilStateLedgerBridge.applyState` into `applySoilStateTable`, where `marked = (data.sf79PHSchema == 1) or xmlRootMarked`. A missing safety copy counts as UNMARKED, named in the comment: a released player's pre-marker save with a ledger block must still freeze.

Design-origin: SF-79 build brief v1.0, section 3.B (SDS v1.3) | Disposition: CARRIES (declared; Bob's confirmation owed on 6a3a8d4e)

## MP

Savegame load and the StateLedger restore are server-side; `_phSeedScalar` is not streamed (no stream site reads it). No event change. A pure client sees nothing new.

## Handoff to the seeding item (item 3)

This PR PRODUCES `_phSeedScalar`; the seeding wiring CONSUMES it at `PositionalPH.lua:600-601`. That fallback, `seed = field._phSeedScalar or field.pH`, is the later report on every marked-no-seed save (every dev and tester save since 71c9bcf2), and the seeding PR replaces it with the genesis rule. The seeding branch is cut after this merges.

## Bench

`tools/test/lua/SF-79-ph_schema_seed_freeze_test.lua`, 46 rows. **Every row is an entry-point row**: each drives the real `loadFromXMLFile`, the real `applySoilStateTable`, or (group I) the real `loadSoilData` through its ledger branch with the real `SoilStateLedgerBridge.applyState`, with an XML handle, a ledger table or a fixture disk as the input, exactly as the mission load and the StateLedger restore call them. No row sets `_phSeedScalar` by hand.

- A: unmarked, `#pH` 8.0 loads as 8.0 and freezes 7.5 (the carrier bound). B: unmarked 6.2 freezes 6.2; `#pH` 4.0 loads clamped to 5.0 and freezes 5.0. C: marked, no seed, freezes nothing. D: marked with `#sf79PHSeed` 6.1 and `#pH` 6.9 keeps 6.1. E: unmarked with a seed keeps the seed.
- F: the round trip. Unmarked load freezes 6.2; the scalar moves to 6.8; the real `saveToXMLFile` writes root `#sf79PHSchema=1`, field `#sf79PHSeed=6.2` and `#pH=6.8`; the reload (now marked) keeps 6.2 and 6.8.
- G: the ledger mirror. `getSoilStateTable` carries `sf79PHSchema=1`; unmarked table 8.0 freezes 7.5; marked-no-seed freezes nothing; marked-with-seed keeps it; a real snapshot of a seedless field restores with no seed; a frozen seed survives the mirror while the scalar differs.
- H: the marker gates only the freeze, plus the helper's own contract.
- **I, the transition (Bob's MAJOR):** through the real `loadSoilData` ledger branch, a block WITHOUT the key with the safety copy MARKED gives no freeze; UNMARKED gives the freeze; NO safety copy gives the freeze at the carrier bound; a post-PR block WITH the key gives none; a seed in the block is carried; and `_readSoilXMLSchemaMarker` itself reads marked, unmarked and missing.

Suite in the build worktree: PASS, 5912 assertions, 0 failed across 114 files. Syntax and lint clean.

Mutations (`tools/test/mutate_sf79_schema_key.py`), each proved to land by exact occurrence count and restored by sha256: M1 ignore the marker KILLED (5 red), M2 drop the clamp KILLED (3), M3 drop `sf79PHSchema` from the ledger snapshot KILLED (2), M4 re-freeze when a seed is present KILLED (2), M5 XML loader drops the freeze call KILLED (7), M6 ledger loader drops the freeze call KILLED (3), M7 ledger loader ignores the XML root marker KILLED (1), M8 loadSoilData stops passing the marker KILLED (1), M9 a missing safety copy counts as marked KILLED (2). 9 killed, 0 survived, 0 unapplied.

## Not proved here

That the game's XML functions write and read the attributes as the mock does, and the on-disk shape of a real main-branch save. Both are the TESTING.md row, which now runs on both load paths.

## Register

Queried `DESIGN-CHECK.md` and `MAINTENANCE.md` at tracking `3d9c8e9`: no row names the schema key beyond the "not carried here" notices on DESIGN-CHECK rows 71 to 73; MAINTENANCE row 70 is the seeding item's. No hold governs this key. R number: none.

## Review

Bob's cold review at ec80899a: FAIL (issuecomment-5785057648), one BLOCKER and one MAJOR, both repaired at 6a3a8d4e above. Re-review owed on 6a3a8d4e.


TheCodingDad-TisonK:
**Bob, cold review of #982 at `ec80899a`: FAIL.** One BLOCKER. The gap is in my own intake shape, which Fred built exactly as specified.

The head matches `ls-remote`. The base is `development` `6df1d63d` (the #981 merge). The branch carries exactly one commit, and the title matches it:
- `ec80899a` feat(SF-79): the sf79PHSchema marker gets its one reader, the seed freeze on load

**Findings**

`src/SoilFertilitySystem.lua:8350`: **BLOCKER.** On the StateLedger path a missing `sf79PHSchema` is read as "unmarked". That is true for a pre-SF-79 snapshot, but it is also true for every ledger snapshot any build has written until now, including every dev and tester build since `71c9bcf2` (2026-09-10).
- Why: `getSoilStateTable` never carried the key before this PR.
- Which path runs: when StateLedger delivers a soil block, `loadSoilData` runs ONLY `applySoilStateTable` and returns (`src/SoilFertilityManager.lua:1301-1309`). The XML root marker, which those builds did write (`:7775`), is never consulted.
- What goes wrong: on those saves `_phFreezeSeedFromLoad` freezes `field.pH`. `_phRefreshScalar` republishes that value from the report after every positional write, so on any field that has been limed it is "the later report", which SF-79 3.B forbids as a seed.
- Why it cannot be undone: the next save writes the value as `#sf79PHSeed` and `sf79PHSeed` with the marker set, so it is frozen for good. The later seeding PR would then fill unwritten ground from it.
- The exposure is live, not theoretical. The office's own test save (savegame20, log.txt 2026-09-22 22:17:59) loads soil through StateLedger ("Registered with StateLedger as 'SoilFertilizer_Soil'", 24 module blocks parsed). Released saves are unaffected (the marker is absent and the scalar is genuinely pre-migration there), but every tester save with StateLedger is not.

The fix: on the ledger path an absent key means "a snapshot written before the ledger carried the marker", not "unmarked". Resolve it from the safety copy, which `saveToXMLFile` writes on every save beside the ledger:
- In the ledger branch of `loadSoilData`, read the root `#sf79PHSchema` of `savegameDirectory/soilData.xml` when the file exists, and hand it to `applySoilStateTable`.
- There: `marked = (data.sf79PHSchema == 1) or xmlRootMarked`.
- If the XML file is missing, the snapshot counts as unmarked. That is the one remaining edge; name it in the comment.

Do NOT fix it by never freezing on the ledger path. Players on released builds who have StateLedger must still freeze their genuine pre-migration scalar.

`tools/test/lua/SF-79-ph_schema_seed_freeze_test.lua:146-147`: **MAJOR.** G3/G4 assert the freeze on an unmarked ledger table, and there is no row with the transition shape: a ledger snapshot WITHOUT the key while the XML root IS marked. Add it (expected: no freeze) beside the pre-SF-79 case (ledger without the key and XML without the marker: freeze), and drive both through `loadSoilData`'s ledger branch so the XML read is exercised from production's own entry. Add a mutation that ignores the XML marker and a row that kills it.

**Everything else checked cold, and it holds**
- The helper refuses correctly in every case: a non-table record, a marked save, an existing seed, a non-number pH, missing limits.
- The clamp is to `NUTRIENT_LIMITS` 5.0/7.5, which equals the carrier bounds.
- The XML read sits beside the f66 read, and the per-field call comes after `_phLoadFieldXML`, so a carried seed wins.
- On the ledger side the call comes after the seed restore and before `fieldData` is assigned.
- `getSoilStateTable` now mirrors the root marker. The save path is untouched.
- The marker gates nothing else (group H).
- Multiplayer: server-side load only, nothing streamed.
- The entry-point bars drive the real `loadFromXMLFile` and `applySoilStateTable`. The ledger-branch selection in `loadSoilData` itself is not driven, which is how this case slipped past both of us.

**Not re-run on this head:** the mutation battery and the full suite. I will run both on the fix commit.

**Design check.** The disposition cannot be confirmed as CARRIES while the ledger path freezes "the later report" against 3.B. The DESIGN-CHECK row stays unconfirmed until the fix lands; I will confirm or raise it on re-review.

Re-review on the fix commit.

TheCodingDad-TisonK:
**Bob, cold re-review of #982 at `6a3a8d4e`: PASS.** The BLOCKER and the MAJOR from my first-cut verdict are both closed. No new findings.

The head matches `ls-remote`. The base is `development` `6df1d63d`. The branch carries two commits, and the PR body lists both:
- `ec80899a` feat(SF-79): the sf79PHSchema marker gets its one reader, the seed freeze on load
- `6a3a8d4e` fix(SF-79): the ledger load path reads the schema marker from the soilData.xml safety copy

**The BLOCKER, closed.** On the StateLedger path the marker now also comes from the root of the `soilData.xml` safety copy:
- `SoilFertilityManager:_readSoilXMLSchemaMarker` reads only `soilData#sf79PHSchema` and deletes its handle.
- The ledger branch of `loadSoilData` passes the result through `SoilStateLedgerBridge.applyState` into `applySoilStateTable`, where `marked = (data.sf79PHSchema == 1) or xmlRootMarked`.

Checked one level out:
- These are the only callers of `applyState` and `applySoilStateTable`, so no path reaches the ledger loader without the marker.
- `saveSoilData` writes `soilData.xml` on every save and has no StateLedger branch, so the safety copy exists beside every ledger snapshot any build has written.
- A missing copy counts as unmarked. That is the right way round: a released player's pre-marker save still freezes its genuine scalar. The edge is named in both comments.
- The XML path is unchanged from the first cut.

**The MAJOR, closed.** Group I drives the real `loadSoilData` through its ledger branch with the real bridge. I1 proves the ledger, not the XML, did the loading (6.9 from the block against 5.5 in the copy). I2 is the transition case: key absent, copy marked, no freeze. I3 and I4 are the pre-marker cases, which freeze (I4 at the carrier bound). I5 and I6 cover a post-#982 snapshot and a carried seed. I7 to I9 cover the reader itself.

**Re-run by me in a detached worktree at `6a3a8d4e`**
- `SF-79-ph_schema_seed_freeze_test.lua`: 46 passed, 0 failed.
- `mutate_sf79_schema_key.py`: 9 killed, 0 survived, 0 unapplied, sources byte-identical after. The three new mutants (ignore the XML marker, stop passing it, a missing copy counts as marked) are each killed.
- Full suite: 5901 passed, 0 failed across 114 files. The one file that did not load is `om_213_organic_premium_test.lua`, which needs the sibling MarketDynamics repo; that is the scratch environment, not the change.

**Who populates the world.** Every row drives a real loader. Groups A to H use the real `loadFromXMLFile` and `applySoilStateTable` on XML or ledger input. Group I, the entry-point bar for the path Tyson's own save takes, uses the real `loadSoilData` ledger branch. Nothing is hand-set on the field record.

**Design check.** Disposition CARRIES against SF-79 3.B, confirmed on this commit, not raised.
- The marker's one reader is the seed freeze.
- A marked save with no seed is left for the genesis rule, which is the seeding item's job.
- The marker gates nothing else.

**Rows present at tracking `main` `67a0af2`.** The `TESTING.md` row for #982 runs both load paths, with StateLedger enabled for the dev-save case. The `DESIGN-CHECK.md` row for #982 records the first-cut FAIL and the fix; Fred adds the confirmation wording now that this verdict is on the PR.

sasha-rf (approved):

Reviewed at fix head 6a3a8d4e (not the first cut). This closes Bob's original BLOCKER: the StateLedger load path now reads the root #sf79PHSchema marker from the soilData.xml safety copy via SoilFertilityManager:_readSoilXMLSchemaMarker and threads it through SoilStateLedgerBridge.applyState -> applySoilStateTable(data, xmlRootMarked), defaulting to unmarked/false when the safety copy is missing, which correctly forces the freeze for pre-marker and hand-edited saves. getSoilStateTable now always stamps sf79PHSchema=1 so future ledger restores make the same decision an XML load would. Both loaders share the one _phFreezeSeedFromLoad helper, gated only on the marker, clamped to PH_MIN/PH_MAX. Bob's re-review: PASS, bar 46/0, 9/9 mutations killed, CARRIES.

UTF-16 scan: 5 file(s) read at the PR head, all clean.

PR #983, opened 2026-09-22
fix: say so when the fill type width floor is already satisfied at load

TheCodingDad-TisonK:
## What

Tyson's ruling of 2026-09-22, maintenance lane: when `SoilFillTypeWidth.applyFloor()` finds the fill type index width already at the floor or above, print one INFO line, so a log alone shows the floor is in force.

Why it was needed: `applyFloor` returned `false, current` silently when another mod had raised the width first (`src/utils/SoilFillTypeWidth.lua:88-90`), and the only print was the caller's, only on a raise. A tester's log with no "raised" line read as a possible miss for an evening until FS25_ProductionStorageControl was found printing `FillType(1024)` before SoilFertilizer sourced. On Tyson's own machine Realistic Livestock RM sets 10 first and silences the line the same way.

The change:

- `SoilFillTypeWidth.loadLine(raised, width)`, pure, returns the one load-time line: `[SoilFertilizer] Fill type index width raised to N (M fill types). FillType Extender is not required.` when this mod did the raising (unchanged text), `[SoilFertilizer] Fill type index width already N (M fill types) at SoilFertilizer load, floor 10 satisfied. FillType Extender is not required.` when it found the floor met, and nil when the width was unreadable. "At SoilFertilizer load" because it is the width in force when this mod sources; a mod sourcing later can still change it.
- `main.lua` calls `applyFloor()`, then `loadLine`, then prints the line if there is one. `applyFloor` is unchanged, so the existing bar's return-value rows stand and the bench drives the real branch rather than a copy. Plain `print`, not SoilLogger: Logger.lua is sourced after this call.
- Once per load, never per frame: `main.lua` is the mod's only extraSourceFile (`modDesc.xml:68-70`), sourced once by loadMod (`mods.lua:933-937`), and the call is at top-level file scope. A hot-reload re-source prints it again, which is a real second load.

No behaviour change. No gating. No network or save change.

Design-origin: none | Disposition: none (maintenance lane, MAINTENANCE.md row added and closed with this PR)

## MP and dedicated

No state. Every peer (dedicated server, host, each client) sources `main.lua` and runs `applyFloor` on its own, so each machine prints its own line into its own log.txt, dedicated included. That helps: `FillTypeManager.SEND_NUM_BITS` sets the stream width, and a server-client mismatch desyncs; with this line both logs show the width.

## Bench

[...]

PR #984, opened 2026-09-22
fix(SF-79): both soil re-rolls route the new pH through the positional writer

TheCodingDad-TisonK:
## What

SF-79 section 3.C: "Route physical lime/gypsum, normalization, reroll/admin/recovery through the writer." Both soil re-rolls did not.

`rerollAllFields` (console `SoilRerollFields`, `SoilFertilitySystem.lua:3524`) and `rerollUnownedFields` (console `SoilRerollUnownedFields`, `:3555`) set `field.pH = soil.pH` and call `vmSeedField(fieldId, true)`, which walks `VM_NUTRIENT_KEYS`. pH left that list in f9e3db4a when the map became the chemical authority (main still has it at `:3611`). So the pH map kept its old pixels: `getFieldInfo`'s FIELD_REPORT showed the OLD mean at once, and `_phRefreshScalar` (`PositionalPH.lua:418-425`) wrote that old mean back over the scalar after the field's next pH write. The re-roll was invisible and then undone, the plow's shape before #979.

The change, one function shared by both re-rolls: `_phRerollField(fieldId, field)` asks the writer for a FIELD SET at the re-rolled scalar with source `'reroll'`, the admin path's shape (`:3766-3772`), then refreshes the scalar from the derived report so it equals the quantised map value. SET's band is `0..RAW_MAX` (`PositionalPH.lua:368-372`), so raw-zero ground in the union is written too, which is the right outcome for "re-roll the starting soil" and needs no separate seed call. `_phSeedScalar` and `_phPending` are deliberately left alone (re-freezing the seed would change what "frozen" means; pending remainders are sub-step by construction). With PositionalPH absent nothing is painted, which is what this branch did before. Admin, N/P/K/OM and the console commands' server gate are untouched.

Design-origin: SF-79 build brief v1.0, section 3.C (SDS v1.3) | Disposition: CARRIES (declared; Bob's cold verdict owed on 7462537b)

## MP

Both console commands refuse off-server (`SoilSettingsGUI.lua:1835-1838`, `:1856-1859`) and the writer refuses off-server (`PositionalPH.lua:294-297`). Clients get the pH pixels by whatever path carries every other writer write; that path was not measured for re-rolls, so the TESTING row carries an MP step.

## Bench

`tools/test/lua/SF-79-reroll_ph_routing_test.lua`, 32 rows. **Entry-point bars: every group** drives `rerollAllFields` or `rerollUnownedFields` themselves, the functions the console commands call, over the real PositionalPH writer, report and refresh. The parcel polygons come from the production resolver `_getFarmlandPolygons` reading a `g_fieldManager.fields` fixture through `getWorldTranslation`; the re-rolled values come from the real `_computeInitialSoil`; only the native map is a model (a 2 m pixel grid implementing `setPolygonWhere`, `applyRawDeltaToPolygonBand` and `readAverageRawInBand`, the three calls the writer and report make, with recording stubs for the N/P/K/OM seeding). Ownership is a `g_farmlandManager` fixture and NPC status a `NpcSoilBridge` fixture, both stated in the bar.

- A: `rerollAllFields`. The writer is asked once per field, as SET over the FIELD domain with source `reroll` at the genesis value; every union pixel equals `phRaw(new pH)`, the raw-zero strip included; the report is CURRENT and equals the new pH quantised; the scalar is refreshed to exactly the report value; the N/P/K/OM seed still runs and `vmSeedField` never seeds pH itself (the gap is real).
- B: the consequence, two-sided. Routed: re-roll, then lime through the real adapter, and the scalar moves up from the NEW base. Control with the routing removed: the scalar alone says the new base, the map still holds the old pixels, and the next pH write republishes the OLD mean plus the lime, the re-roll undone.
- C: `rerollUnownedFields`. Exactly one field goes to the writer (the unowned, non-NPC one); the owned field's and the NPC field's pixels are byte-identical before and after; the unowned field's pixels all equal the new value and its scalar is the quantised map value.
- D: the writer absent: both re-rolls survive and pH pixels are untouched.

Suite in the build worktree: PASS, 5898 assertions, 0 failed across 114 files. Syntax and lint clean.

Mutations (`tools/test/mutate_sf79_reroll.py`), each proved to land by exact occurrence count and restored by sha256: M1 call dropped from `rerollAllFields` only KILLED (10 red), M2 call dropped from `rerollUnownedFields` only KILLED (4), M3 OP_DELTA in place of OP_SET KILLED (8), M4 the refresh dropped KILLED (2). 4 killed, 0 survived, 0 unapplied.

## Not proved here

Native density-map execution, how a joined client learns a re-roll, and the console commands' server gate. All in the TESTING row.

## Register

Queried `DESIGN-CHECK.md` and `MAINTENANCE.md` at tracking `67a0af2`: MAINTENANCE row "Both soil re-rolls set the pH scalar and never reach the pH map" (Bob, eefd58e) is this item and closes with this PR; no DESIGN-CHECK row named reroll before it. R number: none (SF-79 is a travelled feature). Independent of the seeding item: it touches only `:3518-3594` plus the new helper above it, reads no seed.


TheCodingDad-TisonK:
**Bob, cold review of #984 at `7462537b`: PASS with MAJORs.** One MAJOR, and it is in the register, not the code. The code has no findings.

The head matches `ls-remote`. The base is `development` `6df1d63d`. The branch carries exactly one commit, and the title matches it:
- `7462537b` fix(SF-79): both soil re-rolls route the new pH through the positional writer

**Checked cold**
- One helper, `_phRerollField` (`src/SoilFertilitySystem.lua:3527-3537`), called from both re-rolls right after `vmSeedField` (`:3571`, `:3620`). It sends an OP_SET, SCOPE_FIELD request with source 'reroll' to the writer, then refreshes the field's scalar. That is the admin path's shape, as my intake specified.
- `_phSeedScalar` and `_phPending` are untouched, as specified. N/P/K/OM are untouched, because `vmSeedField` still runs (A11).
- On a client the writer returns 'not-server', and both console commands are gated to the server. There is no network change.

**Fred's two questions, answered cold**
1. The bar's SET model matches `SoilValueMaps:setPolygonWhere` as the writer uses it (`src/maps/SoilValueMaps.lua:1394-1421`): a pixel whose current raw lies in the inclusive band [lo, hi] takes the value, and at lo = 0 that includes raw zero. Production gets that band from a BETWEEN compare (`:1413`); whether the engine's BETWEEN is inclusive is a native confirm, the same assumption every SF-79 bar and the seeders rely on. The model leaves out the value clamp and the `#verts < 3` refusal, which cost this item nothing.
2. Shadowing `_phRerollField` on the instance is exactly the code before this item: the diff is only the helper and its two calls. B3-B5 therefore show the real regression: scalar new, map old, and the next write undoes the re-roll.

**Findings**

`MAINTENANCE.md` (tracking `d5c8ae3`), the row "Both soil re-rolls set the pH scalar and never reach the pH map...": **MAJOR.** Its Closed cell is already filled with #984. The file's own rule says a row is closed by MERGED work, filled by the seat that ships it in the same turn as the merge. #984 is open, and a row that says closed while its PR is unmerged reads as done if the PR stalls or is refused. Revert the cell to empty and fill it at merge. #983's MAINTENANCE row was closed the same way at PR open; I missed that in my #983 review, so it needs the same correction.

**Re-run by me in a detached worktree at `7462537b`**
- `SF-79-reroll_ph_routing_test.lua`: 32 passed, 0 failed.
- `mutate_sf79_reroll.py`: 4 killed, 0 survived, 0 unapplied, source byte-identical after (each re-roll's call dropped separately, OP_DELTA for OP_SET, refresh dropped).
- Full suite: 5887 passed, 0 failed across 114 files. The one file that did not load is `om_213_organic_premium_test.lua`, which needs the sibling MarketDynamics repo; that is the scratch environment, not the change.

**Who populates the world.** The bar drives both real re-rolls, the functions the two console commands call. They run over the real PositionalPH writer, report and refresh. The polygons come from production's `_getFarmlandPolygons` over a `g_fieldManager` fixture, not a hand-built domain. Ownership and NPC status come from `g_farmlandManager` and `NpcSoilBridge` fixtures, named in the bar. The console gate itself is not driven, and there is nothing in it to pin.

**Design check.** Disposition CARRIES against SF-79 3.C, confirmed, not raised.

**Rows.** `TESTING.md` row present, with the MP step. `DESIGN-CHECK.md` row present, declared, verdict owed; the confirmation wording can go in now.

sasha-rf (approved):

Reviewed at 7462537b. Both rerollAllFields and rerollUnownedFields now route the re-rolled pH through the shared _phRerollField -> _applyPHFootprint (OP_SET, SCOPE_FIELD, source 'reroll') -> _phRefreshScalar, matching the admin path's shape at SoilFertilitySystem.lua:3766-3772. N/P/K/OM untouched, _phSeedScalar/_phPending deliberately left alone. Bob's cold review: PASS, no code findings, CARRIES confirmed against SF-79 3.C. His one MAJOR was register hygiene (MAINTENANCE row marked Closed before merge), already corrected by Fred (9f8fc77). TESTING.md and DESIGN-CHECK.md rows present. Suite 5887/0, mutations 4/0/0.

PR #985, opened 2026-09-22
feat(RSF-F196): X1, the price map rebuilt on every registration attempt and completion honest about price

TheCodingDad-TisonK:
## What

RSF-F196 v0.9 clause X1 (registration completeness, brief `:215-221`), with V12's "separately rebuild `customFillTypePrices` idempotently ... after late descriptors resolve" (`:112`) and V12a's retry half (`:114`).

The identity catalogue was already rebuilt on every registration attempt (#977), and completion was `_sprayTypesComplete = (skipped == 0)` (`HookManager.lua:1061`). The missing half was PRICE: `customFillTypePrices` had exactly one writer, `installPurchaseRefillHook` (`:7508-7514` local, `:7667` assign), which runs once from `installAll` (`:557`) and never on a retry. On a dedicated server with late descriptors (#431), a product that resolved after install never got a price, so helper buying refused it for the whole session (V12b) while the completion flag still went true and the retry loop stopped.

The change (`src/hooks/HookManager.lua` only):

- `HookManager.customPriceFor(name)`: the ONE price source. Constants overrides (`PURCHASABLE_SINGLE_NUTRIENT.pricePerLiter`) win, else the fillTypes.xml economy fallback table, lifted verbatim out of the install so there is still one table, never two.
- `HookManager:rebuildCustomPriceMap()`: builds a COMPLETE replacement map for every resolved, non-refused population index whose name has a price, assigns it to `self.customFillTypePrices` in one step (no partial map is ever observable by the two call-time readers, `billedExternalFill` and the backup refill), and returns whether every eligible index carries a finite, nonnegative price. Called at the end of `registerCustomSprayTypes` beside the catalogue rebuild.
- `self._sprayTypesComplete = (skipped == 0) and priceComplete`. Eligibility is "the name has a price entry": blends (`SoilBlends.ORDER`) have none and never count as missing; if they did, every load would time out.
- `installPurchaseRefillHook` builds its install-time copy through the same rebuild. The purchase WRAPPER is untouched: it still reads the captured install-time copy (`:7627`), and making it read the live map at call time plus installing when the first map is empty is V12a's wrapper half, which rides with V12c in slice 4. `getExternalFill` and the backup refill already read `hookMgr.customFillTypePrices`, so they gain the late prices immediately.
[...]

PR #987, opened 2026-09-22
feat(RSF-F196): V12c the owned registered-function layer, V13 the native-drain token

TheCodingDad-TisonK:
## What

RSF-F196 v0.9 clauses V12c (the owned registered-function layer) and V13 (the native-drain token), one PR. They are one PR because V12c's propagation would make today's untokened buy-mode intercept (`HookManager.lua:7611-7660` at 6df1d63d) reach live vehicles for the first time, and V13 then replaces that intercept; split, the first PR ships a behaviour change only for the second to undo.

**Finding 1, in one line:** a registered specialization function is copied into every vehicle type at game boot (`TypeManager:finalizeTypes`, `FillUnit.lua:197`, `Sprayer.lua:81-82`) and onto every instance at `Vehicle.lua:486` (`SpecializationUtil.copyTypeFunctionsInto`), so a class-level patch made at mission load reaches no loaded vehicle; the purchase-refill FillUnit hook could never fire in production, and BUY was carried by the opt-in plus `getExternalFill` and by the backup refill.

The change (`src/hooks/HookManager.lua` and `tools/test` only):

- **`installOwnedRegisteredFunctionLayer(class, funcName, makeWrapper, specName, label)`.** One installer, three surfaces, one owner record per reference: (i) the class slot from its current value; (ii) every vehicle type carrying `specName` whose `functions` table holds the slot, wrapping the function actually present (a composed `Utils.overwrittenFunction` chain such as SowingMachine's is wrapped, never bypassed); (iii) every live instance holding a raw copy (`rawget`), reusing the type wrapper when the raw pointer equals a type predecessor. Records `{owner="F196", slot, predecessor, wrapper}`; refuses to wrap a slot already holding one of its own wrappers (a retry cannot stack layers); one cleanup restores a slot only if it still holds the recorded wrapper. `ensureOwnedLayers(vehicle)` gives a late vehicle exactly one wrapper, called from the `FillUnit.onPostLoad` prepend (raised at `Vehicle.lua:905`, after the copy at `:486`). Written as a reusable method on purpose: RSF-741 item 9 tags the same shape once per surface and reuses it.
- **The three installs move onto it.** `getExternalFill`: the billing body takes the predecessor as a parameter and one factory serves all three surfaces; the old `propagateExternalFillHookToLiveVehicles`, which assigned the CLASS wrapper over any type-level predecessor and whose cleanup restored the class only, is now a thin idempotent re-sweep kept for the three older bars that call it. `getIsSprayerExternallyFilled`: the inline class assignment, type scan and vehicle scan (unowned: no records, no guard, no late-vehicle ensure, class-only cleanup) are replaced by one layer call with the same `makeReplacement` factory. `addFillUnitFillLevel`: the class-only assignment becomes the FillUnit factory.
- **V12a's wrapper half.** The FillUnit wrapper reads `hookMgrRef.customFillTypePrices` and identity (`resolveCustomProductIntent`, `isRefusedProduct`) at CALL time, never a captured copy, and the layer installs even when the first price map is empty (the early return became a warning), so a product that resolves after install is billable.
- **V13, `installDrainTokenHook`.** A prepend on `Sprayer.onEndWorkAreaProcessing` stores, on `wap.sprayVehicle` (the receiver of native's drain, not always self, `Sprayer.lua:867-880`), `{fuIdx = wap.sprayVehicleFillUnitIndex, delta = -wap.usage, fillType = wap.sprayFillType}`, only under `self.isServer and wap.isActive` with a finite positive usage: exactly the call native makes at `Sprayer.lua:950`. The FillUnit wrapper consumes a token only on the exact `(fillUnitIndex, delta, fillType)` match; then, in BUY mode with a resolved, unrefused, priced product, it charges the farm, stamps `_soilBuyHandledAt`, counts the charge site and returns without calling the predecessor (native's drain becomes a purchase, the tank untouched). Every untokened delta delegates: `-math.huge`, `emptyAllFillUnits`, the settings drain, the residual snap, stream sync, SoilDrainVehicle. An append clears any unused token. Installed last in `installAll` so the prepend is the outermost wrapper; Bob checked cold that the order is not load-bearing (the constant-remap wrapper swaps global tables, never `wap`).
- Six-argument pass-through (`unloadInfo`) kept in every wrapper.

Out of scope: R7 (slice 5), X1 (#985; this PR shares `installPurchaseRefillHook` with it in different hunks and rebases if git asks), RSF-741, U5a, #764 speed, PositionalPH.

Design-origin: RSF-F196 v0.9, September 19 paired wave (`Office Tyson/DESIGN-WAVE-F196-DC14-2026-09-19/DELIVERY.md`; brief `Office Tyson/RSF-F196-dry-rate-unit-contract-IMPLEMENTATION-BRIEF.md`, clauses V12c and V13, with V12a's wrapper half) | Disposition: CARRIES (declared; Bob's cold verdict owed)

## MP

The server owns tank and money. A token exists only under `self.isServer`, so on a pure client every wrapper delegates and never bills; clients learn the level through FillUnit's own stream. The token is per call, never saved or streamed. No new event.

## Bench

`tools/test/lua/RSF-F196-owned_layer_test.lua`, 79 rows. The world is built the way `finalizeTypes` and `Vehicle:load` build it: class tables with the engine functions, one type whose `addFillUnitFillLevel` is `Utils.overwrittenFunction(FillUnitOrig, SowingMachineFn)`, one bare, one without the spec, instances copied from the type tables BEFORE the installers run; then the mod's own install sequence for these hooks in `installAll`'s order with nothing populated by hand (identity from `registerCustomSprayTypes`, prices from `installPurchaseRefillHook`, layers from the installers). `Utils.prepended/appended/overwrittenFunction` are the engine's verbatim; native's `onEndWorkAreaProcessing` is modelled as `Sprayer.lua:938-957`.

- **Entry-point bar: group C.** Native's own drain from `onEndWorkAreaProcessing` on a sprayer whose spray vehicle is a DIFFERENT vehicle reaches that vehicle's instance slot: BUY charges once on the spray vehicle's farm for usage x price, the tank is untouched, the predecessor chain is not called, the charge site is the layer (A3), and the stamp sits inside the backup refill's own 200 ms window; not BUY, the predecessor drains with `unloadInfo` passed through; on the sowing type the order is SowingMachine then FillUnit with F196 outside the chain.
- A: the world is what the copies make, and the old class-only shape reaches nothing. B: class, both fillUnit types and both instances wrapped, instances reusing the type wrappers, 5 records, the plow type untouched, sprayer layers on the sprayer type only. D: every untokened delta delegates even on a BUY vehicle (`-math.huge`, a snap-sized delta, SoilDrainVehicle's call shape which drains and charges nothing (A4), a fill); a hand-set live token (modelling the prepend's output) with a non-matching delta or fill type delegates and survives; the exact match bills and consumes. E: tokens only under isServer, isActive and a spray vehicle, observed where native runs through a spy inside the hook (native's own gates would otherwise hide a wrongly minted token); a swallowed drain leaves no token. F: one layer per slot across two registration retries and a second installer call while the identity table swaps beneath; a late vehicle through the real `onPostLoad` prepend gets nothing when its copy is already the type wrapper and exactly one wrapper when it is not, idempotently. G: cleanup restores class, types and instances and leaves a slot a later owner replaced. R: the constant-remap swap around native still matches.

Suite in the build worktree: PASS, 5945 assertions, 0 failed across 114 files. Syntax and lint clean. One existing assertion updated: `RSF-F196-density_refusal_test.lua` R3a G5 now asserts the instance copy is an owned-layer wrapper (it was the class closure before; there is one wrapper per predecessor now).

Mutations (`tools/test/mutate_f196_owned_layer.py`), each proved to land by exact occurrence count and restored by sha256: M1 the isActive guard removed KILLED (1 red, group E's spy), M2 the exact-delta match dropped KILLED (4), M3 the class wrapped only KILLED (12), M4 the type-predecessor reuse skipped KILLED (1), M5 cleanup restores unconditionally KILLED (1), M6 the append clear dropped KILLED (1), M7 wrapped around the class function instead of the slot's own KILLED (1). 7 killed, 0 survived, 0 unapplied.

## Not proved here

Engine dispatch of the event, the real backup refill (its 200 ms window is pinned by the stamp the layer writes), the real `SoilDrainVehicle` command (driven as the call shape it makes), and the six engine overwriters beyond SowingMachine. The TESTING row.

## Register

Queried at tracking `cbb0fc0`: DESIGN-CHECK row 66 lists V12c and V13 as NOT YET CARRIED and counts its population as "37" where the brief's bolded ids count 40; this PR's row walks all 40, marks V12c, V13 and V12a's wrapper half carried here, and corrects the count. Unified R4. Bob's slice 4 verdict (d7d5cd99, resent 2026-09-23 with the addendum; HookManager.lua byte-identical between d7d5cd99 and 6df1d63d).


TheCodingDad-TisonK:
**Bob, cold review of #987 at `c3bcbc58`: PASS.** One MINOR, and a rebase note for after #985.

The head matches `ls-remote` and the PR's `headRefOid`. The base is `development` `6df1d63d`. The branch carries exactly one commit, and the title matches it:
- `c3bcbc58` feat(RSF-F196): V12c the owned registered-function layer, V13 the native-drain token

**The owned layer (V12c), checked cold** (`installOwnedRegisteredFunctionLayer`, `src/hooks/HookManager.lua`)
- It wraps the class slot from its current value.
- It wraps every type whose `specializationsByName[specName]` is present. The engine always sets that table (`specialization/TypeManager.lua:163`), and the key is the spec name ("fillUnit", "sprayer"). It wraps the function ACTUALLY in the type's `functions` slot, so a composed chain such as the SowingMachine overwrite sits inside F196.
- It wraps every live instance with a raw copy, reusing the type wrapper when the raw pointer is that type's predecessor. `vehicleSystem.vehicles` is a plain list (`vehicles/VehicleSystem.lua:8`).
- It refuses to wrap one of its own wrappers, keeps a record per reference, and registers one cleanup that restores a slot only if the slot still holds the wrapper. A second call returns the existing record.
- Late vehicles are handled by `ensureOwnedLayers` from the FillUnit `onPostLoad` prepend, which is raised after the copy at `Vehicle.lua:486`.
- All three surfaces move onto the layer. The opt-in's unowned inline scan and its class-only cleanup are gone, and so is the old propagate pass that assigned the class wrapper over type predecessors.

**The drain token (V13), checked cold** (`installDrainTokenHook`, `makeFillUnitWrapper`)
- Native drains with `local usage = spec.workAreaParameters.usage`, then `-usage`, unmodified (`vehicles/specializations/Sprayer.lua:942`, `:950`). So the token `{fuIdx = wap.sprayVehicleFillUnitIndex, delta = -wap.usage, fillType = wap.sprayFillType}` equals native's arguments exactly, and exact float equality is sound.
- The token is set on `wap.sprayVehicle`, the receiver, only under `isServer`, `isActive` and a finite positive usage. It is consumed only on the exact match, whether billed or not. The append clears a token native did not consume.
- The order is not load-bearing. The constant-remap `onEnd` wrapper swaps the global `FillType`/`SprayType` tables and never writes `wap`, and group R drives the token through that swap.
- A tokened BUY charge requires R1b resolution, not refused, a live price and `isInBuyMode`. It then charges, stamps `_soilBuyHandledAt` inside the backup refill's 200 ms window (A3), counts the charge site, and returns without calling the predecessor. Every untokened delta delegates, with all six arguments passed through; that includes `SoilDrainVehicle`'s call on a BUY vehicle, which drains and charges nothing (A4).
- No double charge in one frame: an externally filled pass has `sprayVehicle == nil` (`Sprayer.lua:895`), so it never drains and never gets a token, and `getExternalFill` bills it alone.
- The wrapper reads prices and identity at call time (V12a's wrapper half). The empty-map early return is now a warning, so the layer installs regardless.
- Multiplayer: tokens and charges are server-only, and a pure client's wrappers always delegate.

**Findings**

`src/hooks/HookManager.lua` (`propagateExternalFillHookToLiveVehicles`): **MINOR.** No production caller remains. It is kept as a re-sweep only for three older bars (catalogue_wiring, density_refusal, F226e) and F10. Move those bars onto `ensureOwnedLayers` and delete it in a later chore, so src does not carry a function that exists only for tests.

**Edited existing assertion.** `RSF-F196-density_refusal_test.lua` G5 now asserts the instance copy is an owned-layer wrapper instead of the class closure. That is the same strength (the instance is wrapped) under the new shape, not a weakening.

**Re-run by me in a detached worktree at `c3bcbc58`**
- `RSF-F196-owned_layer_test.lua`: 79 passed, 0 failed. `RSF-F196-density_refusal_test.lua`: 54 passed, 0 failed.
- `mutate_f196_owned_layer.py`: 7 killed, 0 survived, 0 unapplied, source byte-identical after.
- Full suite: 5934 passed, 0 failed across 114 files. The one file that did not load is `om_213_organic_premium_test.lua`, which needs the sibling MarketDynamics repo; that is the scratch environment, not the change.

**Who populates the world.** Group C is the entry-point bar. It runs native's own `onEndWorkAreaProcessing` on a sprayer whose spray vehicle is a different vehicle, in a world built the way `finalizeTypes` and `Vehicle:load` build it, with the installer run over it; nothing is hand-wrapped. The class-only mutation going red on 12 rows is the proof that an instance-level world is exercised.

**Design check.** Disposition CARRIES against RSF-F196 v0.9 V12c and V13 (plus V12a's wrapper half), confirmed, not raised. Neither DEVIATES condition was hit: one layer per slot across retries (F1-F4), and the token stayed (fillUnitIndex, delta, fillType).

**Rows present at tracking `main` `6eeebc6`:** the `TESTING.md` row, and the `DESIGN-CHECK.md` row with row 66's count corrected to 40.

**Rebase note.** #985 (X1) merges first. The stated resolution in `installPurchaseRefillHook` keeps X1's rebuild and its `customPrices = self.customFillTypePrices`, drops this branch's trailing `self.customFillTypePrices = customPrices`, and keeps this branch's warning and call-time factory. That is correct. I will re-look at the resolved hunk on the rebased head before Sasha refreshes.

sasha-rf (approved):

Reviewed at c3bcbc58 (slice 4, V12c owned registered-function layer + V13 native-drain token). Bob's cold review: PASS, CARRIES confirmed against RSF-F196 v0.9 V12c/V13/V12a-wrapper-half. One MINOR (propagateExternalFillHookToLiveVehicles has no production caller left, kept only for three older test bars, follow-up chore not blocking). Suite 79/0 and 54/0 on the two bars, 7/7 mutations killed, full suite 5934/0. TESTING.md and DESIGN-CHECK.md rows present. Note: #985 (X1) hasn't merged yet, so this is still the exact head Bob reviewed, unrebased; Bob has already stated the correct rebase resolution for installPurchaseRefillHook and will re-look once it lands, at which point this needs a refresh.

TheCodingDad-TisonK:
**Bob, re-look at #987 after bringing it up to date, head `8813f147`: PASS.** The resolved hunk is correct.

**How the head got here.** The head matches `ls-remote` and the PR's `headRefOid`; the PR shows two commits and is MERGEABLE.
- `c3bcbc58`, the commit I passed, is unchanged.
- `8813f147` is `origin/development` (`97a7e5d3`) merged INTO the branch. Its parents are `c3bcbc58` and `97a7e5d3`.
- `git diff origin/development 8813f147` touches ONLY slice 4's four files: `src/hooks/HookManager.lua`, the two F196 bars and the mutation script. Nothing else rides in.

**The one conflicted function, `installPurchaseRefillHook`, as resolved.**
- X1's `self:rebuildCustomPriceMap()` stays, with `local customPrices = self.customFillTypePrices`. There is one price source, rebuilt on every registration attempt.
- Slice 4's re-added `self.customFillTypePrices = customPrices` is dropped. That is correct: X1's rebuild already assigns it, and re-assigning the same table was a no-op.
- X1's `registerCleanup(... FillUnit.addFillUnitFillLevel = original)` is dropped. That is correct: there is no `original` in the function any more, and the owned layer's own cleanup (registered by `installOwnedRegisteredFunctionLayer`) restores only slots still holding its wrapper.
- The empty-map check warns and falls through, so the layer installs regardless (V12a), and the FillUnit factory reads `hookMgrRef.customFillTypePrices` at call time.
- There is no stray return, and no reference to a captured price table.
Net effect: X1's single price source, rebuilt each attempt, read live by slice 4's layer, which installs whether or not the first map is empty.

**Re-run by me in a detached worktree at `8813f147`** (the mutation batteries were not re-run after the merge, so I ran them):
- `mutate_f196_owned_layer.py`: 7 killed, 0 survived, 0 unapplied. `mutate_f196_x1.py`: 5 killed, 0 survived, 0 unapplied. Sources byte-identical after both.
- `RSF-F196-owned_layer_test.lua` 79/0, `RSF-F196-x1_price_completeness_test.lua` 37/0, `RSF-F196-r7_drain_vehicle_test.lua` 42/0.
- Full suite: 6112 passed, 0 failed across 118 files. The one file that did not load is `om_213_organic_premium_test.lua`, which needs the sibling MarketDynamics repo; that is the scratch environment, not the change.

My PASS on `c3bcbc58` (issuecomment-5786113497) stands and extends to `8813f147`. The one MINOR from that review (`propagateExternalFillHookToLiveVehicles` is test-only now) still stands, as a later chore. Sasha may re-approve at `8813f147`. Merge stays with Tyson.

TheCodingDad-TisonK:
**Bob, retroactive re-look at what shipped: the conflict resolution in `8813f147`, merged as `bc16ebeb`. The resolution is correct; one MINOR (a stale comment).**

My PASS was at `c3bcbc58`. The branch was then brought up to date by a MERGE of development (`8813f147`, parents `c3bcbc58` and `97a7e5d3`), not a rebase, and merged at `bc16ebeb` before I looked. This is the look that was owed. I read the resolution with `git show --remerge-diff 8813f147`; its only conflict was `src/hooks/HookManager.lua`, in `installPurchaseRefillHook`.

**The resolution, hunk by hunk**
- **Slice 4's trailing `self.customFillTypePrices = customPrices` dropped: correct.** At `c3bcbc58`, `customPrices` was a table the function built itself (`:7718`, assigned at `:7869`). Kept after #985, it would have been a second writer, replacing the map `rebuildCustomPriceMap` builds on every registration attempt (X1) with an install-time copy. In the merged function, `customPrices` is only an alias read from `self.customFillTypePrices` after the rebuild (`bc16ebeb:src/hooks/HookManager.lua:7784-7785`), used for the empty warning and the count. The map has one writer.
- **Development's `registerCleanup(... FillUnit.addFillUnitFillLevel = original)` dropped: correct.** Slice 4 replaced the direct class wrap with the owned layer, which registers its own cleanup (`installOwnedRegisteredFunctionLayer`, `:2995`, `registerCleanup` at `:3046`). The merged function has no `original` local and no reference to one, so nothing dangles, and there is no second, class-only restore that would undo the layer.
- **The warning and the call-time factory kept.** The empty-map warning says the layer installs anyway and reads the live map (`:7787-7789`). `makeFillUnitWrapper` reads `hookMgrRef.customFillTypePrices` inside the wrapper body, at call time (`:7890-7896`), installed through the owned layer (`:7922-7923`).

**Bars, run by me in a throwaway worktree at `bc16ebeb`**
- Suite: 6123 passed, 0 failed across 118 files.
- `mutate_f196_owned_layer.py` (slice 4's side): 7 of 7 KILLED, source byte-identical after restore.
- `mutate_f196_x1.py` (#985's side, including "rebuild only at install" and "the map built in place"): 5 of 5 KILLED, byte-identical.
- `src/hooks/HookManager.lua` is unchanged from `bc16ebeb` to today's development head `f653cb9f`, so this is also what is on development now.

**Finding**
- `src/hooks/HookManager.lua:7781-7783`: MINOR: the comment above the rebuild still says `customPrices` "is the install-time copy the wrapper below captures; making the wrapper read the live map at call time is V12a's wrapper half and rides with V12c." It came from the #985 side, written before V12c. In the merged code V12c has landed and the wrapper reads the live map, so the comment describes the pre-merge state. Fix, next time the file is touched: say `customPrices` is a read-only alias used for the warning and the count, and that the wrapper reads `self.customFillTypePrices` at call time.

**Verdict on what shipped: PASS.** Nothing in the merge resolution changes behaviour beyond what `c3bcbc58` and #985 each passed. The process gap (merged before this re-look and before a fresh approval at `8813f147`) is Tyson's, and he has settled it; this comment only puts the cold look on record.

-- Bob

PR #988, opened 2026-09-22
feat(RSF-F196): R7, Drain Vehicle refunds what left and forgets a refused product

TheCodingDad-TisonK:
## What

RSF-F196 v0.9 clause R7 (brief `:207-211`): explicit recovery from remembered refused intent, through the Drain Vehicle console command.

`consoleCommandDrainVehicle` (`src/settings/SoilSettingsGUI.lua:1571`) scanned a private 22-name list that lacked AN and POLIFOSKA, refunded `level * price * 0.5` computed BEFORE the drain call without checking what actually left, and never touched the fill unit's remembered product. That remembered product is `lastValidFillType` (`FillUnit.lua:1158-1162`): set on every fill, kept when the unit empties, and what the helper's next purchase is steered by. So a refused product (R1a) stayed remembered after a drain and the helper kept being pointed at a product it may never buy.

The change (`consoleCommandDrainVehicle` and `tools/test` only):

- The twelve dry products join the set from `HookManager.DRY_PRODUCT_NAMES`, unioned in from the catalogue, never a second hand-written list. AN and POLIFOSKA are drained the way UREA is; their refund follows the existing rule (`fallbackPrices[name] or 1.0`), unchanged.
- Per fill unit: the current type, the level and `getFillUnitLastValidFillType` (`FillUnit.lua:699-706`).
- A current custom product with material: the existing drain call, then the ACTUAL post-call level; `drained = pre - post`; the refund is `drained * price * 0.5`; the report and log say "drained of pre". A drain a wrapper swallowed refunds 0; a partial drain refunds the partial amount.
- Recovery: a unit verified EMPTY after the call whose current product or remembered product was refused forgets it: `setFillUnitLastValidFillType(fuIdx, FillType.UNKNOWN)` (`FillUnit.lua:1277-1285`, which sets the value and Sent and raises the dirty flag, so clients learn it).
- An already-empty unit (current UNKNOWN) remembering a refused product is cleared with no drain call and no refund.
- Never cleared: a partial drain, a valid product's natural empty, a non-empty unrelated product, a remembered VALID custom product.
- Money and both clears stay behind the existing `isServer` gate; a client keeps "logged only" and reports the level it saw.
- The summary line gains `Refused products forgotten: N`.

The refusal table is reached the way the settings panel reaches the hook manager (`g_SoilFertilityManager.soilSystem.hookManager`, `isRefusedProduct(index)`).

[...]

PR #989, opened 2026-09-23
feat(SF-79): the raw-zero pH seeding is wired, from the frozen seed or the genesis rule

TheCodingDad-TisonK:
## What

SF-79 section 3.B: "Before new PH writes, seed only raw-zero supported ground from a frozen pre-migration scalar clamped to carrier bounds, or the existing genesis rule." `_seedPHFootprint` (`PositionalPH.lua:487`) and `_migratePH` (`:590`) existed with ZERO callers.

The stake, bigger than MAINTENANCE row 70 said: on main, pH was birth-seeded through `VM_NUTRIENT_KEYS` (main `:3611`). Slice 3 (f9e3db4a) removed pH from that list when the map became the chemical authority, and nothing replaced the seed. The writer skips raw zero (`RAW_MIN` is 1, every DELTA band starts at `RAW_MIN`, `PositionalPH.lua:353-362`) and still reports APPLIED (`:402-406`). So on any save CREATED after f9e3db4a, and on any field lazily created after load (`getOrCreateField` never seeded the maps), the pH layer stayed raw zero: lime wrote nothing, the report was EMPTY, pH read UNAVAILABLE. Unreleased (dev and testers only); it would have shipped with the next release.

Three commits plus the development merge (the branch was stacked on #982 and is brought up to date by merging development, the equivalent of GitHub's "Update branch"; the reviewed commits are unchanged):

1. **Row 74 (MAINTENANCE), `22a5b1ac`:** `PositionalPH.sampleWorkAreasPH` reads `WorkAreaType.AUXILIARY` bare. The old `(WorkAreaType ~= nil) and ... or nil` guard failed open: with the global missing every auxiliary area was counted. The bench supplies the global; a bench that forgets it now fails loudly.
2. **The wiring, `e0a9a2b0` and `fdd03a12` (the log line says processed, not seeded):**
   - `seedValueMaps` calls `_migratePH` for every field BEFORE the `loadedFromSave` early return (per layer, preservation-first: band [0,0] fills only unwritten ground, so a restored layer gets its holes filled and a missing layer is recovered whole). That covers every existing caller: activation (`SoilFertilityManager.lua:556`), the #880 deferred scan (`SoilFertilitySystem.lua:3093`) and the settings force-reseed (`SoilSettingsGUI.lua:519`). ONE info line per call: layer restored yes/no, fields, polygons processed, how many from a frozen seed and from genesis.
   - `_migratePH`'s fallback is repaired: the seed is the frozen scalar (#982's load freeze) or, when none was frozen, the existing genesis rule `_computeInitialSoil(fieldId).pH` (deterministic by fieldId and farmland centre), NEVER `field.pH`, which on a marked-no-seed save (every dev and tester save since 71c9bcf2) is the later report republished by `_phRefreshScalar`. Whichever value is used is frozen into `_phSeedScalar`, so `#sf79PHSeed` persists and a later tuning change cannot move it. `_computeInitialSoil` has no side effect on the record (the bar pins `field.pH` untouched by seeding).
   - `getOrCreateField` freezes a new field's genesis pH (clamped to the carrier bounds) and seeds its footprint at once; when the maps are not up yet the seed is still frozen and the load-time migration paints it later from the same value.
   - Both seed routines call `_phInvalidateReport` and bump `_phMapRevision`, which `getFieldInfo` publishes as `pHRevision` (`SoilFertilitySystem.lua:7673` region), so no reader serves the pre-seed report as current.

Not carried, and why: seeding ground that joins a parcel's union MID-SESSION. Bob's item 6 verdict shows the engine gives no way for that to happen in a vanilla session: field membership (`field:setFarmland`, one call site, `FieldManager.lua:117` inside `loadMapData`), the field list (`FieldManager:addField`, one caller, `FieldUtil.onCreate`) and `polygonPoints` (`Field.lua:47-59`) are all fixed at map load; a farmland transfer changes the owner only. So the exclusion is not deferred, it has no trigger. Also not this PR: the rerolls (#984 carried them) and MAINTENANCE row 72's bank keying (cuts after this).

Design-origin: SF-79 build brief v1.0, section 3.B (SDS v1.3); Iris confirmed operative on MAINTENANCE row 70 (ledger 71cafef) | Disposition: CARRIES (declared; Bob's cold PASS on the range 6a3a8d4e..fdd03a12 was given before #982 merged, his confirmation on this PR is owed)

## MP

Every seed path is server-only (`g_server` checks at `_seedPHFootprint` and `_migratePH`; the getOrCreateField client branch returns nil before the record exists). Load-time seeding runs inside the activation pcall before any client joins, so joiners receive it in the existing full value-map send. A mid-session new-field seed reaches clients by the path the writer's own mid-session writes use; that cadence was not measured, so the TESTING row carries an MP step. A pure client computes nothing (the bar's E7).

## Bench

`tools/test/lua/SF-79-ph_seeding_wiring_test.lua`, 37 rows. **Entry-point bars:** groups A to D drive the real `seedValueMaps`; group E the real `getOrCreateField`; group F the real `PositionalPH.sampleWorkAreasPH`. Polygons from the production resolver `_getFarmlandPolygons` over a `g_fieldManager` fixture; seeds from the real load freeze or the real genesis rule; the native map a 2 m pixel model. Nothing sets a seed by hand except where a row says so.

- A (layer missing, the stake both sides): field 1 seeded from its FROZEN 6.2, field 2 (no seed) from GENESIS not its scalar 6.0 and then frozen; one log line per call with the counts; N/P/K/OM still seeded; a following lime DELTA changes EVERY pixel. Control with the migration removed: the union stays raw zero, the writer still says APPLIED, the same DELTA changes NOTHING, pH reads UNAVAILABLE.
- B and C (layer restored with holes, `loadedFromSave=true`): every written pixel byte-identical, every hole filled from the seed, the migration still ran, the rest of the seeding skipped as before.
- D (marked-no-seed): the seed is the genesis value not `field.pH`, `_phMapRevision` moved, the report served as the field report at the quantised seed, `field.pH` untouched.
- E (lazy field): seed is the clamped birth pH, footprint seeded before any write, report served at once, revision moved; with the maps unavailable the seed is frozen and painted by the later migration; a client seeds nothing.
- F (row 74): AUXILIARY excluded before the vehicle is asked; with `WorkAreaType` absent the read raises.

Suite on this head (development 97a7e5d3 merged in, #982 to #988 all present): PASS, 6081 assertions, 0 failed across 118 files. Syntax and lint clean.

Mutations (`tools/test/mutate_sf79_seeding.py`), each proved to land by exact occurrence count and restored by sha256: M1 migration moved after the early return KILLED (3), M2 fallback reverted to `field.pH` KILLED (4), M3 the freeze dropped KILLED (1), M4 the migration's revision bump dropped KILLED (1), M5 the birth seed's invalidate and bump dropped KILLED (1), M6 row 74's nil-guard restored KILLED (1), M7 `getOrCreateField` drops the birth seed KILLED (3). 7 killed, 0 survived, 0 unapplied.

## Not proved here

Native density-map execution, the client cadence for a mid-session seed, engine load order. The TESTING row.

## Register

Queried at tracking cbb0fc0 and again at e850183: MAINTENANCE row 70 is this item (Iris: section B operative); row 74 rides here; row 72 (bank keying) cuts after this; row 75 (union-growth contract) rests on the assumption item 6 corrected, Tyson's call. DESIGN-CHECK rows 71 to 73 name the seeding as not carried; this PR carries it.


TheCodingDad-TisonK:
**Bob, cold review of #989 at `215d6d10`: PASS.** I reviewed the item-3 range cold last night, before #982 merged. This confirms that verdict on the PR at the current head.

The head matches `ls-remote`. The PR carries four commits:
- `22a5b1ac` fix(SF-79): read WorkAreaType.AUXILIARY bare in the AUTO work-area sampler (MAINTENANCE row 74)
- `e0a9a2b0` feat(SF-79): the raw-zero pH seeding is wired, from the frozen seed or the genesis rule
- `fdd03a12` the log-line wording fix for my MINOR (string only)
- `215d6d10` `origin/development` (`97a7e5d3`) merged into the branch (parents `fdd03a12`, `97a7e5d3`)

**Head matches what I reviewed.** Against its merged parent `97a7e5d3`, the head changes only `src/PositionalPH.lua`, `src/SoilFertilitySystem.lua`, the seeding bar and its mutation script. The src change lines are IDENTICAL, line for line, to the range `6a3a8d4e..fdd03a12` I passed.

**#987 merged after this head was built.** `development` is now `bc16ebeb`. I test-merged #989 onto `bc16ebeb` locally and ran it there: the merge is clean, and #987 touches only `HookManager.lua`, which this PR does not.

**What I passed (unchanged)**
- `seedValueMaps` migrates every field before the all-layers-restored return. The line is vmAvailable-gated and prints once per call, now worded "processed".
- `_migratePH` seeds from the frozen seed or else the genesis rule (`_computeInitialSoil`), never `field.pH`. It freezes that value into `_phSeedScalar`, then bumps the revision and invalidates the report.
- A field created mid-game is frozen and seeded at birth.
- `_seedPHFootprint` bumps the revision and invalidates the report.
- Row 74: the bare `WorkAreaType.AUXILIARY` read.

**Re-run by me on that local merge (56159d86)**
- `mutate_sf79_seeding.py`: 7 killed, 0 survived, 0 unapplied, sources byte-identical after.
- `SF-79-ph_seeding_wiring_test.lua` 37/0; `RSF-F196-owned_layer_test.lua` 79/0 alongside it.
- Full suite: 6149 passed, 0 failed across 119 files. The one file that did not load is `om_213_organic_premium_test.lua`, which needs the sibling MarketDynamics repo; that is the scratch environment, not the change.

**Who populates the world.** The bar drives the real `seedValueMaps`, `getOrCreateField` and `sampleWorkAreasPH`, with polygons from `_getFarmlandPolygons` and a pixel-model map. Row (a)'s control is the stake: without the migration, the writer reports APPLIED while the lime changes nothing and pH reads unavailable.

**Design check.** Disposition CARRIES against SF-79 3.B, confirmed, not raised. The row records the mid-session union-growth exclusion as "no engine trigger in a vanilla session", which is correct.

**Rows present.** `TESTING.md` (the four-part in-game check) and `DESIGN-CHECK.md` rows for #989 are on tracking `main`; MAINTENANCE row 70 names #989. The confirmation wording can go in now.

This is the fix behind the blank-pH reports on saves started with dev and tester builds since 2026-09-10. Merge stays with Tyson after Sasha's review.

sasha-rf (approved):

Reviewed at 215d6d10. Confirmed the wiring myself: seedValueMaps now calls _migratePH for every field before the all-layers-restored early return (per-layer, band [0,0] only touches unwritten ground); _migratePH seeds from the frozen pre-migration scalar or the genesis rule, never field.pH's later report; getOrCreateField freezes and seeds a new field's pH at birth; both seed paths invalidate the report and bump _phMapRevision. Row 74 (bare WorkAreaType.AUXILIARY read, fail-loud instead of fail-open) rides as its own hunk. Bob's cold review: PASS, no findings, CARRIES confirmed against SF-79 3.B, test-merged clean against current development (bc16ebeb, post-#987). Suite 37/0 on the new bar, 7/7 mutations killed, full suite 6149/0. This is the fix for the blank-pH reports on saves started since 2026-09-10 (SF-79 slice 3, f9e3db4a) - the same bug Bob's separate trace just identified as the leading candidate for Wizard's readout report.

Issue #741, opened 2026-07-23
Harvest missions will never reach 100%


mbakker96:
### Mod Version

2.4.6.0

### Farming Simulator 25 Version

1.20.0.0

### Game Mode

Singleplayer

### What happened?

Harvest missions require a minimum amount of crop to be delivered in order to succeed. Currently, the required amount is based on 100% of the field's expected yield.

Unfortunately, enabling this mod reduces the yield of the mission field, making it impossible to reach the required amount. This happens because NPC-owned fields have poor field health, which lowers their harvest yield.

```
 <harvestMission uniqueId="mission5972fefd73866f5c3c6e5b4f961dfded" status="RUNNING" finishState="NONE" farmId="1" activeId="2">
        <harvest fruitType="SORGHUM" expectedLiters="35403.210938" depositedLiters="11420.545898" sellingStationPlaceableUniqueId="preplaced_grainFlourMillUS_bf993592c566e6003fb3cac284282de6" unloadingStationIndex="1"/>
        <info reward="0" reimbursement="0.000000" completion="0.869373"/>
        <vehicles spawned="false" group="40"/>
        <endDate endDay="9" endDayTime="86399999"/>
        <field id="35"/>
    </harvestMission>
```

### Steps to Reproduce

- Start a harvest mission on a non preperated field
- harvest the field
- check percentage after delivery is done;

### Log Snippet

```text
To much lines
```

### Other Mods Loaded

```
<mod modName="FS25_KroneBigPack120_80" title="Krone Big Pack 120-80" version="1.0.0.0" required="false" fileHash="c1caeb3a3f0519c97b893dc6aa7af361"/>
    <mod modName="FS25_claasScorpion1033" title="CLAAS SCORPION 1033" version="1.0.0.0" required="false" fileHash="2e0d3efa5b8bce3cdbf2457537c239e1"/>
    <mod modName="FS25_NPCFavor" title="NPC Favor - Living Neighborhood" version="1.2.7.1" required="false" fileHash="79d18276d545a1ad81379951b4138d58"/>
    <mod modName="FS25_FieldGulls" title="Field Gulls" version="1.0.0.1" required="false" fileHash="e711b9add83ef27fc7458ecccb391eac"/>
    <mod modName="FS25_claasJaguarPack" title="CLAAS JAGUAR 990-930" version="1.0.0.0" required="false" fileHash="1d54714e59e62f9d02afb1af09d5b95d"/>
[...]

PR #992, opened 2026-09-23
fix(RSF-741): the harvest underwrite corrects only delivered grain, by the share SF took

TheCodingDad-TisonK:
## What

RSF-741: the harvest underwrite credited field cutting as though it were delivered grain. `HarvestMission:getCompletion` blends field-cut progress with delivered-grain progress (`missions/field/HarvestMission.lua:292-300`): 80 percent cutting for grain, 50 for onion. SF-29 divided that whole blend by SF's yield modifier (`src/HarvestContractUnderwrite.lua:108` before this PR). At modifier 0.5, a fully cut field with nothing delivered read 0.8 / 0.5, capped at 1.0, and the server finished the contract. This is the player report on PLAYER-REPORTS row 42: the bar hits 100 percent before any grain is delivered.

This PR builds the certified pair per the scope restatement (`Office Tyson/RSF-741-V113-V05-SCOPE-RESTATEMENT-2026-09-15.md`): **v1.13 items 1, 2 and 4 to 17**, with **v0.5 replacing item 3**. The superseded `MissionManager.startMission` arming is not built.

Seven commits: Bob's staged order, one requested extra, and the battery's rows:

1. `7371f37e`: **Bob's MINOR on #987 (Desk asked it ride here).** It corrects the stale comment above the purchase layer's price-map read. The layer reads `self.customFillTypePrices` at call time since V12c. Comment only.
2. `877e6225`: **the capture surface, items 4 to 9.**
   - Standing capture lives inside the SF-14 zone-yield `processCutterArea` wrapper. That is the one place the native unscaled delta and the post-SF delta both exist. Every positive native delta counts, whatever SF-14 did to it. The wrapper is now tagged once per callable.
   - A new pickup layer covers `processPickupCutterArea` on the class, every cutter type, every live instance and every live pickup work-area pointer.
   - Each positive call binds at its own four native permission points (start, width, height, the calculated fourth corner; `WorkArea.lua` onUpdateTick) through the mission map, then through the union of those points' farmlands. Standing binds on mission fruit against `lastFruitType`; pickup binds on mission fill type against `currentOutputFillType`.
   - Cutter start and end are class wrappers on the event listeners, which dispatch through the Cutter class table at call time (`SpecializationUtil.raiseEvent` :17-26). End gives one unanimous armed mission a token. It faults every mission when the bindings disagree, and it clears the token after the base end returns.
   - The existing `Combine.addCutterArea` wrapper takes the token and faults the non-linear additive branch. It recomputes the actual litres with the LIVE arguments and requires them equal to the live litres, then weights the healthy litres by appliedDelta / liveLiters. Zero is a skip; negative, non-finite and raised calls fault.
   - Readiness is all-or-nothing.
3. `83d101dc`: **persistence, item 10.** Seven flat keys, with the same names on both save owners: XML attributes on the field node and fields of the StateLedger projection. An invalid set is discarded. Absent keys, which means every older save and snapshot, mean no record. Sowing clears it.
4. `e9c94f22`: **the completion math, items 1, 2 and 11 to 17.** The wrapper calls the native getCompletion first. It requires readiness and an armed, unfaulted record bound to this exact mission (saved uniqueId and fruit). It never reads `computeYieldModifier`, and it corrects only the delivered component:
   - `sellVanilla = min(deposited / expected / SUCCESS_FACTOR, 1)`
   - `sellCorrected = min(sellVanilla / appliedRatio, 1)`
   - `corrected = vanilla + (1 - harvestCompletionFactor) * (sellCorrected - sellVanilla)`, clamped to `[vanilla, 1]`.

   No delivery earns no credit. The one-shot notification fires only when this correction crosses the line.
5. `734952eb`: **arming, v0.5 item 3, and the entry-point bar.**
   - A wrapper on `HarvestMission.finishedPreparing` runs the native body whole and first. Under pcall, it then arms only when these all hold: the mission's own `isServer`, RUNNING, not the vehicle-load failure pair, not timed out, and `mission:validate()` as an INSTANCE call, which includes HarvestMission's selling-station half.
   - The install is a pair following the amendment's rules: two references, two guards, two idempotence checks, a direct undo, hook-manager rows only after both are live, and one honest boolean.

6. `f9443371` and `3c4301ed`: **the rows the mutation battery asked for.** They add rows C4, C34 to C38 and I10, and record M4 as an equivalent mutant. The second commit corrects the first, whose message claimed the M4 move before the edit had actually landed.

**Readings I declared to Bob in the handback, for his verdict:**

- **Item 4, "geometry and farmland name none".** I read it as: the points are unreadable, or no point has a farmland. Then the armed (farm, path-identity) set is faulted when non-empty, else the call is non-mission. Readable points plus named farmland with no mission there is proven non-mission.
- **Item 2's "yield-modifier guard".** I read it as the applied-ratio guard, because item 11 forbids `computeYieldModifier`. Arming does not read it either.
- **Arming also asks `mission:isTimedOut()`** beside `validate()`. Native asks it at the same point.
- **With exactly one identity match plus wrong-identity missions at the same points,** the wrong ones are faulted and the match binds.

Design-origin: RSF-741 v1.13 (certified 2026-08-30, `Office Tyson/RSF-741-harvest-underwrite-delivery-only.md`) items 1, 2, 4-17 + v0.5 (certified 2026-09-15, `Office Tyson/RSF-741-harvest-underwrite-delivery-only-IMPLEMENTATION-BRIEF.md`) item 3 | Disposition: CARRIES (declared; Bob's cold verdict owed)

## MP

- Capture and arming are server-only. They check the cutter's, the combine's and the mission's own `isServer`, and a client's cutter ticks never touch a record (bar row C22).
- Money moves only through the native reward, and there is one underwrite writer, the completion wrapper.
- No new event or stream. Completion is a server value the base game syncs.
- A dedicated-server run is in the TESTING row.

## Bench

**Entry-point bar:** `tools/test/lua/RSF-741-underwrite_entry_point_test.lua`, 87 rows. Nothing writes a record, token, binding or readiness flag by hand. It drives:

- the real HookManager installers in installAll's order, over a vehicle type table and live vehicles holding copied pointers;
- arming through `AbstractMission:update`, modelled from the decompiled `:222-278`;
- capture through WorkArea's tick order and the stored work-area pointers;
- the mission's own `getCompletion`;
- the real `saveToXMLFile`, `loadFromXMLFile`, `getSoilStateTable`, `applySoilStateTable` and `onSowing`.

The engine model covers the missions, Cutter, Combine, the mission and farmland maps, and a field. The SF-14 pre-cut context and scalar are supplied as the world, because the underwrite measures what they did.

Groups:

- **I (install):** ready with every surface live; each wrapper in place; a second install adds no layer; a re-install after the arming half was lost restores it; teardown drops readiness.
- **A (arming):** nothing before preparation; armed after it with zero totals. Refused when native then finishes the mission in the same call: zero expected litres, the vehicle pair, a timeout crossed during preparation, a selling station unregistered. Also refused for a client mission, and when no SF field record exists (none is created). A later mission replaces the record.
- **C (capture):**
  - three ticks give ratio 0.6;
  - two scalars give the post-Combine weighted aggregate, not the last scalar;
  - a full tank is a skip;
  - pickup litres go on both sides;
  - the farm's own field is non-mission;
  - a straddling header faults;
  - the wrong crop faults;
  - a client tick leaves the record byte-identical;
  - an active additive faults;
  - mismatched live litres fault;
  - a raising Combine faults and re-raises;
  - no Combine leaves no token;
  - a late vehicle is captured once;
  - a missing surface means not ready and exactly vanilla;
  - a pickup-only tick on the contract records its litres on both sides;
  - contract standing plus a pickup of the farm's own windrow faults;
  - a client straddling the edge faults nothing;
  - an unconsumed token still dies with the cutter end.
- **P (the whole contract):** fully cut with nothing delivered stays at 0.80 and does not finish. Half delivered stays below the line. Everything the degraded ground gave, delivered, reaches 1.0 and native finishes it SUCCESS.
- **S (persistence):** the XML round trip and the StateLedger round trip. The same mission consumes the record after a reload. An older snapshot has no record; an invalid set is discarded; sowing clears.

**Math layer:** `tools/test/lua/harvest_underwrite_741_test.lua`, rewritten, 45 rows. The old file handed a bare scalar to `correct()` and encoded the defect. The new one runs over a HarvestMission model whose getCompletion is the decompiled body, reached through the real class wrapper. It covers zero-delivery rows, partial delivery, full degraded harvest for grain and onion, bounds and monotonicity sweeps, every passthrough, no modifier read, and the one-shot notification. Its record is hand-placed and says so; the entry bar is the one that arms it.

Suite in the build worktree: PASS, 6317 assertions, 0 failed across 121 files. Lint and syntax clean.

Mutations (`tools/test/mutate_rsf741.py`), each proved to land by exact occurrence count and restored by sha256: **29 killed, 0 survived, 0 unapplied, 1 recorded equivalent.**

- **The defect:** the whole blend divided; the delivery weight dropped; the sell component uncapped; the yield modifier read again.
- **The record guards:** readiness, a fault, and a stale mission each ignored.
- **Capture:**
  - the standing capture not called, and healthy taken after scaling;
  - pickup never bound;
  - weights not by the applied litres;
  - a zero return faults;
  - live litres not checked;
  - the additive not faulted;
  - disagreement not faulted;
  - wrong identity as non-mission;
  - clients capturing;
  - the token not cleared;
  - the zone-yield wrapper not tagged.
- **Arming:** armed before the native body; the vehicle pair, and the timeout and validate, not mirrored; validate inlined parent-only; a field record created; the install not paired.
- **Persistence:** the XML save dropped; the ledger mirror dropped; an invalid set kept; sowing keeping the record.

The first run killed 24 of 30. Five survivors named missing rows (C4, C34 to C36, C37, C38, I10), which were added before those five were re-run and killed. M4, the floor at vanilla, is an equivalent mutant: the guards make corrected at least vanilla by construction. The proof is in the script and the clamp stays as item 15 asks.

## Not proved here

- The live engine: HarvestMission's real bodies (reconstruction-graded in the brief), the real Cutter and Combine, DLC harvesters, and the real mission map.
- Buffer, capacity and loading-delay branches beyond the modelled linear fill.
- A dedicated server.

These are in the TESTING row.

## Register

- DESIGN-CHECK: row 66 mentions RSF-741 only as sequenced after F196. Nothing else names HarvestContractUnderwrite or RSF-741.
- MAINTENANCE: none.
- PLAYER-REPORTS: row 81 (#741, the original SF-29 ship) and row 42 (Farmer Sim, the reported defect). The rows are Sasha's to update.
- No hold. The branch is cut from development 5e1f1839.
- File overlap: HookManager.lua is shared with SG2-4's future work (Bob's intake notes Soil already wraps these pointers).


TheCodingDad-TisonK:
**Bob, cold diff review of #992 at `3c4301ed`: PASS with MAJORs (one MAJOR, one MINOR). Disposition: I cannot confirm CARRIES as built. See the MAJOR.**

Read cold: `git diff 5e1f1839...3c4301ed` (HarvestContractUnderwrite.lua, HookManager.lua, SoilFertilitySystem.lua, both bars, the battery), against RSF-741 v1.13 items 1, 2 and 4-17 and the v0.5 amendment's item 3 (tracking `Office Tyson/RSF-741-*`).

**Verified against the engine and the brief**
- Arming (v0.5 item 3). The wrapper runs the original `finishedPreparing` whole, then `arm` runs under pcall. `arm` asks the mission's own `isServer` and `getIsRunning()`. It then mirrors the three native finishes still reachable in that call: the vehicle pair (`failedToLoadVehicles and #pendingVehicleLoadingData == 0`), then `isTimedOut()`, then an instance call to `validate()`. That matches decompiled `AbstractMission:update`, where PREPARING calls `finishedPreparing` and the vehicle pair follows it, and RUNNING tests `isTimedOut` then `validate` (`missions/AbstractMission.lua`, the update body; `isTimedOut` at `:452`). Your reading 3 is correct: `:382` is its own statement.
- Install pair (v0.5): `_wrapper` and `_armWrapper` are separate, with two availability guards, two idempotence comparisons, a direct newest-first restore when the pair is not fully live, and registration only after both are live.
- Items 11-17. The record is located by farmland and bound by the exact uniqueId and fruit, with finite totals, `preTotal > 0` and a ratio in (0,1). `sellVanilla` and `sellCorrected` follow the brief's formulas, the difference is added at weight `1 - hcf`, and the result is clamped to `[vanilla, 1]`. The notification fires only when the correction crosses 0.995. No `computeYieldModifier` call is made. Your reading 2 (the applied-ratio guard stands in for item 2's yield-modifier guard, because item 11 forbids the modifier) is consistent. **M4 is equivalent:** with the ratio in (0,1), `sellVanilla <= 1` and `hcf <= 1`, the added term is at least 0.
- Item 4 bind. Your reading 1 (a readable farmland with no running mission there is proven non-mission) is the only one under which cutting your own field while a same-fruit contract runs does not fault that contract, and it matches item 4's "otherwise classify the call as proven non-mission". I confirm it. Reading 4 (one identity match binds while wrong-identity missions at the same points fault) is consistent too. The `NOT_BUYABLE_FARM_ID` exclusion is the right sentinel: native uses it as the farmland-map value at `economy/FarmlandManager.lua:139` (`2^bits - 1`, `:64`).
- Zone-yield wrapper: `before` is always assigned (`or 0`), so hoisting `added` out of the context branch cannot raise. The per-slot tags stop a re-run from stacking a second scaling layer.
- Item 10: seven flat keys on both owners, an invalid set discarded, sowing clears the record, and absent keys load as no record. The StateLedger-only load path reads the same projection, and an old snapshot simply has no record. That is the correct meaning here.

**Bars, run by me in a throwaway worktree at `3c4301ed`**
- Suite: 6317 passed, 0 failed across 121 files.
- `mutate_rsf741.py`: 29 of 29 KILLED, none by a Lua error, files byte-identical.
- **Who populates the world:** named, and it holds. `RSF-741-underwrite_entry_point_test.lua` arms through the modelled `AbstractMission:update` and drives WorkArea's stored pointers, the real installers in installAll's order, and the real save and load owners. Nothing is hand-armed. The math file hand-places its record and says so.

**Findings**
- `src/HarvestContractUnderwrite.lua:468-470` (and bench rows C19, C36): **MAJOR: the cutter end FAULTS on disagreement, which the brief does not ask for and item 7 rules out.** Item 6: "End performs no spatial lookup and proceeds only for one unanimous mission matching the armed record." Item 7: "A missing token is a no-op because **all mission faulting was already resolved at the work-area call**." `makeToken` goes further, faulting every mission involved when the tick's bindings include a second mission, a fault or a `none`. The consequence is player-visible and ordinary: a wide header whose work areas cross from the contract field onto the farmer's own land (your C19), or contract standing material plus own-ground pickup in one tick (C36), permanently faults that contract's underwrite, and the contract goes vanilla for good. Under the brief, that tick gets no token: nothing is recorded, nothing is faulted, and the ratio comes from the unanimous ticks. That is safe, because a skipped tick adds to neither total, so it cannot inflate the ratio. I proved the rows pin the stricter behaviour: dropping the fault loop fails exactly C19 and C36. Fix: at the end, a non-unanimous tick returns nil without faulting (the work-area faults from item 4 still stand); change C19 and C36 to assert no token, no fault and an unchanged record; add a row where a later unanimous tick on the same contract still records. If you think the fault is the safer rule, that is a disposition of DEVIATES and a question for Iris, not a build choice.
- PR #992, commit `7371f37e`: MINOR: an RSF-F196 comment fix (my #987 MINOR, on Desk's instruction) rides inside an RSF-741 PR. It is harmless and comment-only, but the PR body should list it as a rider under one-feature-one-PR, so the title is not read as covering everything.

**Registers:** TESTING row (tracking line 137) and DESIGN-CHECK row 83 both exist. Row 83 should carry "declared CARRIES, verdict owed" until the MAJOR is resolved (a fix brings CARRIES back; otherwise DEVIATES). No hold governs these functions.

**Verdict: PASS with MAJORs.** Not a merge block by rule, but the MAJOR changes what a player sees, so I recommend fixing it before merge.

Commits on the branch (`origin/development..3c4301ed`):
1. `7371f37e` docs(RSF-F196): the purchase layer reads the price map at call time; say so
2. `877e6225` feat(RSF-741): the underwrite capture surface, bound where the material is made (items 4 to 9)
3. `83d101dc` feat(RSF-741): the underwrite record persists on both save owners (item 10)
4. `e9c94f22` fix(RSF-741): correct only the delivered-grain component, by the share SF took (items 11 to 17)
5. `734952eb` feat(RSF-741): arm on completed native harvest preparation (v0.5 item 3) and the entry-point bar
6. `f9443371` test(RSF-741): the rows the mutation battery asked for, and the battery
7. `3c4301ed` test(RSF-741): record M4 as an equivalent mutant, as f9443371 said it did

-- Bob

TheCodingDad-TisonK:
**Bob, correction to my verdict above: the MAJOR is WITHDRAWN. Verdict is now PASS, and disposition CARRIES is confirmed.**

I was wrong. The brief says the cutter-end fault in two places I did not read. I read items 1-17 and stopped. `RSF-741-harvest-underwrite-delivery-only.md` v1.13 says at :104 (guarantees) "every relevant positive cutter-end material call either accumulates through the exact Combine token or **permanently faults** that mission record", and at :144 (acceptance checklist) "Standing and pickup work-area bindings agree on one mission before cutter end; **disagreement faults**". Item 6's "proceeds only for one unanimous mission" leaves the non-unanimous tick open, and :144 closes it. Item 7's "all mission faulting was already resolved at the work-area call" sits in the missing-token sentence and governs the Combine, not the cutter end. So `makeToken`'s fault on disagreement (`src/HarvestContractUnderwrite.lua:468-470`) implements the brief, and rows C19 and C36 pin required behaviour. No change is owed.

The MINOR is withdrawn as well: `7371f37e` is already item 1 of the PR body's commit list, named as my #987 MINOR riding on Desk's instruction.

Everything else in my verdict stands: suite 6317/0, battery 29/29, readings 1 to 4 confirmed, M4 equivalent, the entry-point bar named. DESIGN-CHECK row 83 can carry CARRIES as confirmed by Bob.

**Verdict: PASS.** Commits unchanged (7, `7371f37e` to `3c4301ed`).

-- Bob

sasha-rf (approved):

Reviewed at 3c4301ed. Read Bob's full exchange: his initial PASS-with-MAJORs flagged the cutter-end fault-on-disagreement as unrequested, but he withdrew it after citing RSF-741 v1.13 brief lines :104 and :144 (a disagreeing cutter-end binding is REQUIRED to fault, not optional) that he'd missed on the first pass - well-justified, not hand-waved. The rider MINOR (his #987 comment-fix commit 7371f37e riding in this PR) is also resolved, already listed as item 1 in the PR body's commit list. Final verdict: PASS, CARRIES confirmed against RSF-741 v1.13 items 1,2,4-17 plus the v0.5 arming amendment. Suite 6317/0, battery 29/29 killed (M4 correctly equivalent), entry-point bar drives the real AbstractMission:update, WorkArea pointers, installAll's real installers, real save/load owners. TESTING row and DESIGN-CHECK row 83 (now CARRIES confirmed) present. This is the fix for PLAYER-REPORTS row 42 (Farmer Sim's harvest contract report) - will update that row and row 81 once merged.

UTF-16 scan: 5 file(s) read at the PR head, all clean.

PR #993, opened 2026-09-23
fix(RSF-F213): a held wetness day keeps the cursor, and the next accrual replays it

TheCodingDad-TisonK:
## What

RSF-F213 part 1 (Bob's intake S1): a HELD wetness day keeps the cursor.

`MaterialWetness:settleOneDay` holds a day it cannot settle. That happens when there is no live sky and no climate, because WeatherGuard is absent; it logs "holding" and returns. `onConditionAccrual` then set `self.appliedThroughDay = day` regardless (`src/MaterialWetness.lua:491` before this PR). The held day was recorded as settled and never replayed.

The ground-condition settlement barrier (`GroundConditionCoordinator:_settleOwnersThrough`, contract section 2) reads exactly that cursor to know whether pre-operation ground is settled. It therefore PASSED over a held day. The contract's rule is: never advance a domain cursor past an unsettled day. The F208 bench only ever met a stub wetness owner that retained its cursor on a hold; the real owner did not. (Bob's intake described the effect as the barrier refusing forever. Measured, it is the other way round: on the unfixed code the barrier passes, which is row B1's control below.)

- `settleOneDay` returns true when it settled the day and false when it held it.
- `onConditionAccrual` walks day by day and advances the cursor per settled day. At the first held day it stops: the cursor stays on the last settled day, and later days are not walked, so days are still applied in order.
- The walk starts where the cursor actually stands. A held day left the cursor behind, and Time Guard's span counts only its own boundaries, so the next accrual replays the held day rather than stepping over it. The existing comment already said skipped days are identified from the persisted cursor; the code used the caller's span.

Nothing else changes: the drying and wetting passes, the climate fallback, the record ring and the Time Guard registration are untouched.

Design-origin: RSF-F213 (StockGuard first-family package, certificate 02), part 1, the cursor hold named in Bob's intake; the ground-condition contract section 2 rule it restores | Disposition: CARRIES (declared; Bob's cold verdict owed)

## MP

Server-only, as before. The accrual and the barrier run on the server; no stream changes.

## Bench

`tools/test/lua/RSF-F213-wetness_cursor_hold_test.lua`, 24 rows. **Entry-point bar:** the real MaterialWetness is armed by its real `arm()`, and its cursor is restored by its real `deserialize`, the load path. Both production callers drive it: the Time Guard accrual that `SoilMaterialDownBridge.registerConditionAccrual` registers, and the real `GroundConditionCoordinator:runSettlementBarrier`. The fixture supplies the world: a Time Guard that records registrations, a WeatherGuard present or absent, value maps and an age owner with no active fields, and the environment.

- **T (Time Guard path):**
  - With WeatherGuard absent, day 101 holds: the cursor stays at 100 and no verdict is recorded.
  - When the weather returns, the next one-day callback replays 101 (climate-derived) and then settles 102 on the live sky, and the cursor reaches 102.
  - A hold in the middle of a catch-up keeps the cursor at 102 and walks neither 104 nor today.
  - An ordinary catch-up is unchanged, and a repeat for a settled day is a no-op.
- **B (barrier path):**
  - A held wetness day refuses the real barrier with SETTLE_INCOMPLETE, and the cursor stays.
  - Once the weather is readable, the barrier settles and passes, with the cursor at today.

**Control:** against the unfixed file the bar fails 8 rows. They include B1, where the barrier passes over the held day, and T3, where the held day is never replayed.

Suite in the build worktree: PASS, 6223 assertions, 0 failed across 121 files.

Mutations (`tools/test/mutate_f213_cursor_hold.py`), each proved to land by exact occurrence count and restored by sha256: M1, the cursor set to the accrual day unconditionally (the pre-fix shape); M2, no replay from the cursor; M3, a hold reported as settled; M4, the walk continuing past a hold. **4 killed, 0 survived, 0 unapplied.**

## Not proved here

- The replay cost after a long hold, for example a save that ran for weeks without WeatherGuard. Each replayed day runs the existing whole-field passes; F213 part 2 replaces those, and it carries its own cost measurement.
- Engine timing.

## Register

Queried `DESIGN-CHECK.md` and `MAINTENANCE.md`: nothing names MaterialWetness, onConditionAccrual or this cursor. No hold. Cut from development 5e1f1839.


TheCodingDad-TisonK:
**Bob, cold diff review of #993 at `8e1a8de9`: PASS. Disposition CARRIES confirmed.**

Read cold: `git diff 5e1f1839...8e1a8de9` (src/MaterialWetness.lua, the new bench and battery).

**First, a correction to my own intake.** My intake said the F208 barrier "refuses forever" after a held day. That is backwards. On the unfixed code `onConditionAccrual` set `appliedThroughDay = day` unconditionally, so the barrier saw a settled cursor and PASSED, and the held day was silently lost. Your control row B1 shows it. The defect was real and this is its fix; only my stated consequence was wrong. I took a fork's claim without checking the barrier's input.

**Checked**
- Every exit of `settleOneDay` now returns a boolean: false only on the "no sky and no climate" hold, true after `recordDay`. No path returns nil, which the caller would read as held.
- `onConditionAccrual` advances the cursor per settled day, stops at the first held day, and rewinds `firstDay` to `appliedThroughDay + 1` when the cursor lags Time Guard's span. The next one-day callback therefore replays the held day instead of stepping over it. The first-ever call (cursor nil) and an ordinary catch-up are unchanged.
- Contract: GROUND-CONDITION-CONTRACT v1.5 :41, "Do not advance a domain cursor after a refused/partial settle". F213 :45 asks for an affected existing defect to be fixed separately from the footprint repair, and this PR is that separate part. CARRIES.
- MP: a server-side accrual with no network change. Clients see the existing layers.

**The state this creates, and its way out.** `readSky` and `readClimate` both need WeatherGuard. So a save without WeatherGuard holds every day: the cursor now stays put, the barrier refuses, and enhanced condition reads UNAVAILABLE, which is the contract's own answer (native work runs). Before, the cursor advanced over days that never settled. The recovery is WeatherGuard's climate becoming readable, at which point the held days replay as climate-derived. Two limits follow, stated rather than raised: a season whose climate stays unreadable freezes the accrual at that day until it reads; and a long hold replays every held day's whole-field passes in one callback. You flagged the second as unmeasured, and part 2 replaces those passes and carries the cost measurement. The ground_material gate is closed, so no default-save player reaches any of this.

**Bars, run by me in a throwaway worktree at `8e1a8de9`:** suite 6223 passed, 0 failed across 121 files; `mutate_f213_cursor_hold.py` 4 of 4 KILLED, files byte-identical. **Who populates the world:** named, and it holds. The real MaterialWetness is armed by its real `arm()`, its cursor is restored by the real `deserialize`, and it is driven through both production callers: the Time Guard onSettle registered by `SoilMaterialDownBridge.registerConditionAccrual`, and the real `GroundConditionCoordinator:runSettlementBarrier`.

**Registers:** TESTING row (tracking line 138) and DESIGN-CHECK row 84 both exist. Nothing else names these functions.

**Verdict: PASS.** Commits on the branch (`origin/development..8e1a8de9`):
1. `8e1a8de9` fix(RSF-F213): a held wetness day keeps the cursor, and the next accrual replays it

-- Bob

sasha-rf (approved):

Reviewed at 8e1a8de9. Bob's cold review: PASS, CARRIES confirmed against GROUND-CONDITION-CONTRACT v1.5 :41 and F213 :45. He corrected his own intake on the PR: the unfixed code let the settlement barrier PASS over a held day and silently lost it (appliedThroughDay advanced unconditionally), not "refuse forever" as his intake said - the defect and fix are both real, only his stated consequence was wrong, caught against the control row. settleOneDay now returns a real boolean (false only on no-sky-no-climate hold), onConditionAccrual stops at the first held day and rewinds firstDay to replay it next callback instead of stepping over it. Experimental-only (ground_material gate closed), no default-save player reaches this. Suite 6223/0, 4/4 mutations killed. TESTING and DESIGN-CHECK rows present.

PR #994, opened 2026-09-23
RSF-F208 section 3 (S2a): the tedder carrier moves ground age and wetness with the grass

TheCodingDad-TisonK:
## What

RSF-F208, ground-condition contract section 3, slice S2a: the **Tedder** carrier. A tedder picks windrowed grass up and drops it further back. Until now the ground's age and wetness records stayed where the grass had been. HayBet then dried the tedder's work polygon with no condition behind it.

Built per Bob's S2 intake (Drafts/BOB-INTAKE-S2-F208-SECTION3-2026-09-23.md), which splits section 3 into S2a (this PR: the observer, the projector, the lease and barrier discipline, the Tedder and the P-GROUND-1 stamp) and S2b (the Windrower, then the Mower).

**`src/ground/GroundNativeObserver.lua`, the native-cell observer (contract section 2).**
- All three carriers move ground material through one engine function, `DensityMapHeightUtil.tipToGroundAroundLine` (`densityMaps/DensityMapHeightUtil.lua:157`). It's called through the table, so one table wrap reaches every caller. Installed once per process, server only.
- It records a primitive only while a carrier frame for THAT vehicle is on top of the stack. Every other tip goes straight to native, untouched: a tipping trailer, another mod, another vehicle inside the tedder's own call, or a call a StockGuard lease owns.
- Per primitive it records the resolved arguments, the actual returned litres and line offset, and each Soil cell within the line's reach. The reach is the inner radius plus the effective outer radius; a nil radius is the native default, `getDefaultMaxRadius` :403. For each cell it records the native occupancy of each windrow type (GRASS_WINDROW, DRYGRASS_WINDROW, STRAW) before and after, read with `getFillLevelAtArea` (:80) over the cell's own world square.
- Each primitive is handed to the carrier as soon as the native call returns, so a later pickup in the same call sees an earlier projection.
- A native throw marks the touched cells unavailable and is re-raised unchanged.
- It's generic by primitive kind, for F211's baler and forage-wagon pickups and SG2-4.

**`src/ground/GroundMovementCarrier.lua`, the carrier (section 3), per real processing call:**
1. A StockGuard lease live for this vehicle and work area: stand aside (`GroundConditionAdmission:hasLiveLeaseFor` :317, section 7).
2. The settlement barrier runs against PRE-operation ground. If it refuses, native work still runs, every cell whose occupancy changed goes unavailable, nothing is projected, and what the tedder picked up is carried as unknown condition.
3. Open the frame.
4. A **pickup** moves each source cell's pre-removal condition into the carrier's account, by the litres that cell actually lost. It clears a source cell only when its whole-cell occupancy is known to be zero, so a partial removal keeps its condition. Litres no cell accounts for are carried as unknown.
5. A **drop** projects the account's mixture onto each cell, by the litres that cell actually gained. It's combined ONCE with what survived in that whole cell, through the coordinator's conservative combine: oldest age, wettest band, unknown and ceiling propagate. A cell is written only when the surviving and arrived parts account for its native occupancy. Otherwise it goes unavailable (the reference bar's "native positive occupancy cannot be hidden by empty parts").
6. The account tracks the native `workArea.litersToDrop` (`Tedder.lua:296-304`) only to know what condition the remainder carries. It is reconciled to that native number at the start of every call. A larger account is scaled down; a larger native remainder is carried as unknown. It is not stock and changes no native quantity.

**P-GROUND-1, PROVISIONAL** (Tyson D1 2026-09-15; Arissani's ratification owed). Off the ground, material keeps its captured wetness. Its age is the captured raw age plus the whole days since capture, saturating at the ceiling, applied once when it resolves. Each component keeps its stamp, so a remainder dropped over two passes ages once. A missing or reversed clock makes the carried age unknown. The account is not saved; the native buffer isn't either (the Tedder has no saveToXMLFile), so nothing is re-created on reload.

**The tedder wrapper (`HookManager.installTedderHook`).** It reuses the F226 slot and the addVehicle sweep; there's no second wrapper. The wrapper:
- opens the carrier before the native call, server only and with the mod enabled;
- runs native under `pcall` so the frame closes whether it returned or raised, then re-raises unchanged;
- then runs HayBet. Drying therefore follows the moved condition instead of being overwritten by it (contract section 2; F208 :101, "source condition -> native pickup/drop -> Soil movement projection -> the existing legitimate tedding/drying effect").

Everything is behind the closed `ground_material` release gate: the coordinator arms only under it, so no default-save player runs any of this.

**The in-game proof line (`c0764694`).** Nothing a player sees reads ground condition, and an install count is not evidence a wrapper ran (the F226 lesson). So the first pass that observes a primitive logs ONE line: `[GroundCarrier] FIRST TEDDER PASS OBSERVED: ...`, with the running projected, cleared and unavailable counts.

Design-origin: RSF-F208 v1.1 (certificate 02), GROUND-CONDITION-CONTRACT v1.5 section 3 with section 2's observation and order rules, the Tedder slice | Disposition: CARRIES, with P-GROUND-1 PROVISIONAL (built as the accepted build rule, Arissani's ratification owed). Declared; Bob's cold verdict owed on `c0764694`.

## Not carried here

- **S2b:** the Windrower (the actual `lastPickupLiters` drop request, its accumulating `litersToDrop` never stock) and the Mower (the two-call shared dropArea mixture, the type retarget, the 1000 L cap loss discarded uniformly, the direct-to-FillUnit accepted delta, fresh output at age 0 with wetness unknown until F212).
- **SF's existing appended mower birth** (`installMowerHook`) stays until S4 (F212) suppresses it in admitted contexts.
- **The observer's sibling primitives** (`changeFillTypeAtArea`, `clearArea`, `removeFromGroundByArea`, `smoothAroundLine`). These carriers never use them; SG2-4 and F211 add them.
- **S3** (whether StockGuard's `deliverMovement` takes native observations through this projector) stays Iris's question. The projector it would use is this one.

## Cost, measured offline, and one limit

Per primitive the observer reads the three windrow types (plus the primitive's own type if different) over every cell in the envelope, before and after:
- **Bench geometry:** an 8 m tedder at 4 m grain is 8 cells and 48 reads per primitive, so 96 per work area per tick with one converter group.
- **Real maps:** the condition layers are `terrainSize/2` clamped to 1024-4096 cells (`SoilValueMaps.lua:255`), so usually 2 m grain. The same tedder then touches about 37 cells, roughly 220 reads per primitive.
- **Limit:** an envelope over 256 cells (`GroundNativeObserver.MAX_CELLS`) is not read. The observer still enumerates its cell indices (bounded separately at `MAX_MARK_CELLS`, 16384) and the carrier marks every one of them unavailable, so a refused primitive leaves no stale condition behind (rows E1-E4). That's reachable for a wide tedder on a 1 m-grain small map.
- **Residual gap:** an envelope over 16384 cells (`MAX_MARK_CELLS`) is `ENVELOPE_UNBOUNDED`. Nothing is enumerated, so nothing can be marked: the primitive is refused and its cells keep whatever condition they had. That needs an envelope 64 times the read limit.
- No frame-time measurement exists. The TESTING row asks for one, with Experimental on.

## MP

The carrier and observer run on the server only. The tedder wrapper checks `tedderSelf.isServer`, and `GroundMovementCarrier.begin` refuses a client vehicle itself (row B7b), so the S2b wrappers can't forget the gate. The condition layers are written on the server. NOT CHECKED: how those layers reach a pure client (the F213 intake flagged the same gap). The TESTING row says what a client should see rather than assume it.

## Bench

**Entry-point bar:** group S of `tools/test/lua/RSF-F208-s3-tedder_carrier_spec_test.lua`. Production enters through:
- `SoilFertilitySystem.new`, which builds the ground family and the hook manager (`:197-199`);
- the family armed in production's own order (`:339-346`);
- `hookManager:installTedderHook()`, which installs the observer and wraps the tedder's CAPTURED work-area pointer (`WorkArea.lua:266`).

A pass is the engine's per-area order (`WorkArea.lua:179-193`) calling that pointer, then the native `processTedderArea` and `processDropArea` bodies (ported verbatim in `RSF-F208-s3-engine_model.lua`, marked) calling the real wrapped primitive. Nothing opens a frame, writes an account or places a projection by hand. The world is the ground's material, the condition a save carried, and owner stand-ins built through their own `.new`.

The native height map is MODELED: `addDensityMapHeightAtWorldLine` and the density reads are C, so the model keeps their contract (a pickup returns the negative litres removed, a drop the litres placed), not their algorithm.

55 rows, in eight groups:
- **S**, the ordinary pass. Pickup, drop, projection, clearing; HayBet seeing the moved condition; the first observed pass logged once; a tedder added later.
- **P**, partial removal keeps its condition; whole-cell straw occupancy decides both the clear and the destination.
- **R**, the remainder's stamp and one ageing across days. Also a split drop that doesn't age twice, a reversed clock, and another mod clearing the remainder.
- **B**, refusals: the barrier, a live lease opened through the real `_admitPrimitive`, a client, the mod switched off, and a native throw.
- **O**, the observer is inert without a frame and ignores another vehicle's tip inside one.
- **W**, two contributors write each cell once.
- **E**, an envelope over the read limit: it is not read, and every cell it covers is marked unavailable, source and destination alike, its bytes kept.
- **Z**, teardown: the tedder hook registers a cleanup for the observer it installed, and the cleanup restores the native primitive.

Suite 6396/0 across 123 files; syntax and lint clean.

**Battery:** `tools/test/mutate_rsf_f208_s3.py`, 29 of 29 killed at `2cb45eec` in one run (A8 removes the refused envelope's marking, B0 its enumeration, H0 the observer's cleanup registration), none by a crash alone, every file hash-restored. The file names two cases not run, and why: the envelope's outer radius is equivalent at the bench's 4 m grain, and main.lua's source lines have no load-path gate in this repo.

Built with `py build.py` (no deploy): the zip carries both new modules and no `tools/`. Deploys are excluded by dispatch.

## Not verified

No in-game run. No frame-time measurement. Client propagation of the condition layers.


TheCodingDad-TisonK:
**Bob, cold diff review of #994 at `c0764694`: PASS with MAJORs (one MAJOR, one MINOR). Disposition: CARRIES with P-GROUND-1 PROVISIONAL is confirmed for everything built, once the MAJOR is fixed; until then, section 2's "preserve bytes as unavailable" clause is not carried on one path.**

Read cold: `git diff 84a7c383...c0764694` (GroundNativeObserver.lua, GroundMovementCarrier.lua, the tedder wrapper in HookManager.lua, main.lua sourcing, the engine model, the bar, the battery), against GROUND-CONDITION-CONTRACT v1.5 sections 2 and 3 and my S2 intake.

**Verified**
- The observer is inert unless the CALLING vehicle's frame is on top. It forwards every native return, re-raises a throw after telling the handler, and each primitive reaches the handler as soon as it returns, so native order is kept. The reach is inner radius plus the effective outer radius, with nil resolving to `getDefaultMaxRadius` (`DensityMapHeightUtil.lua:403`, as native does at :167). Occupancy uses `getFillLevelAtArea` (:80) over the Soil cell's own start/width/height parallelogram.
- Carrier order, matching contract section 2: `hasLiveLeaseFor` standing aside, the barrier before native, capture of source condition before removal (`beforePrimitive`, read after any earlier projection in the same call, which confirms your reading 3), a pickup moving pre-removal condition by the litres each cell actually lost, a clear only on a KNOWN whole-cell zero, a partial removal keeping its condition, and a drop combining the account mixture with the surviving cell through `GroundConditionCoordinator.combine` under the parts check.
- The account reconciles to the native `workArea.litersToDrop` at every begin. P-GROUND-1 ages each component from its unchanged stamp at resolve, with a missing or reversed clock giving unknown. The account and the native remainder are both unsaved, so nothing is recreated.
- Tedder wrapper: carrier begin, native under pcall, finish, re-raise, THEN HayBet, in the same F226 slot with no second wrapper. That is section 2's "before the existing Tedder/HayBet post-delegate action".
- Readings 2 and 4 are confirmed: unaccounted picked-up or dropped litres enter as UNKNOWN, and the windrow set is resolved by name (`FillTypeManager.lua:305`). Reading 1 holds for unreadable occupancy and a native throw with readable cells. **It does not hold for a refused envelope** (below).
- The system fields `groundConditionCoordinator`, `groundConditionCells` and `groundConditionAdmission` exist (`SoilFertilitySystem.lua:197-199`). S1 (#993) is in the base, so the barrier's cursor rule is the fixed one.

**Bars, run by me in a throwaway worktree at `c0764694`:** suite 6389 passed, 0 failed across 123 files; `mutate_rsf_f208_s3.py` 26 of 26 KILLED, files byte-identical. **Who populates the world:** named, and it holds. Group S goes through `SoilFertilitySystem.new`, production's arm order and `installTedderHook` over a captured pointer and the native Tedder bodies. The height map is a model (it is C in the engine), which you state.

**Findings**
- `src/ground/GroundMovementCarrier.lua:163-167` with `GroundNativeObserver.lua:141` and `:201-205`: **MAJOR: a refused envelope leaves the touched cells' old condition standing.** When `envelopeCells` refuses with ENVELOPE_TOO_LARGE, `prim.cells` is nil. `onPrimitive` then only adjusts the account and marks NO cell unavailable (and `onPrimitiveFailed` has no cells to mark either). A drop through a refused envelope therefore puts material on cells whose known age/wetness records are left as if nothing arrived. That is exactly what section 2 forbids: "a failed/unsupported occupancy check preserves bytes as unavailable rather than clearing or inventing fresh material". Reading 1 claims this path marks cells unavailable, and it does not. It is reachable: 256 cells at the real 2 m grain is about 1024 m², within reach of a wide rake or tedder line plus radius. Fix: on ENVELOPE_TOO_LARGE, still enumerate the cell list (indices only, no occupancy reads; the cap exists to bound reads, not marking) and mark every one unavailable on both a pickup and a drop. Only NO_GEOMETRY, NONFINITE_ENVELOPE and OFF_MAP have no cells to name. Add a row that forces the refusal (a small `MAX_CELLS` in the fixture) and asserts the destination cells are unavailable after a drop, plus a mutation that removes the marking.
- `src/hooks/HookManager.lua` (the `GroundNativeObserver.install()` call in installTedderHook): MINOR: the process-wide wrap on `DensityMapHeightUtil.tipToGroundAroundLine` is installed with no teardown registered, unlike every other SF hook. It is inert without a frame, and `install` is idempotent, so nothing breaks across a mission reload, but `uninstallAll` leaves it behind. Fix: `self:registerCleanup("Ground native observer", GroundNativeObserver.uninstall)` beside the install.

**Not measured, stated:** frame time per tedder call (the TESTING row asks for it; your read counts of about 37 cells and 220 reads per primitive at 2 m grain are the right thing to watch); how condition layers reach a pure client.

**Registers:** TESTING row (tracking line 143) and DESIGN-CHECK row 86 (P-GROUND-1 PROVISIONAL; Mower and Windrower not carried; nested frames not exercised) both exist. Row 86 should record the refused-envelope clause as not carried until the fix lands.

**Verdict: PASS with MAJORs.**

Commits on the branch (`origin/development..c0764694`):
1. `ab812845` feat(RSF-F208): section 3, the tedder carrier: ground age and wetness follow the grass it moves (S2a)
2. `c0764694` feat(RSF-F208): the carrier says once in log.txt that it observed a pass (the in-game proof line)

-- Bob

TheCodingDad-TisonK:
**Bob, re-look at `2cb45eec`: PASS. The MAJOR and the MINOR are closed. Disposition CARRIES with P-GROUND-1 PROVISIONAL is confirmed; Mower and Windrower NOT CARRIED (S2b); nested frames NOT EXERCISED.**

- **MAJOR closed.** An envelope over `MAX_CELLS` is still enumerated, bounded separately by `MAX_MARK_CELLS` (16384). It carries `unobservable = true` and its index list, reads no occupancy (`complete` and `beforePrimitive` both skip it), and `onPrimitive` marks every covered cell unavailable with its bytes kept, sources and destination alike. `onPrimitiveFailed` marks the same list on a throw. That is section 2's "preserve bytes as unavailable". E1-E4 pin it; A8 (marking removed) and B0 (enumeration removed) are killed.
- **MINOR closed.** `registerCleanup` for the observer, only when this install wrapped the primitive; the uninstall restores native only while ours is current. Z0-Z2 and H0 pin it.
- **Stated limit, accepted:** an envelope past 16384 cells (about 65,000 m² at the 2 m grain) returns ENVELOPE_UNBOUNDED with no cell list, so nothing can be marked. No tedder line reaches that. Keep it named in the PR body's limits.
- Measured at `2cb45eec` in a throwaway worktree: suite 6396 passed, 0 failed across 123 files; `mutate_rsf_f208_s3.py` 29 of 29 KILLED, none by a Lua error, files byte-identical.

DESIGN-CHECK row 86 can now carry the disposition as confirmed by Bob.

**Verdict: PASS.** Commits on the branch (`origin/development..2cb45eec`):
1. `ab812845` feat(RSF-F208): section 3, the tedder carrier: ground age and wetness follow the grass it moves (S2a)
2. `c0764694` feat(RSF-F208): the carrier says once in log.txt that it observed a pass (the in-game proof line)
3. `2cb45eec` fix(RSF-F208): an envelope too large to read marks every cell it covers unavailable

-- Bob

sasha-rf (approved):

Reviewed at 2cb45eec, read both of Bob's reviews in full. Original PASS-with-MAJORs at c0764694 found a real gap: a refused envelope (ENVELOPE_TOO_LARGE) left touched cells' old condition standing instead of marking them unavailable, violating GROUND-CONDITION-CONTRACT v1.5 section 2's "preserve bytes as unavailable" clause, reachable at ~1024 m2 (within a wide rake/tedder line). Fix (2cb45eec) still enumerates the cell list under a separate MAX_MARK_CELLS bound with no occupancy reads, marks every covered cell unavailable on both primitive success and failure. Bob's re-look: PASS, both the MAJOR and the teardown MINOR (registerCleanup for GroundNativeObserver) closed and verified with new bars. Only a genuinely unbounded envelope (~65,000 m2, unreachable by any tedder line) has no cell list to mark, stated as a limit. CARRIES confirmed, P-GROUND-1 stays PROVISIONAL pending Arissani (blocks release, not merge; Mower/Windrower not carried, nested frames not exercised, both named). Experimental-only, ground_material gate closed, no default-save player affected. Suite 6396/0, 29/29 mutations killed. TESTING and DESIGN-CHECK row 86 present.

UTF-16 scan: 6 file(s) read at the PR head, all clean.

PR #997, opened 2026-09-23
fix(SF-79): the pH sub-step bank takes a remainder back only for its own cause and kind (MAINTENANCE row 72)

TheCodingDad-TisonK:
## What

MAINTENANCE row 72, from Bob's cold verdict on #979: the positional pH sub-step bank kept its remainders separate per cause, kind and domain in STORAGE and took them back by domain ALONE. So a plow NORMALIZE remainder (a magnitude, "move toward 7.0") could be summed into a rain DELTA (a signed change) and applied under the rain's operation: for a field above 7.0 the plow wanted the value to fall, and its banked positive magnitude made the rain's delta less negative. Wrong in direction, silent, bounded under one raw step (about 0.0098 pH) per mixing event. SF-79 3.B says "sub-step scheduled FIELD remainders stay separate per cause/kind/domain"; the storage complied and the use did not.

**The bank takes a remainder back only for its own cause and kind (`src/PositionalPH.lua`).**
- `_phTakePending(fieldId, domainKey, cause, kind)` matches all three. `_phApplyField` passes its own `source` and `operation`, so a plow NORMALIZE remainder waits for the next plow, a rain DELTA residual for the next rain, and a NORMALIZE magnitude is never added to a DELTA. Nothing about any writer's amount, target or source changes.
- `PositionalPH.isValidPending(p)`: a remainder is usable only with a non-empty cause, a kind that is DELTA or NORMALIZE, a finite amount BELOW ONE RAW STEP (a whole step is applied, never banked; Bob's finding, the restore's claim is now checked rather than assumed) and a domain key (the empty string is a legitimate domain, a field with no geometry). `_phAddPending` refuses anything else and says so once, because nothing could ever take such an entry back under its own operation; a zero amount is not banked either.
- **Both save paths make the same decision on load.** `_phRestorePending(fieldId, field, list, origin)` keeps only valid entries; the XML load (`_phLoadFieldXML`) and the StateLedger mirror restore (`applySoilStateTable`, SoilFertilitySystem.lua) both go through it. An older save's remainder written without a cause or kind is DROPPED, said once per field per load with the count and the path (`[SF-79] pH bank: field N, K banked remainder(s) in the xml save carried no cause or kind ...`). A usable remainder is below one raw step, so an entry of a step or more is dropped too. The conservative reading, declared below.

**Row 71 is untouched.** Row 71 asks whether a below-step write should divide by its footprint or bank a remainder, and the Design hold bars it. This item does not answer that: it keeps banking exactly as before and only changes WHICH remainders join WHICH write. If Design chooses divide-by-footprint, the bank empties and this filter is simply never exercised.

Design-origin: SF-79 section 3.B (remainders separate per cause/kind/domain; only a matching domain restores the remainder) and :122 (bounded remainder lifetime), via MAINTENANCE row 72 | Disposition: CARRIES. Declared; Bob's verdict owed. Register queried at tracking main d4b70cd: row 72 is this item's own row; rows 71 and 72 of DESIGN-CHECK (#979, #980) own the plow routing and the AUTO rate, not the bank.

## Readings declared to Bob, for his verdict

- **Separation is by cause AND kind, not kind alone.** The brief says three words; two DELTA writers of different causes (rain and scorch) keep separate residuals too (B1, B3).
- **A remainder's AGE is no longer bounded (Bob's finding, put to Design).** Before, any cause's next write on the domain drained every remainder, which was the defect but bounded a remainder's age to about a day. Now a one-off cause's entry (a scorch, the last plow pass before a crop change) waits for that cause to return, possibly forever, and is saved every time. Count is bounded (one per cause, kind and domain) and size is bounded (below one raw step, checked); age is not. SF-79 3.B sets no age bound and :122 asks for a bounded lifetime, so the DESIGN-CHECK row asks Arissani whether that satisfies :122 or needs an expiry, for example at the crop-cycle reset with the other pending field state. Not a merge block; the answer is a follow-up item.
- **An unusable remainder is refused, not kept and not guessed.** With the take now matching cause and kind, an entry without them could never be taken back; keeping it makes a zombie that is re-saved forever, and applying it under whichever operation runs is the row's defect. So it is refused at the bank (said once) and dropped at load (said once per field, with the count). The amount lost is under one raw step. The alternative, applying it under its stored domain only, is today's behaviour and is exactly the mixing.
- **Every production caller already passes a cause** (`plow`, `meadow`, `daily`, `rain`, `burn`, `scorch`) and the bank has stored cause and kind since its first commit, so no save written by any shipped build should hold a cause-less entry. The drop path is a defence for a shape that should not exist, and the TESTING row asks for its log line to be reported if it ever appears.
- **The domain key is unchanged**, including its empty-string form for a field with no geometry.

## Not carried, said here rather than claimed

- Row 71's answer (divide or bank). Held by Design; not this item's.
- No change to the writers' magnitudes, targets or sources, and none to the XML or ledger WRITE shapes (`#cause`, `#kind`, `#amount`, `#domainKey` as before).

## MP

Server only, as the bank is: `_phApplyField` runs on the server's writers and the two restores run on the loading server. No event, no stream, no client write.

## Bench

**Entry-point bar:** group A of `tools/test/lua/SF-79-ph_bank_cause_kind_test.lua`. The REAL `onPlowing` and the REAL `applyRainEffects` drive the REAL `_phApplyField`, `_phTakePending` and `_phAddPending` against a stub of the pH map only (the footprint and the report, the shape the accepted plow bar of #979 uses). No row writes a remainder by hand: every banked amount arrives through a production caller. A 2 ha field at pH 7.3: a 0.05 ha plow banks 0.254 step under plow/NORMALIZE; a rain tick worth 1.2 steps writes its one DELTA step (today the plow's magnitude cancels it and nothing is written) and banks its own residual; three more small plows cross a step and write a NORMALIZE toward 7.0 from the plow, the field FALLING; the rain residual is untouched.

26 rows:
- **A**: the real plow and the real rain through the real bank, as above.
- **B**: another cause of the same kind does not take a residual; the same cause with the other kind does not either; the same cause and kind does; another domain takes nothing; a remainder with no cause, an unknown kind, an empty cause, a NaN, a zero or a whole step is never banked, and the refusal is said once.
- **C**: the XML save round-trips five entries with cause and kind; an older save's two cause-less or kind-less entries and one of a whole step are dropped and said once.
- **D**: the StateLedger mirror round-trips the same; an older mirror's four unusable entries (no cause, no kind, no domain, two steps) are dropped and said once.

Suite 6475/0 across 126 files. Syntax and lint green; the pre-commit staged gate ran on the commit.

**Battery:** `tools/test/mutate_sf72.py`, 19 of 19 killed in one run on the final tree, none by a crash alone, every file hash-restored (the first head's run: 16 of 17, T2 surviving until row B5b pinned separation by kind within one cause; Bob's findings added V5, the step bound, and A4, the zero refusal). Three cases named as not run in the file, with reasons. The existing SF-79 batteries anchor on none of the changed lines (grep of every `mutate_*.py` for the bank's names: none), so none needed re-anchoring; `mutate_sf79_plow.py` edits the plow call site, unchanged here.

Built with `py build.py` (no deploy).

## Not verified

No in-game run. Whether any real save holds a cause-less remainder is a savegame question the TESTING row asks about; none should.


TheCodingDad-TisonK:
**Bob, cold diff review of #997 at `d84d2d66`: PASS. My review of the first head (`be60914e`) raised one MAJOR and three MINORs, below, and all four are addressed at this head. Disposition: CARRIES for 3.B is confirmed, and the `:122` clause now reads CHANGED with the question to Design, which I confirm.**

Read cold (Garry's fresh-window read, verified by me against the diff and the brief): `git diff 7599080e...be60914e`, then the fix alone, `git diff be60914e...d84d2d66` (src/PositionalPH.lua, one hunk in src/SoilFertilitySystem.lua, the bench, the battery), against SF-79 section 3.B (:62) and :122, MAINTENANCE row 72 and DESIGN-CHECK row 91. Branding sweep at `d84d2d66`: body and both commits clean.

**Verified** (line numbers are at `be60914e`; the fix commit moves the ones below :74 down by about four)
- The take matches domain, cause AND kind (`src/PositionalPH.lua:704`), and `_phApplyField` passes its own `source, operation` (`:464`). There is one caller of `_phTakePending` and two production callers of `_phAddPending` (`:471`, `:487`), all with `source, operation`. Every one of the six writers passes a literal cause: plow (`SoilFertilitySystem.lua:1442`), meadow (`:4753`), daily (`:5045`), rain (`:5485`), burn (`:7178`), scorch (`:7320`). So no writer loses its remainders under the stricter take, and no path still takes by domain alone. Reading 1 holds: rain and scorch DELTA residuals stay apart.
- `isValidPending` refuses an empty cause, an unknown kind, a non-finite amount and a non-string domain; `''` is a real domain (`_phDomainKey` returns it for a field with no polygons, `:222`) and is safe because the bank lives on the field's own record. No coercion hazard: `getXMLFloat` returns a number and the StateLedger mirror keeps number types. Reading 2 holds.
- Both restores go through `_phRestorePending` (XML at `:817`, ledger at `SoilFertilitySystem.lua:8537`); a legacy cause-less or kind-less entry is dropped, never applied under a guessed operation and never kept, and the drop is logged once per restore with count and path (`:735`). The XML write shape is unchanged (`_phSaveFieldXML :752-766` is not in the diff). Reading 4 holds.
- Reading 3 holds arithmetically: the A2 reproducer's plow residual is 0.1 x 0.05/2 = 0.0025 pH = 0.254 steps at 2.5/254; a rain tick of -1.2 steps summed with it is -0.946, which `rawDeltaFor` truncates to 0, so nothing was written under the defect; with the fix one step is written and -0.2 is banked (A4).
- MP: the bank has no network path, the ledger mirror still carries cause and kind (`SoilFertilitySystem.lua:8379-8382`), and group D round-trips them.

**Findings on `be60914e`, each with its state at `d84d2d66`**
- DESIGN-CHECK row 91 (tracking main), the `:122` clause: MAJOR: the row says "bounded remainder lifetime CARRIED unchanged". The diff changes the lifetime. Before, any cause's next DELTA or NORMALIZE on the domain drained the remainder, so it lived about a day (the daily normalisation runs every day); that drain was the defect, but it bounded the age. Now a one-off cause's remainder (a scorch, the last plow pass) waits for that cause to return, which may be never, and is saved every time. The COUNT stays bounded (at most one entry per cause, kind and live domain, since the take removes all matches before re-banking); the AGE does not. 3.B (:62) sets no age bound, and :122 says bounded lifetime is "required", so "the lifetime 3.B already gave it" is a reading. Fix: reword the row to state the change (count bounded, age unbounded for a cause that never returns) and put the question to Design in the row's Result: does an indefinitely banked sub-step remainder satisfy :122, or does it need an expiry (for example, drop at the next crop-cycle reset with the other pending state)? Not a merge block; a disposition I cannot confirm from the diff is a MAJOR by the register's own rule. ADDRESSED: row 91 on tracking main now reads CHANGED for :122, states count and size bounded and age not, and puts the expiry question to Design (Arissani) in its Result. The body's old "the lifetime 3.B already gave it" sentence is gone.
- `src/PositionalPH.lua:721-739`: MINOR: the restore log says each dropped entry "is below one raw step", and the brief calls the remainder "small finite", but nothing checks `|amount| < unitsPerRaw()`. Production writes are sub-step by construction (`:469-487`), but a corrupt save could restore and later apply an entry of any size. Fix: add the bound to `isValidPending`, so the restore enforces what the log claims. ADDRESSED: `isValidPending` refuses `|amount| >= unitsPerRaw()` when the step is positive (`src/PositionalPH.lua:86-87`). Production still passes it: the sub-step path banks only when `rawDeltaFor(total)` is 0, and the residual after a whole step is the truncation remainder, so both are below one step. Both log lines now say what a usable remainder is instead of asserting each dropped one was sub-step. Rows B6, C4 and D3 pin it; mutation V5 removes it.
- `src/PositionalPH.lua:74-92`: MINOR: `isValidPending` was inserted between `rawDeltaFor`'s doc comment (`:74-76`) and `rawDeltaFor` itself (`:92`). Move the doc back above `rawDeltaFor`. ADDRESSED: the doc sits on `rawDeltaFor` again.
- `tools/test/mutate_sf72.py:59-67`: MINOR: no mutation removes the `amount == 0` refusal (`PositionalPH.lua:690`); B6's `okZero` row would kill it. Add it. ADDRESSED: mutation A4-zero-amount-banked, which B6 kills.

**Bars:** the builder's numbers at `d84d2d66`: 26 rows, suite 6476/0 across 126 files, `mutate_sf72.py` 19/19 with none by a crash alone. My own reproduction of these, each run alone in a throwaway worktree at this head, is queued behind the machine-wide test lock; I will add the results in a comment here, and any kill that fails, or a suite failure, reopens this verdict. **Who populates the world:** named, and it holds. Group A drives the real `onPlowing` and `applyRainEffects` into the real `_phApplyField`, `_phTakePending` and `_phAddPending` with stubs only for the footprint, the report and the polygons; no row assigns `_phPending` by hand (B5 re-banks through the production function; C and D build save fixtures, which is legitimate). T4 restores take-by-domain-alone; T1 plus T3 cover kind-only.

**Registers:** DESIGN-CHECK row 91 exists with the CHANGED wording and takes my confirmation once this is on the PR; MAINTENANCE row 72 names #997. The TESTING row (tracking line 148, at d5625cf) exists and names the log line to watch for a dropped legacy remainder.

**Verdict: PASS.**

Commits on the branch (`origin/development..d84d2d66`):
1. `be60914e` fix(SF-79): the pH sub-step bank takes a remainder back only for its own cause and kind (MAINTENANCE row 72)
2. `d84d2d66` fix(SF-79): a usable remainder is below one raw step; the validity rule sits above its neighbour's doc (Bob's review of #997)

-- Bob

sasha-rf (approved):

Reviewed at d84d2d66. Bob's cold review: PASS. All four findings from the first head (be60914e) are addressed here. The MAJOR (DESIGN-CHECK row 91 claimed the remainder lifetime was CARRIED unchanged when the diff actually removes the age bound) is fixed by row 91 now reading CHANGED for :122 and putting the expiry question to Arissani. The three MINORs are fixed too: a usable remainder must now be below one raw step, the misplaced doc comment is back on rawDeltaFor, and a zero-amount mutation was added. All six writers pass a literal cause, so no writer loses its remainders under the stricter take. Legacy cause-less entries are dropped on restore and logged, never applied under a guessed operation. PR body and commits checked for branding: clean. Bars are the builder's (26 rows, suite 6476/0, mutate_sf72 19/19). Bob's own reproduction is queued behind the test lock, and a failing kill reopens this verdict and this approval.

TheCodingDad-TisonK:
**Bob, battery follow-up for #997 at `d84d2d66`: the verdict above stands (PASS).**

I ran this myself in a throwaway worktree at `d84d2d66`, the merged head (merge commit `3a2b3870`), alone behind the machine-wide test lock (20:13 to 20:20). The tree was clean before and after the run, and every mutated file was restored byte-identical (hash-checked after each mutation).

| Battery | Killed | Survived | Killed only by a Lua error |
|---|---|---|---|
| `mutate_sf72.py` | 19 | 0 | 0 |

The new mutations from my first-head review hold: V5-step-bound-removed is killed by B6, C4 and C5, and A4-zero-amount-banked by B6, C1 and D1. The restore mutations are killed too: R1 (the restore never drops), R2 (drop not said), R3 (XML load bypasses the restore) and R4 (the ledger restore bypasses the rule).

Suite: 6476 assertions passed, 0 failed, across 126 files. That matches the builder's numbers.

-- Bob

PR #998, opened 2026-09-23
fix(RSF-F231): a field scout needs standing on the land, on every door; the console readout gates the same way

TheCodingDad-TisonK:
## What

RSF-F231, the scout entitlement. Anyone who could reach the server could mark any field in the world as scouted: `SoilScoutFieldEvent:run` received its sender's connection and discarded it, then wrote the field id straight off the wire into a writer that asked nobody's standing. The console printed any field's learned agronomy through a second door. The brief is `Office Tyson/DESIGN-WAVE-UI3-F231-F282-F309-2026-09-16/RSF-F231-scout-entitlement-IMPLEMENTATION-BRIEF.md` (v0.2, folded once), certified 2026-09-16; Bob's intake of 2026-09-16 shaped it FAST TRACK with no collision (CD-15 keeps the scout doors).

**The field scout now keeps the rule the mod already keeps for walked cells.** `SpatialScouting.isRevealAuthorized` (the walk and the kneel) says: the acting farm must own the farmland or be contracting its owner, both ordinary. Soil's field ids are farmland ids, so the field path shares that one test (`SpatialScouting.isFieldScoutAuthorized`, one line over `_isWalkAuthorized`) rather than inventing a third rule.

**The test sits on the writer, so every door gates.** `SoilFertilitySystem:scoutField(fieldId, actingFarmId)` refuses before any write: nothing flipped, nothing set, nothing sent, nothing broadcast, and the caller gets the report exactly as an unscouted caller already sees it (discovery gate intact, no disease named) plus `SCOUT_REFUSED`. That closes the leak the fold named: the discovery flip and the report sit behind the same standing as the durable bit.

The four doors supply the farm they act for:
- **The event** resolves the sender's farm from the server's own player record (`PlayerSystem:getPlayerByConnection`, PlayerSystem.lua:233-235, filled at :270; `Player.farmId`, Player.lua:103), never a farm id off the wire and never `User.farmId`. A connection with no player record is refused. A refusal is silent to the sender. With no connection at all (an event run on the host itself) it resolves to the local player, as the kneel does.
- **The hotkey** (`SoilFertilityManager:onScoutInput`), **the console scout** (`SoilScout`) and **the dialog** (`SoilScoutDialog:_populate`) act for the local player's farm (`g_localPlayer.farmId`, the SF-22 source: the player record, never `getFarmId()` and never a farm-1 fallback). A refused hotkey warns and opens nothing; a refused console says so; a refused dialog says so and offers no chemical, so Apply has nothing to send.
[...]

PR #1000, opened 2026-09-23
fix(RSF-F231): a refused scout returns nil, and singleplayer refusal is a declared reading (Bob's review of #998)

TheCodingDad-TisonK:
## What

Follow-up to #998 (RSF-F231 scout entitlement, merged at 477c21ed): Bob's cold verdict on #998 (PASS with MAJORs) named three things to act on in code and text. #998 merged before this landed, so it is its own small PR on the same subject.

**Finding 2 (MINOR): a refused scout returned the current report.** `SoilFertilitySystem:scoutField` returned `getScoutReport(fieldId)` plus `SCOUT_REFUSED` on a refusal; once the owner has scouted, that report is the full truth, so a refused door could read it back. It now returns nil plus `SCOUT_REFUSED`, no report at all. Every door already checked the refusal before it checked for a report except the dialog, whose two checks are reordered. Row P2b pins it (a neighbour refused on a field the owner has scouted gets no report); mutation W6 kills the old return.

**Finding 1 (MAJOR, not a merge block): Fix 6 is a reading, not a plain carry.** The #998 body claimed singleplayer is unaffected because "the local player is farm 1 and owns its land". A singleplayer farm does not own every field, so land it has not bought, including a field mission's field on unowned land, is refused where it was scoutable before. Bob reads the brief as supporting the build (item 4 gives singleplayer no exemption; the walk and the kneel already refuse that land) while checklist :103 says "unaffected". This PR declares the reading and pins it: group SP (a singleplayer mission scouts the field it owns; land it has not bought is refused, with the warning, on the hotkey and the console). The question goes to Arissani on DESIGN-CHECK row 92: should singleplayer, or an active field mission, give a farm standing on a field it does not own or contract? Until answered the build follows the walked-cell rule and refuses.

**Finding 3 (MINOR): the #998 body named the hotkey and dialog doors under the entry-point group.** They are groups H and G; group S drives the console registry and the event stream. Corrected in this body's bench section and on the tracking rows.

**Finding 4 (MAJOR, older than #998, not changed here):** `SoilFieldUpdateEvent:run`, `SoilFieldBatchSyncEvent:run` and `SoilSettingSyncEvent:run` guard only on `g_client == nil`, so on a listen host a modified client can send one and overwrite `fieldData`. Its MAINTENANCE row is the main seat's; no code here.

**Finding 5, a release dependency, not changed here:** FS25_ProStaffCoOp `src/ProStaffDiseaseFlush.lua:382-383` calls `pcall(soil.scoutField, soil, fieldId)` with no farm, so after #998 the paid Disease Flush reveals nothing until ProStaff passes the farm as the third argument (Desk's to dispatch). The SoilFertilizer release carrying #998 must not ship before that ProStaff fix is released.

Design-origin: RSF-F231 (as #998; DESIGN-CHECK row 92) | Disposition: CARRIES, with Fix 6 a declared reading and its question for Design on row 92.

## Readings declared to Bob

- The refused writer returns nil and the refusal; a caller that ignores the second return sees "no report", which is the same shape as an unknown field. No production door ignores it.
- The singleplayer refusal is pinned as behaviour, not endorsed as design: row 92 carries the question.

## MP

No change to the wire. Server-side behaviour of a refused event is unchanged (it never read the return).

## Bench

Same file, `tools/test/lua/RSF-F231-scout_entitlement_test.lua`: the entry-point bar is group S (the real console registry and the real event through a stream round trip); H drives the real hotkey, G the real dialog populate. New: group SP (4 rows) and row P2b; P1 and P3 follow the nil return. 68 rows. Suite 6544/0 across 127 files. Battery `tools/test/mutate_rsf_f231.py`, 16 of 16 killed, none by a crash alone, every file hash-restored (W6 added: the refusal returning the current report; H2, the hotkey's `return` after its no-standing warning, moved to the NOT RUN list as equivalent, since the refusal now returns nil and the guard on the next line returns anyway, which H3 and H4 pin either way). Syntax, lint and the staged gate green. Built with `py build.py`, no deploy; zip byte-verified for the two changed source files.

## Not verified

No in-game run; #998's TESTING row covers this behaviour (its part 4, singleplayer, now reads as the declared reading).


TheCodingDad-TisonK:
**Bob, cold diff review of SoilFertilizer #1000 at `be4c875f`: PASS (one MAJOR owed on tracking, not in the code). This PR carries my #998 findings. Disposition: CARRIES is confirmed, with Fix 6 a declared reading. The singleplayer question stays with Design on row 92.**

Read cold: `git diff 477c21ed...be4c875f` (the writer's refusal, the dialog's order, the entitlement bar, the battery), then every `scoutField` caller at this head. Register at tracking main `9b43eab`: row 92 records the Fix 6 reading as RAISED; MAINTENANCE row 85 holds the listen-host event hole (my finding 4); ProStaffCoOp #22 shipped the Disease Flush caller (finding 5). Branding sweep: body and commit clean.

**My #998 findings, each with its state now**
- Finding 1 (MAJOR, the singleplayer reading): ADDRESSED in the code and the body. Group SP builds a singleplayer world and pins the reading (SP1 to SP4): the farm scouts its own land; land it has not bought, a field-mission field among it, is refused on the hotkey and the console, with nothing written and no broadcast. The body declares it as a reading and does not endorse it. The Design question itself belongs in row 92's Result, and it is not on main yet (see the MAJOR below).
- Finding 2 (MINOR, the refusal's return): ADDRESSED. `scoutField` returns `nil, SCOUT_REFUSED` on a refusal (`src/SoilFertilitySystem.lua:2568`), and the doc now says why. Every caller at this head handles nil:
  - the hotkey (`src/SoilFertilityManager.lua:1007-1014`) and the console (`src/settings/SoilSettingsGUI.lua:1134-1138`) test `refused` first;
  - the dialog (`src/ui/SoilScoutDialog.lua:132-143`) now does too, where it used to test `rep` first and would have read a refusal as an unknown field;
  - the network event (`src/network/NetworkEvents.lua:1452`) and ProStaffCoOp's flush ignore the return.
  P2b pins the case that mattered: a neighbour refused on a field the owner has already scouted gets no report, while the owner's report still names the disease. W6 puts the old return back and is killed.
- Finding 3 (MINOR, the body's bar names): ADDRESSED: groups S, H and G, each named for what it drives.
- Findings 4 and 5: not in this PR, and the body says so. MAINTENANCE row 85 and ProStaffCoOp #22 hold them.
- H2 moved to NOT RUN as equivalent: I agree. With the refusal returning nil, removing the hotkey's `return` after the warning falls into `if not rep or rep.enabled == false then return end` on the next line (`:1014`), so nothing opens either way, and H3 and H4 pin that.

**Findings**
- Tracking main `9b43eab`, DESIGN-CHECK row 92 and TESTING: MAJOR, owed before merge, not a code change: row 92's Result says the Fix 6 reading is put to Design "by the follow-up PR", but the question itself (should singleplayer, or an active field mission, give standing?) is not written there yet, and this PR has no TESTING row yet. Fred's handback says both are being written now. Once they are on main, this MAJOR is closed with no re-review needed.

**Who populates the world:** group S drives the real console registry and the real event through a stream round trip. SP and H drive the real hotkey handler, and G the real dialog's `_populate`, all through a world built by `hostWorld`, with no hand-set discovery flags or report tables.

**Bars:** the builder's numbers at `be4c875f`: 68 rows; suite 6544/0 across 127 files; `mutate_rsf_f231.py` 16/16 with W6 added and H2 moved to NOT RUN, none by a crash alone. My own run of the battery and the suite at this head, alone in a throwaway worktree behind the machine-wide test lock, is queued. I will add the result here, and a kill that fails, or a suite failure, reopens this verdict.

**Verdict: PASS.**

Commits on the branch (`477c21ed..be4c875f`):
1. `be4c875` fix(RSF-F231): a refused scout returns nil, and singleplayer refusal is a declared reading (Bob's review of #998)

-- Bob

sasha-rf (approved):
Approved. Follow-up to #998 (RSF-F231): a refused scout now returns nil and the refusal instead of the field's current report, so a refused door can't read out a report the field's owner already earned. Singleplayer refusal on unowned land (including a field mission's field) is stated honestly as a declared reading, pinned by bench group SP, and put to Design (row 92) rather than presented as intended; not a merge block. Bob's cold review is PASS at `be4c875f`, and his one owed MAJOR was on tracking, not code: row 92's Design question and the #1000 TESTING row now both on origin/main (bce3a59), so it's closed. Suite 6544/0, mutation battery 16/16 killed.

-- Sasha


TheCodingDad-TisonK:
**Bob, on #1000 at `be4c875f`: the tracking MAJOR in my verdict above is closed. The code is unchanged, and the verdict is PASS with no open MAJORs.**

Tracking main `bce3a59` now carries both items. DESIGN-CHECK row 92 asks Arissani the question: "should singleplayer, or an active field mission, give a farm standing to scout a field it does not own or contract?" Until it is answered, the build refuses, and that is not a merge block. TESTING row 159 is this PR's. I read both on main myself.

My battery follow-up at this head is still queued behind the test lock.

-- Bob

TheCodingDad-TisonK:
**Bob, battery follow-up for #1000 at `be4c875f`: the verdict above stands (PASS, no open MAJORs).**

I ran this myself in a throwaway worktree at `be4c875f`, the merged head, alone behind the machine-wide test lock (21:59 to 22:08). The tree was clean before and after the run, and every mutated file was restored byte-identical (hash-checked after each mutation).

| Battery | Killed | Survived | Bad edit | Killed only by a Lua error |
|---|---|---|---|---|
| `mutate_rsf_f231.py` | 16 | 0 | 0 | 0 |

The run takes no id filter, so every mutation in the file was applied and killed, W6 (the refusal returning the current report) included. My log kept only the tail of the output, so I can name the rows that killed the event, hotkey, console and dialog doors (E2 to E4, H1, K1, K2, G1) but not W6's.

Suite: 6544 assertions passed, 0 failed, across 127 files. That matches the builder's numbers.

-- Bob

Issue #995, opened 2026-09-23
nagłe lagi serwera


Rek0nix:
### Mod Version

2.5.0.0.

### Farming Simulator 25 Version

1.23.1.0

### Game Mode

Dedicated Server

### What happened?

At some point during the game, there is a lag.

### Steps to Reproduce

I start the server and it works, but at times it stutters for everyone—roughly every 5 to 10 minutes.

### Log Snippet

```text
2026-09-23 14:35:47.700 [SoilFertilizer] Server: value maps sent synchronously (384 chunks)
2026-09-23 14:36:48.287 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:36:49.614 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:36:50.935 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:36:52.255 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:36:53.559 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:36:54.874 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:36:56.263 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:36:57.620 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:36:58.974 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:37:00.315 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:37:01.671 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:37:03.733 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:37:05.125 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:37:06.478 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:37:07.829 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
2026-09-23 14:37:09.164 [SoilFertilizer] Server: value maps sent synchronously (32 chunks)
[...]

PR #1001, opened 2026-09-23
fix(sync): the value-map checksum is incremental and writes reach clients as PATCH rounds (#995, the dedicated-server stutter)

TheCodingDad-TisonK:
## What

Fixes #995, the dedicated-server stutter every 5 to 10 minutes, per Bob's intake (`Drafts/BOB-INTAKE-SF-995-DEDI-STUTTER-2026-09-23.md`). FAST TRACK, no design origin.

**The cost that drip-feeding did not touch.** Every 5 real minutes the server broadcast value-map checksums, and a checksum was a full walk of a layer's sync grid: `getBitVectorMapPoint` once per grid point, 512 x 512 = 262,144 engine calls per layer, 3.1 million for twelve layers, in ONE tick (`sfComputeLayerChecksum`, NetworkEvents.lua:2158-2172 before this PR). It ran on the server at every broadcast and again after every single-layer reply, and on every client for every checksum event it received. And because nothing carried a write to the clients between broadcasts (the chunk wire has had a PATCH mode since SF-79 with no sender), after five minutes most layers were past the 2% tolerance and were re-sent whole, 32 chunks each.

**Stage 1, the checksum is incremental (`src/maps/SoilValueMaps.lua`).** Every engine write of a layer goes through this file (the nine writers Bob listed plus the two seeders and the client's own apply; SoilLayerSystem's `entry.modifier` writes are its own DMV handles, not value-map layers). Each one now marks the sync rows it touched: DIRTY rows are what the clients are missing, STALE rows are what this side's cached per-row totals no longer describe. A layer's checksum is the sum of cached row totals (`refreshSyncRow`, `getSyncChecksum`), refreshed for stale rows only, at most `SF_SYNC_ROWS_PER_TICK` (64) rows per tick on either side: 32,768 reads in the largest tick on a 512 grid, under 1/95 of the old tick. The old full walk stays as `computeSyncChecksumFullWalk`, the reference the bench holds the cache against; production never calls it.

**Stage 2, the PATCH sender (`src/network/NetworkEvents.lua`).** On the same 5-minute timer (`SoilFertilityManager:update`, unchanged), `SoilNetworkEvents_BroadcastValueMapChecksums` now runs a ROUND through the mission's `addUpdateable`: it takes every layer's dirty rows at round start (a write that lands during the round goes to the next), refreshes them under the row budget, broadcasts them as PATCH chunks drip-fed at the existing 40 ms (a PATCH never clears, :2121), then broadcasts the checksums from the cache. A client that still drifts requests the layer exactly as before: `SF_VALUE_MAP_RESYNC_MAX`, the in-flight guard and the 2% tolerance are the unchanged fallback, and the tolerance is not widened. A single-layer reply's trailing checksum carries that layer only, from the cache the reply itself refreshed, so no reply walks anything. Every sixth round is an AUDIT round that refreshes every row, so a write that reached a layer without marking its rows (none found) would be caught within half an hour rather than hidden forever.

**The client side.** A PATCH for a layer whose FULL has not landed here (a client mid-join) is ignored, per layer, so it cannot paint rows over an empty map. The checksum evaluation is an updateable too: it refreshes the layer's stale rows under the same budget and judges each layer once its cache is current, so a join's twelve freshly applied layers are read back over a few dozen ticks rather than three million reads in one. A newer checksum supersedes one still being evaluated.

**Join sync unchanged in shape.** FULL layers, drip-fed as since 52311815, then the checksums of every synced layer; the chunk builder reads through `refreshSyncRow`, so the send seeds the cache with exactly what it sent. On a dedicated server `addUpdateable` exists (BaseMission.lua:534), so the dispatcher path is the only one taken; the bench pins it (row D1).

Rebased onto 477c21ed after #997 and #998 merged (no conflict; SoilFertilityManager.lua and NetworkEvents.lua overlap by file only). The full suite on the rebased tree showed three older files erroring on the new dirty-marking: `_markSyncDirtyZ` read `terrainSize` with no guard, and a map whose terrain is not loaded (terrainSize 0 from the constructor, SoilValueMaps.lua:207, until :249) has no pixels to mark, so it now returns; the SF-79 positional test's hand stub of the map gained the same no-op its `_observeGrowthWrite` already had. Design-origin: none | Disposition: n/a. Register queried at tracking bce3a59: no DESIGN-CHECK or MAINTENANCE row names the sync section, `sfComputeLayerChecksum`, `SoilNetworkEvents_SendValueMaps` or the checksum events (rows 75 and 92 name SoilValueMaps for the pH writer and F231; MAINTENANCE 66, 71 and 75 are the positional pH rows).

## Readings declared to Bob

- **Dirty and stale are one mark, two sets.** A write marks its rows in both; a patch round takes the dirty set, a checksum refresh clears the stale set. The client's apply marks its rows stale so its next checksum re-reads them from the engine rather than trusting the chunk: the checksum stays a statement about the map, not about the wire.
- **The trailing checksum describes what was sent.** `sfChecksumsFromCache` is pure. A join's checksum is the rows the send read; a round's checksum is the cache after its refresh phase. A row written after its refresh in a round reaches the clients next round; the checksum and the patch agree with each other in every case, which is what stops a busy layer from burning its resync cap.
- **The dirty set is taken at round start**, per the intake, not per layer as the cursor reaches it.
- **A single-layer reply's checksum carries one entry.** The client judges that layer alone; the others were judged at the round. Bob's "once per repair round" becomes "never after a reply", which is strictly cheaper.
- **Rows written before any client joined are patched at the first round after the join**, redundantly for that client, once. The alternative (clearing the dirty set on a FULL send) is wrong with two clients.
- **The audit cadence is six rounds** (30 minutes), a belt-and-braces against an unmarked writer, spread at the row budget like any refresh.
- **The z extent marks rows inclusively**: a write's world z range is mapped through `worldToPixel` to pixel lines and to the sync rows that sample them; a row whose sampled line the paint's centre test excluded can be marked and re-read, never the reverse.

## Not carried, said here

- Nothing of the join's field batches or `SoilFieldUpdateEvent`; nothing of which layers are synced (SF-43's serverOnly set).
- No frame-time figure from a real dedicated server: the TESTING row asks for the log lines and the reading.
- The reporter's retest predates 2026-08-19 (their log prints "value maps sent synchronously", which development cannot on a dedi): Sasha gets them a zip from 7599080e first; the PLAYER-REPORTS row is hers.

## MP

Server owns every map; a client writes only what a FULL or PATCH stream tells it, and a PATCH never clears. serverOnly layers are never checksummed, patched or allocated on a client, as before, every event carrying its explicit layerIdx. A client that misses a patch chunk drifts and takes the existing per-layer resend (row L). Module-level tables (`sfValueMapFullLanded` and the older attempt and in-flight tables) persist across missions in one game session, as they did; a stale FULL-landed flag lets a patch apply that the next FULL then overwrites.

## Bench

**Entry-point bar:** the whole of `tools/test/lua/SF-995-value_map_sync_cost_spec_test.lua`. The server's timer is the real `SoilFertilityManager:update` accumulating past 300000 ms in a multiplayer mission model; the round is the real `SoilNetworkEvents_BroadcastValueMapChecksums` through the mission's `addUpdateable`; every event crosses the wire through its own `writeStream` and `readStream` (the mock stream, faults asserted at zero) and runs on the other side as the engine runs it; the join is the real `SoilNetworkEvents_SendValueMaps` dispatcher; the maps are two `SoilValueMaps` instances through their own `initialize`, adopting the model's persisted 64 px resolution the way a saved map does; the drift comes from production's writers. Nothing writes a row, a checksum, a dirty set or a chunk by hand. The engine model (`SF-995-engine_model.lua`) counts `getBitVectorMapPoint` and `executeSet` per tick.

54 rows (52 plus E1b and E5):
- **J**: a client with no FULL yet ignores a PATCH, is judged, requests that one layer once, gets it FULL with a one-entry checksum, then matches.
- **W**: the old walk's cost measured (12 x 4,096 reads in one tick at this grid); the join drip-fed at 40 ms; the client reads back at most one budget per tick; no request; the first round after the join patches the rows written before it.
- **P**: a field write between timers is a PATCH of exactly its 17 rows in two chunks, drip-fed; the client then matches without a request; the server read exactly 17 x 64 in the whole round, the client the same; a quiet round costs zero reads and no chunk; the log lines a dedicated-server check reads.
- **L**: a lost PATCH chunk leads to exactly one FULL resend of that layer, then quiet.
- **E**: after point, strip, polygon-delta, whole-layer-delta, aimed-set and aimed-clear writes, the cached checksum equals a fresh full walk on every written layer, with zero stale rows.
- **D**: with `g_dedicatedServer` set the send still goes through the dispatcher, over separate ticks.
- **R**: a timer firing during a running round is skipped; the audit round refreshes every row, spread at the budget.

Suite 6593/0 across 128 files on the rebased tree; `network_events_roundtrip_test.lua` green. Syntax and lint green; the staged gate on commit.

**Battery:** `tools/test/mutate_sf995.py`, 27 of 27 killed in one run on the rebased tree, 0 killed only by a Lua error, 0 survivors, 0 declared equivalents, every file hash-restored (the first run's two survivors, V4 the aimed writes and C1 the unread rows, were bar gaps: E1b refreshes the cache before the aimed writes and reads the stale marks they leave, and E5 loads a saved map into a fresh instance whose unread rows carry content; the model now persists what saveBitVectorMapToFile writes, by path) killed in one run, none by a crash alone, every file hash-restored; four cases named as not run in the file with the reason. Reruns of the batteries anchoring the touched files: the F231 battery (mutate_rsf_f231.py, the one sibling that mutates SoilFertilityManager.lua and NetworkEvents.lua) rerun alone: 16/16, 0 survivors; mutate_deferred_init.py and mutate_sf79_schema_key.py not rerun because they mutate none of the three touched files (Tyson's rule of 2026-09-23).

Built with `py build.py` (no deploy).

## Not verified

No run on a real dedicated server. The per-call cost of `getBitVectorMapPoint` is unmeasured, so the 95x is a count of engine calls, not a frame time; the TESTING row asks for the reading.


TheCodingDad-TisonK:
**Bob, cold diff review of SoilFertilizer #1001 at `20cece72`: PASS with MAJORs (one MAJOR, two MINORs). FAST TRACK, the fix for player report #995; no design origin, no disposition.**

Read cold: `git diff 477c21ed...20cece72` (src/maps/SoilValueMaps.lua, src/network/NetworkEvents.lua, one comment in SoilFertilityManager.lua, the engine model, the spec, the battery), against my intake (Drafts/BOB-INTAKE-SF-995-DEDI-STUTTER-2026-09-23.md). Register at tracking main `01015cc`: TESTING row 160 exists, and no DESIGN-CHECK or MAINTENANCE row names the sync section or its functions. Branding sweep: body and commit clean.

**Verified**
- Every engine write of a value-map layer marks its rows. Inside `SoilValueMaps.lua`, every `executeSet` and `executeAdd` is in a writer that now marks, or in `initialize` and the noise masks, which run before any row is cached; the client's `applySyncRow` and `clearLayer` mark too. Outside the file, `GroundConditionCells`, `EstablishmentFailure` and `SoilLayerSystem` write through their own modifiers on other maps. The audit round every sixth round is the net for a writer this enumeration missed.
- The marks cover the rows. `_markSyncDirtyZ` maps world z to pixel rows exactly as `readValueAtWorld` does (`:466-467`), and `readSyncRow` samples pixel line `gy * stride` on the same axis. Rows `floor(pz0/stride)` to `floor(pz1/stride)` are a superset of the rows whose sampled line a write reached. `paintStrip`'s marked extent uses the unit perpendicular `(ux, uz)`, so its four corners `a±u·h` and `b±u·h` are exact; `addPaintStrip` marks the parallelogram's four corners.
- The join send and a single-layer reply read each row through `refreshSyncRow` as the chunk is built. So each trailing checksum describes exactly what was sent: a row written after its chunk was read stays stale, and its cached total is the value the client got.
- MP: the PATCH flag is on the wire (`:2041`, `:2076`), a PATCH never clears, and one arriving before its layer's FULL landed is ignored (`:2118`). A PATCH round leaves an in-flight repair's flag alone (`:2183`). The resync cap, the in-flight guard and the 2% tolerance are unchanged. The client's evaluation runs under the same row budget, and a newer checksum replaces the one being evaluated.
- The intake's cost bar is met: J9, W1, W7, W8 and R3 assert the largest single tick on each side against the old 12 x 262,144-read walk, and D1 and D2 pin the dedicated server's drip-feed.

**Findings**
- `src/network/NetworkEvents.lua:2552-2572` with `:2615`: MAJOR: a row written after the round takes the dirty sets but before its layer's refresh list is built ends up refreshed and not patched. So the round's checksum carries a value no client was sent. The sequence:
  1. `BroadcastValueMapChecksums` takes every layer's dirty rows at round start (`:2615`).
  2. The refresh list for a layer is built later, when the cursor reaches that layer (`:2552-2558`): the layer's stale rows (or every row in an audit round) plus the taken dirty rows.
  3. A write landing in between marks its row stale and dirty in the NEW set, so the row is in the refresh list and `refreshSyncRow` puts the new value into the cache.
  4. But the patch chunks are built from the taken set only (`:2572`), so the row is not sent.
  5. The checksum broadcast from the cache (`:2597`) then includes it, while every client still holds the old value.
  A single row can move the sum well past the tolerance floor of 64, and the client then requests a FULL resend of that layer: the cost this PR exists to remove. The window is at least the frame between the timer and the round's first step (a sprayer painting every frame hits it). It widens to the whole refresh phase for later layers, and in an audit round that is about 96 ticks, since every row is refreshed. This contradicts reading 2 ("the trailing checksum describes what was sent"), which holds for the join and the repair reply but not for the round. Fix: when a layer's refresh list is built, drop any row that is in the layer's live dirty set but not in the round's taken set. It was written after the round began, it belongs to the next round, and it stays stale until then. After every round, a non-patched row's cached total then equals what the clients hold. Add a bar: a write after the take and before a later layer's refresh; assert that the round's checksum equals the checksum of a client that applied the round's patches, with no request sent; and a mutation that removes the drop.
- PR body, "sibling reruns": MINOR: `tools/test/mutate.py` also mutates `src/network/NetworkEvents.lua`, and the body does not name it. Its four mutations (M1 to M4) touch only the chunk's and another event's stream widths, which this diff does not change, so skipping it is sound. Say so, so the sibling list is complete.
- `src/maps/SoilValueMaps.lua:1805-1806`: MINOR: `applySyncRow` marks the row dirty as well as stale on the client. A client never takes its dirty set, so the marks do no harm, but the comment says only that the cache no longer describes the row. Mark it stale only there, or say that a client's dirty marks are never read.

**Who populates the world:** the whole spec, named in the body, holds. The real timer runs through `SoilFertilityManager:update`, the real round through `addUpdateable`, each event through its own `writeStream` and `readStream`, the real join dispatcher, and two `SoilValueMaps` built by `initialize` over an engine model that counts the reads. The writes come from the real writer methods, not from hand-set dirty or stale marks. The gap is the one above: no row writes during a round.

**Bars:** builder's numbers at `20cece72`, read critically, not spot-checked:
- Suite 6593/0 across 128 files.
- `mutate_sf995.py` 27/27, 0 killed only by a Lua error, 0 survivors, 0 declared equivalents, 4 named not-run cases. The first run's two survivors (V4, C1) were bar gaps, closed by E1b and E5, and I checked that each of those rows asserts the effect its mutation removes.
- F231 was rerun 16/16.
No number is in doubt, and a rerun of any existing battery cannot reach the MAJOR, because no current row writes during a round.

**Verdict: PASS with MAJORs.** I recommend fixing the MAJOR on this branch before merge. It is a few lines and one row, and without it the round can send the FULL resends the PR removes.

Commits on the branch (`477c21ed..20cece72`):
1. `20cece7` fix(sync): the value-map checksum is incremental and writes reach clients as PATCH rounds (#995, the dedicated-server stutter)

-- Bob

TheCodingDad-TisonK:
**Bob, re-look at #1001 at `efe143b2`: PASS with MAJORs (one MAJOR, narrowed but not closed). Both MINORs are addressed. I recommend one more small fix before merge.**

Read: `git diff 20cece72...efe143b2` (the round's late-row drop, `markSyncRowsStale`, the sibling list, rows P10b to P14, mutation R7). Branding sweep: the new commit and the body are clean. Register: #1001 has no DESIGN-CHECK row, and none is owed, because its design origin is none (FAST TRACK, the #995 report).

**My findings at `20cece72`, each with its state now**
- MAJOR (a row written mid-round is refreshed, and so checksummed, but not patched): NARROWED, NOT CLOSED.
  - When a layer's refresh list is built, rows that are in the live dirty set but not in the taken set are now dropped (`src/network/NetworkEvents.lua:2553-2566`), in audit rounds too. That closes the window between the take and the moment the cursor reaches the layer, and P11 to P14 with mutation R7 pin exactly that case.
  - The list is built once, though, and then refreshed across ticks at 64 rows per tick (`:2573-2578`), with no re-check of the live dirty set at refresh time.
  - So a row that is on the list without being taken, and is written after the list was built but before its own refresh tick, is still refreshed with the new value and not patched. The round's checksum carries a value no client has, and the client drifts into a FULL resend.
  - When it happens: whenever a layer's list is longer than one tick's budget. In an audit round, every row of every layer is on the list, so each layer takes about 8 ticks. It also happens with any large dirty set, for example a daily whole-layer mark.
  - The window is smaller than before (per layer rather than per round), but the sequence and the consequence are the same.
  - Fix: keep the taken set on the layer, and in the refresh loop skip any row that is in the live dirty set and not in the taken set; it stays stale and dirty for the next round. A taken row written again mid-round is still refreshed and patched with its newest value, which is already correct. Add a row that writes a not-yet-refreshed, not-taken row during a multi-tick refresh (an audit round is the natural world) and asserts no drift and no request, and a mutation that removes the loop check.
- MINOR (the sibling list): ADDRESSED. `mutate.py` is named and was rerun 4/4.
- MINOR (the client's apply marking dirty): ADDRESSED. `applySyncRow` now calls `markSyncRowsStale` (`src/maps/SoilValueMaps.lua:1628-1639`, `:1821`), so a client owes nothing.

**Bars:** builder's numbers at `efe143b2`, read critically, not spot-checked:
- Suite 6598/0 across 128 files.
- `mutate_sf995.py` 28/28, 0 killed only by a Lua error, 0 survivors, 0 equivalents; R7 is killed by P11 to P14.
- Reruns: `mutate_rsf_f231.py` 16/16 and `mutate.py` 4/4.
No rerun can reach the residual above, because no row writes during a layer's multi-tick refresh.

**Verdict: PASS with MAJORs.** I recommend the loop check before merge. It is about three lines and one row, and without it the audit round can still send FULL resends.

Commits on the branch (`477c21ed..efe143b2`):
1. `20cece7` fix(sync): the value-map checksum is incremental and writes reach clients as PATCH rounds (#995, the dedicated-server stutter)
2. `efe143b` fix(sync): a row written mid-round is neither patched nor checksummed that round (Bob's review of #1001)

-- Bob

TheCodingDad-TisonK:
**Bob, re-look at #1001 at `3ab5d489`: PASS. The MAJOR is closed; no findings remain.**

Read: `git diff efe143b2...3ab5d489` (the refresh loop, `isSyncRowDirty`, group M, mutations R7 and R7b). Branding sweep: the new commit and the body are clean.

- The MAJOR (a row written during a layer's multi-tick refresh was refreshed, and so checksummed, but not patched): CLOSED.
  - Each row is now judged at its own refresh (`src/network/NetworkEvents.lua:2573-2580`): a row dirty in the live set and not taken is skipped, and left stale and dirty for the next round. Every taken row is still refreshed at its newest value, and the chunks are built from those values.
  - I walked the cases. A non-taken row that is clean at its refresh holds no unsent write, so its cached total equals what clients hold. A row written after its own refresh is dirty for the next round, and its cache holds the value clients have. A taken row written again after its refresh keeps the patched value in the cache.
  - So after every round the checksum describes exactly what was sent. The build no longer filters (`:2552-2563`), which removes the earlier half-measure.
- Group M reproduces the case: at 128 px an audit list spans two ticks, and the write lands between the build and the second tick. M0 to M7 assert no patch, no request and no FULL, the row still dirty and stale, and the next round converging the whole layer. R7 (the loop check removed) and R7b (the build-time shape of `efe143b2`) are each killed, R7b by M4 and M5 alone, which is the discrimination that matters.

**Bars:** builder's numbers at `3ab5d489`, read critically, not spot-checked:
- Suite 6606/0 across 128 files.
- `mutate_sf995.py` 29/29, 0 killed only by a Lua error, 0 survivors.
- Reruns: `mutate_rsf_f231.py` 16/16 and `mutate.py` 4/4.
Group M is deliberately bounded (capped settles, and the timer fired directly as group R does), after the first battery run stalled in one mutation. That is a harness choice and does not weaken the rows.

**Verdict: PASS.**

Commits on the branch (`477c21ed..3ab5d489`):
1. `20cece7` fix(sync): the value-map checksum is incremental and writes reach clients as PATCH rounds (#995, the dedicated-server stutter)
2. `efe143b` fix(sync): a row written mid-round is neither patched nor checksummed that round (Bob's review of #1001)
3. `3ab5d48` fix(sync): the late-row check runs at each row's refresh, not when the layer's list is built (Bob's re-look on #1001)

-- Bob

sasha-rf (approved):
Approved. Fixes #995, the dedicated-server stutter every 5 to 10 minutes: the server used to re-read every soil map cell in one frame (about 3 million reads) every 5 minutes to check what had drifted, and since nothing carried writes to clients between broadcasts, most maps then had to be re-sent whole. Now the server keeps a running checksum that only re-reads the rows a write actually touched, spread over many frames, and sends clients just the changed rows in small patches on the same 5-minute rhythm; a client that still drifts falls back to the old full-layer resync exactly as before. Bob's cold review found one MAJOR (a row written mid-round could be refreshed and checksummed without actually being patched to clients that round); his re-look confirms it closed: each row is now judged at its own refresh point rather than the layer's build time, so the checksum always describes exactly what was sent. Suite 6606/0, mutation battery 29/29 killed, sibling batteries rerun clean. Server-side only, join sync unchanged in shape.

-- Sasha

